PR #294, opened 2025-08-08
Remove the `replace-header` of cookie in cube-cos-api and ceph-dashboard's HAProxy conf

jdjgya:
#### What type of PR is this?

Fix

<br/>

As discussed in the scrum, we realized

- sometimes, not sure why the replace-header won't work as expected so cause api received the `cos_token` which is not able to be verified by api, but sometimes maybe good

- finally, we modified this in the cube-cos-api's SP object directly which makes the overall status more stable and easier 🙇 

<br/>

https://github.com/bigstack-oss/cube-cos-api/issues/214


Eandalf-Bigstack (approved):

jdjgya:
Thanks a lot @Eandalf-Bigstack for the quick review 🙇

Issue #155, opened 2025-06-10
Rolling upgrade definition: no VMs go into error states


leechienpang:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

leechienpang:
Build as below or any newer in develop branch can be verified against this ticket.

root@10.32.200.11:/home/jenkins/workspace/centos9_cubecos_prod/build/core/main/ship/CUBE_3.1.0_20250714-0610_5a804c3.iso

chase-chou:
@leechienpang 

## Manual test : failed

- Env : sky140

- version : CUBE_3.1.0_20250714-0610_5a804c3

- note : This environment is still available for you to access and review

### log :

```
(12/22) bootstrapping : keystone keycloak ceph
Bootstrap failed
```

```
[sky141 ~]# curl -v -k http://10.32.10.140:5000/v3
*   Trying 10.32.10.140:5000...
* connect to 10.32.10.140 port 5000 failed: Connection refused
* Failed to connect to 10.32.10.140 port 5000: Connection refused
* Closing connection 0
curl: (7) Failed to connect to 10.32.10.140 port 5000: Connection refused
[sky141 ~]# curl -v -k http://10.32.10.140:5000/v3
*   Trying 10.32.10.140:5000...
* connect to 10.32.10.140 port 5000 failed: Connection refused
* Failed to connect to 10.32.10.140 port 5000: Connection refused
* Closing connection 0
curl: (7) Failed to connect to 10.32.10.140 port 5000: Connection refused
[sky141 ~]# pcs status
Cluster name: cube-zs7JcQQJ7iUWQxwp
Cluster Summary:
  * Stack: corosync (Pacemaker is running)
  * Current DC: sky141 (version 2.1.7-5.el9-0f7f88312) - partition with quorum
  * Last updated: Mon Jul 14 19:35:27 2025 on sky141
  * Last change:  Mon Jul 14 18:19:00 2025 by root via root on sky141
  * 3 nodes configured
  * 6 resource instances configured

Node List:
  * Online: [ sky141 sky143 ]
  * OFFLINE: [ sky142 ]

Full List of Resources:
  * vip (ocf:heartbeat:IPaddr2):         Started sky143
  * haproxy     (systemd:haproxy-ha):    Started sky143
  * cinder-volume       (systemd:openstack-cinder-volume):       Started sky141
[...]

Issue #208, opened 2025-07-11
[Enhancement] COS - VIP Stability During FTS, Reboot, Cluster Power Cycle, and Upgrade


Eandalf-Bigstack:
**Is your feature request related to a problem? Please describe.**

- VIP is often in unstable state during first time set up, node reboot, cluster power cycle, and upgrade. We would need to make the VIP stable during these bootstrapping events. We would also need to clean up the dirty side effects of VIP moving between nodes. We would also need to ensure the rolling upgrade process smooth, to keep VM responsive during migrations to running hosts.

**Describe the solution you'd like**

- Make VIP stable during bootstrapping events, e.g., first time set up, node reboot, cluster power cycle, and upgrade.
- Clean up the dirty side effects of VIP moving between nodes.
- Ensure the VM responsive during migrations to running hosts while we are performing rolling upgrade.

**Describe alternatives you've considered**

**Additional context**

Eandalf-Bigstack:
Closed by #186 , #201 , #207 .

Eandalf-Bigstack:
Could be tested with CubeCOS v3.1.0 dev build on or after July 11th, 2025.

chase-chou:
FTS can be tested with Installation
Reboot & Cluster powercycle can be tested with IPIM
Upgrade can be tested with rolling upgrade

Eandalf-Bigstack:
From @leechienpang , the build `CUBE_3.1.0_20250718-0201_66b9e87` is ready for testing.

Eandalf-Bigstack:
From @leechienpang , the build `CUBE_3.1.0_20250721-0219_26937c5` is ready for testing.

chase-chou:
## Manual test : Testing

Env : 10.32.10.150

Version : CUBE_3.1.0_20250721-2307_97ada0d

- [x] 1. Upgrade
- [x] 2. Reboot
- [ ] 3. Powercycle
- [ ] 4. FTS

## Test Case

- [x] [CUBECOS-12 3cc cluster auto-rolling upgrade (ver. 3.0.0 GA)](https://app.qase.io/case/CUBECOS-12)

PR #329, opened 2025-08-18
fix: rolling upgrade cluster level SDK pacemaker status #155

leechienpang:

Fixes #

> Note

#### Additional documentation

```docs

```

### DIFF
--- a/core/sdk_sh/modules/sdk_os.sh
+++ b/core/sdk_sh/modules/sdk_os.sh
@@ -2390,7 +2390,7 @@ os_nova_instance_ping()
         return 0;
     fi
 
-    local active_host=$($HEX_CFG status_pacemaker | awk '/IPaddr2/{print $5}')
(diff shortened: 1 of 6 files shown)


Eandalf-Bigstack:
Related to #155 #208

We are reaching the limit of 10 PR linking to an issue, so we would use mentions instead.

core/sdk_sh/modules/sdk_os.sh
```diff
@@ -6,7 +6,7 @@ if [ -z "$PROG" ] ; then
     exit 1
 fi
 
-COOKIE_FILE=/tmp/cookie.txt
+%COOKIE_FILE=/tmp/cookie.txt
```
Eandalf-Bigstack: Not quite sure if `%COOKIE_FILE` is a valid variable name in bash.

hex
Eandalf-Bigstack, line 1: Maybe we also need a PR in https://github.com/bigstack-oss/hex

Eandalf-Bigstack (approved):

PR #350, opened 2025-09-04
bigstack/shell image add a troubleshooting tool by apk add busybox-extras

jdjgya:
#### What type of PR is this?

Feat

<br/>

For bigstack/shell OCI image, I would like to add a troubleshooting tool `telnet` by apk add busybox-extras.

Reasons:

- In the app-framework 2.0, we'll have a feature to check the port access between COS MGMT net and app-framework.

- Would like to reuse the bigstack/shell:latest for port access check, rather than have another similar image and build flow.

- The `telnet` will be more useful to check different service protocols.

### DIFF
--- a/core/k3s/build/buildah-image.sh
+++ b/core/k3s/build/buildah-image.sh
@@ -10,7 +10,7 @@ fi
 ctr=$(buildah from alpine:latest)
 
 buildah run $ctr apk update
-buildah run $ctr apk add --no-cache bash curl jq openssh-client sshpass
+buildah run $ctr apk add --no-cache bash curl jq openssh-client sshpass busybox-extras
 buildah run $ctr mkdir -p /var/alert_resp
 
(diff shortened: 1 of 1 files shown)


Eandalf-Bigstack (commented):
Maybe we could use `inetutils-telnet` instead? The implementation in `inetutils-telnet` seems more complete.

https://pkgs.alpinelinux.org/package/edge/community/x86/inetutils-telnet

jdjgya:
Hi @Eandalf-Bigstack , Thanks for the suggestion.

The reasons to choose `busybox-extras` are

- it's more lightweight than inetutils-telnet

- in the APK community, the `busybox-extras` is more often to be chosen if the use case is only for port access check.

- currently, no any cases will rely on inetutils-telnet's server framework, so I guess we can start from the simpler way

Eandalf-Bigstack (approved):

PR #438, opened 2025-11-04
fix(hex_config): Cirros image fails to import when glance is not full…

leechienpang:
…y ready

Fixes #

> Note

#### Additional documentation

```docs

```


Eandalf-Bigstack (commented):
We would need a signed commit in hex https://github.com/bigstack-oss/hex/pull/72 .

Eandalf-Bigstack (approved):

Issue #479, opened 2025-12-02
[Enhancement] Remove Rancher cube node driver


jdjgya:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Eandalf-Bigstack:
~@bigstack-oss/operation The node driver is planned to be removed in CubeCOS version 3.2.0. Users are recommended to migrate RKE1 workload managed using Rancher driver Cube to RKE2 in CubeCOS version 3.1.0.~

jdjgya:
note:

we finally decide to do this ticket in the 3.1.0 because the security fix below by removing the rancher-docker-machine.

1). https://app.fossa.com/projects/sbom%2B58648%2F20251121-syft-fs-cubecos-3_1_0.spdx/refs/branch/master/CUBE_3.1.0_20251126-1344_0c3ce97/issues/vulnerability/13841396?revisionScanId=93464051

jdjgya:
Hi @bigstack-oss/qa , 

Thanks a lot for pointing out this issue to the team.

The fix is already merged. The build which is generated after the 2025/12/19 can be used for verification.

Please let us know any question before/during testing 🙏

Eandalf-Bigstack:
We could check if the security concern still persists for fresh installs, but for upgrades, users need to remove the node driver from the cache of Rancher via Rancher UI. We might need an SOP or document for removing the node driver residual from upgraded CubeCOS clusters.

jimchen77:
## SBOM Scan: PASS

## Version

3.1.0_20260119-1030_7ef8b08

### DefectDOJO

http://10.32.21.110:8080/test/107
<img width="2339" height="979" alt="Image" src="https://github.com/user-attachments/assets/eb648e5c-0fdc-42d6-a4c4-baa8efad355d" />

PR #525, opened 2025-12-19
Deprecate cube node driver

arasHi87:
#### What type of PR is this?

cleanup

* Since the new release uses the standard openstack driver to create the app-framework, we can deprecate and remove the old cube node driver.

* Related to #479


Eandalf-Bigstack (commented):
We might also need to remove the directory core/rancher/rancher-machine-driver-cube https://github.com/bigstack-oss/cubecos/tree/f76eddb7de962eb02a47b933e37ef04b37cbbab6/core/rancher/rancher-machine-driver-cube , https://github.com/bigstack-oss/cubecos/blob/f76eddb7de962eb02a47b933e37ef04b37cbbab6/core/modules/config_haproxy.cpp#L194 , https://github.com/bigstack-oss/cubecos/blob/f76eddb7de962eb02a47b933e37ef04b37cbbab6/core/modules/config_haproxy.cpp#L215 , and https://github.com/bigstack-oss/cubecos/blob/f76eddb7de962eb02a47b933e37ef04b37cbbab6/core/modules/config_haproxy.cpp#L116-L122

Eandalf-Bigstack (approved):

PR #538, opened 2025-12-29
fix(terraform): skip sum check for self build rancher provider

arasHi87:
#### What type of PR is this?

Fix

* We built the rancher provider by ourself, so the checksum step will failed when `terraform apply`, need to skip the check for terraform.

Fixes # https://github.com/bigstack-oss/cubecos-private/issues/57

> Tested module by running `hex_config -vvve commit bootstrap terraform` manually, and without any error.

image

#### Additional documentation

```docs

```

### DIFF
--- a/core/terraform/scripts/terraform-action.sh
+++ b/core/terraform/scripts/terraform-action.sh
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+export TF_CLI_CONFIG_FILE=/etc/cube/cos/terraform/configs/override.tfrc
+
 pushd /tmp >/dev/null 2>&1
 timeout -k 1s 60s /usr/local/bin/terraform -chdir=/var/lib/terraform "$@"
 [ $? -eq 0 ] || exit 1
--- a/core/terraform/override.tfrc
+++ b/core/terraform/override.tfrc
@@ -1,0 +1,6 @@
+provider_installation {
+  dev_overrides {
+    "registry.terraform.io/rancher/rancher2" = "/var/lib/terraform/.terraform/providers/registry.terraform.io/rancher/rancher2/7.0.0/linux_amd64"
+  }
+  direct {}
+}
--- a/core/terraform/terraform.mk
+++ b/core/terraform/terraform.mk
@@ -6,4 +6,5 @@ rootfs_install::
 	$(Q)cp -rf $(TOP_BLDDIR)/core/terraform/terraform/ $(ROOTDIR)/var/lib/
 	$(Q)chroot $(ROOTDIR) mkdir -p /etc/cube/cos/terraform
 	$(Q)chroot $(ROOTDIR) mkdir -p /etc/cube/cos/terraform/values
+	$(Q)cp -f $(COREDIR)/terraform/override.tfrc $(ROOTDIR)/etc/cube/cos/terraform/configs/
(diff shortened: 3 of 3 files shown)


core/terraform/Makefile
```diff
@@ -23,6 +23,13 @@ terraform: $(RANCHER_UPSTREAM_BIN)
 	$(call RUN_CMD_TIMED,terraform -chdir=$@ init,"  INIT    $@")
 	$(Q)cp -f $(RANCHER_UPSTREAM_BIN) \
         ./$@/.terraform/providers/registry.terraform.io/rancher/rancher2/7.0.0/linux_amd64/terraform-provider-rancher2_v7.0.0
+	$(Q)echo 'provider_installation {' > ./$@/override.tfrc
+	$(Q)echo '  dev_overrides {' >> ./$@/override.tfrc
+	$(Q)echo '    "registry.terraform.io/rancher/rancher2" = "/var/lib/terraform/.terraform/providers/registry.terraform.io/rancher/rancher2/7.0.0/linux_amd64"' >> ./$@/override.tfrc
+	$(Q)echo '  }' >> ./$@/override.tfrc
+	$(Q)echo '  direct {}' >> ./$@/override.tfrc
+	$(Q)echo '}' >> ./$@/override.tfrc
+
```
Eandalf-Bigstack: Maybe we could place this file under `core/terraform` (`core/terraform/override.tfrc`) and use the following to install it into heavyfs in `core/terraform/terraform.mk`.

```make
$(Q)chroot $(ROOTDIR) mkdir -p /etc/cube/cos/terraform/configs
$(Q)cp -f $(COREDIR)/terraform/override.tfrc $(ROOTDIR)/etc/cube/cos/terraform/configs/
```

core/terraform/scripts/terraform-action.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+export TF_CLI_CONFIG_FILE=/var/lib/terraform/override.tfrc
+
```
Eandalf-Bigstack: It would be better to place config files under `/etc/cube/cos/terraform/configs/`.

```diff
- export TF_CLI_CONFIG_FILE=/var/lib/terraform/override.tfrc
+ export TF_CLI_CONFIG_FILE=/etc/cube/cos/terraform/configs/override.tfrc
```

core/terraform/Makefile
arasHi87: Thanks for the review! Do we have any specific guidelines for where to put files, or are we just playing it by ear?

Eandalf-Bigstack: Generally we follow Linux customs, variable data would go to `/var/lib/xxx/`, static data would go to `/usr/share/xxx`, configs would either go to `/etc/xxx/` or `/etc/cube/cos/xxx/` if the service does not own a `/etc/xxx/`. For CubeCOS specific data, for data would not be changed, the static one, we generally place them under `/usr/share/cube/cos/xxx/`. For configs, we would place them under `/etc/cube/cos/xxx/`. Though, for historical reasons, we have other files and destinations, like `/etc/settings.txt`, `/etc/settings.sys`, `/etc/settings.cluster.json`, and `/etc/appliance/`.

Eandalf-Bigstack (approved):
Please rebase to the latest develop branch.

PR #892, opened 2026-06-01
ci(pr): allow merging toward non default branches

Eandalf-Bigstack:
#### What type of PR is this?

- ci

- Allow merging toward non default branches


steven-chiu-bigstack (approved):
LGTM! 👍

PR #898, opened 2026-06-05
893 hex_sdk set GPU to passthrough mode

### LINKED ISSUE #893: [User Story] Set GPU Resource Type to pGPU Infra Implementation
## Description

As a CubeCOS operator, I want to select a GPU and set it to be pGPU for VMs so that we could pass a whole GPU card for a VM to utilize.

## Acceptance Criteria

- [x] GPU config file (`/etc/cube/cos/gpu/config.json`) is created during bootstrap.
- [x] GPU config file is migrated after COS update.
- [x] GPU config file is updated when the command is run successfully.
  - Please see #894 for config file schema.
- [x] Log error when the target GPU is in-use by any VM.
- [x] Re-apply GPU config (with `nvidia-smi` or any necessary tools) after node restarts.
  - Apply configs/update policies in hex_config instead of hex_sdk.

## Story Points: 3 

> | -- | -- | -- |

## QA Verification

> Every row below is a checkable claim: the exact command, and what you should see.
> Run them **on a GPU node as root**, over SSH.
>
> **What shipped.** #894 (`gpu_device_list`, `gpu_vgpu_profile_list`), #898 (`gpu_resource_set`
> in `sdk_gpu.sh` for pGPU), #1015 (`hex_config gpu_resource_set` for passthrough) — all three
> merged. Files: `core/modules/config_gpu.cpp`, `core/sdk_sh/modules/sdk_gpu.sh`, `core/main/cube.mk`.
>
> **Scope of this section.** This is a `User Story` (3 points) and it covers **pGPU
> (passthrough) only**. `sriovVgpu` and `migBackedVgpu` reach the same code path but are
> delivered by later tickets — do not test them here, and do not file their gaps against this
> ticket.
>
> **You do not need a vGPU-capable node.** Passthrough works on any node with a discrete NVIDIA
> card, so **sky150 (`10.32.150.1`) is enough** — no need to queue for cn13. That is the point
> of testing this ticket rather than its siblings.
>
> **There is no UI for this.** The UI half is #876 (still open). Everything below drives
> `hex_config` / `hex_sdk` directly.

### The two things to hold in your head before testing

1. **`/etc/cube/cos/gpu/config.json` is the authority, the hardware is the copy.** Setting a
   card to pGPU does not just bind a driver — it records the decision in that file, and
   `hex_config`'s `gpu` module re-derives the hardware state from the file on every boot and
   every settings commit (`config_gpu.cpp:1666` `Commit()`). If the two ever disagree, the file
   is what the product believes.

2. **A pGPU card disappears from `nvidia-smi`. That is correct, not a defect.** Once bound to
   `vfio-pci` the card is no longer enumerable by the NVIDIA driver. This is why the re-apply
   on boot looks the card up **by PCI address, not by UUID** (`config_gpu.cpp:1730`) — a
   UUID lookup would fail on every boot after the first. Expect `nvidia-smi -L` to stop listing
   a card you just passed through; use `hex_sdk gpu_device_list` instead, which reads the file.

### Preconditions

- A node with at least one discrete NVIDIA GPU, and that GPU **not currently attached to a
  running VM** (rows in §3 deliberately test the opposite).
- **On a node with no NVIDIA GPU every row below passes trivially and proves nothing** —
  `config.json` stays `[]`, every list command returns an empty array, and no binding is
  ever attempted. Check `lspci -nn | grep -i nvidia` returns at least one device before starting.
- You will need the card's UUID. Get it from the product, not from `nvidia-smi`:
  ```
  hex_sdk gpu_device_list | jq -r '.[] | "\(.id)  \(.pciAddress)  \(.type)  \(.status)"'
  ```
- §4 reboots the node. Do not run this section on a node someone else is using.

### 1 · The config file exists before anyone runs a command

|---|---|---|
| 1.1 | `ls -l /etc/cube/cos/gpu/config.json` | The file exists, owned `root:root`. It is created by the `gpu` module's `Commit()` on first boot — no one has to run anything to get it. |
| 1.2 | On a node where no GPU has ever been assigned: `cat /etc/cube/cos/gpu/config.json` | Exactly `[]` — an empty JSON array, not an empty file and not a missing file. |
| 1.3 | `rm /etc/cube/cos/gpu/config.json && hex_config commit bootstrap gpu && cat /etc/cube/cos/gpu/config.json` | The file is recreated as `[]`. **Only do this on a node with no GPU assignments** — on a node with assignments this discards them (see §6). |

### 2 · A successful `gpu_resource_set` updates the file *and* the hardware

Pick an idle card and record its UUID and PCI address from the Preconditions command.

|---|---|---|
| 2.1 | `hex_config -e gpu_resource_set <uuid> pgpu` | Exit code 0, no error on stderr. (The `-e` flag is what makes `hex_config` write its messages to stderr as well as syslog — without it a failure is completely silent. That was #1452.) |
| 2.2 | `jq '.' /etc/cube/cos/gpu/config.json` | An entry for `<uuid>` with `"type": "pgpu"`, its `"pciAddress"` recorded, and `"profiles": null` — profiles are a vGPU concept and are null for passthrough. |
| 2.3 | `lspci -nnk -s <pci-addr-without-domain> \| grep -i "kernel driver"` | `Kernel driver in use: vfio-pci` — the card has actually been taken away from the NVIDIA driver, not just noted in a file. Use the 4-char-domain form, e.g. `0000:42:00.0` → `42:00.0`. |
| 2.4 | `nvidia-smi -L` | The card is **no longer listed**. Per orientation point 2 this is the expected result of 2.3, not a defect. |
| 2.5 | `hex_sdk gpu_device_list \| jq -r '.[] \| select(.id=="<uuid>") \| .type, .status'` | `pgpu` then `idle` — the product still knows about the card even though `nvidia-smi` no longer does, because this reads the config file. |

### 3 · A card in use by a running VM cannot be switched

This is the acceptance criterion most likely to be skipped, and the one whose absence does
real damage — switching a card out from under a running VM.

**Setup.** You need the card attached to a *running* VM. On sky150, boot an instance with a
PCI-passthrough flavor onto the pGPU card from §2, and confirm with:

```
virsh list --state-running --uuid
virsh dumpxml <vm-uuid> | grep -i "<address .*bus='0x<bus>'"
```

where `<bus>` is the card's PCI bus in lowercase hex (e.g. `42:00.0` → `bus='0x42'`). The
in-use test in `sdk_gpu.sh:731-738` matches exactly on that bus/slot pair, so if this grep
finds nothing, the rest of this section will pass for the wrong reason.

|---|---|---|
| 3.1 | `hex_sdk gpu_device_list \| jq -r '.[] \| select(.id=="<uuid>") \| .status'` | `inUse` — the precondition for the rest of this section. If it says `idle`, your VM is not actually holding the card and 3.2 will pass vacuously. |
| 3.2 | `hex_config -e gpu_resource_set <uuid> sriovVgpu '[{"id":100,"count":1}]'` | Non-zero exit, and on stderr: `Error: GPU <uuid> is in-use` — verbatim, from `sdk_gpu.sh:1074`. |
| 3.3 | `jq '.' /etc/cube/cos/gpu/config.json` | The entry is **unchanged** — still `"type": "pgpu"`. The refusal happened before anything was written. |
| 3.4 | `lspci -nnk -s <pci-addr> \| grep -i "kernel driver"` | Still `vfio-pci`. The running VM's card was not disturbed. |

### 4 · The assignment survives a reboot, and is re-applied by `hex_config`

AC 5 says the re-apply happens in `hex_config`, not `hex_sdk` — that is what makes it run on
every boot without anything else scheduling it.

Shut down any VM holding the card first, so a failure here is about the re-apply and not about
a VM's own device claim.

|---|---|---|
| 4.1 | `reboot`, wait for the node, then `cat /etc/cube/cos/gpu/config.json` | The pGPU entry from §2 is still there. The file is on disk and nothing rewrites it at boot. |
| 4.2 | `lspci -nnk -s <pci-addr> \| grep -i "kernel driver"` | `Kernel driver in use: vfio-pci` **again** — the binding does not survive a reboot on its own; `Commit()` re-applied it from the file. This is the row that proves AC 5. |
| 4.3 | `grep gpu_bind_vfio_pci /var/log/hex_config.log \| tail -5` | A line showing `hex_sdk gpu_bind_vfio_pci <pciAddress>` executed during this boot's bootstrap, with the PCI address from the config file. |
| 4.4 | `grep -i "failed to re-apply" /var/log/hex_config.log` | No output. Any hit here is a real defect — file it. |

### 5 · The config file survives a COS upgrade

`config_gpu.cpp:1785` registers `CONFIG_MIGRATE(gpu, "/etc/cube/cos/gpu")`, which copies that
whole directory from the old filesystem to the upgraded one.

|---|---|---|
| 5.1 | Before an upgrade: `cp /etc/cube/cos/gpu/config.json /root/config.json.pre-upgrade` | Baseline saved. |
| 5.2 | After the upgrade completes: `diff /root/config.json.pre-upgrade /etc/cube/cos/gpu/config.json` | No differences — the assignments carried over. |
| 5.3 | `lspci -nnk -s <pci-addr> \| grep -i "kernel driver"` | `vfio-pci` — the upgraded node re-applied the migrated policy through the same §4 path. |

**If you cannot run a real COS upgrade**, say so on the ticket rather than ticking 5.x. This is
the one criterion with no honest shortcut — a manual file copy proves nothing about the
migration registration. It is acceptable for this row to be carried into upgrade testing.

### 6 · The failure mode this design introduces — test it deliberately

Making `config.json` authoritative means the file can now be wrong, and the node will faithfully
re-apply whatever it says.

|---|---|---|
| 6.1 | With a pGPU assignment in place, delete the file and re-commit: `rm /etc/cube/cos/gpu/config.json && hex_config -e commit bootstrap gpu` | The file comes back as `[]` and **the card is not unbound** — the binding is left behind with nothing recording it. Restore by re-running §2. This is known and expected; do not file it. |
| 6.2 | Hand-edit the file to a malformed JSON (e.g. delete a closing bracket), then `hex_config -e commit bootstrap gpu` | Non-zero, and in `/var/log/hex_config.log`: `Failed to parse GPU config file /etc/cube/cos/gpu/config.json:` followed by the parser's reason. It must **fail loudly**, not silently treat the file as empty. Restore your backup afterwards. |
| 6.3 | Hand-edit a `pgpu` entry to remove its `pciAddress`, then `hex_config -e commit bootstrap gpu` | Non-zero, and in the log: `GPU <id> has type pgpu but no recorded pciAddress; cannot re-apply binding` (`config_gpu.cpp:1720`). |

- **A pGPU card is absent from `nvidia-smi`.** By design (orientation point 2).
- **`gpu_device_list` reports extra fields** — `totalVramMiB`, `sriovVgpuProfileCountLimit`,
  `supportTypes`. These arrived after this ticket (`totalVramMiB` is #1424) and are not part
  of its schema. `totalVramMiB` being `null` on a card that was already pGPU before that change
  landed is expected — tracked in #1424.
- **No UI.** Setting a resource type from the web UI is #876, still open.
- **`sriovVgpu` / `migBackedVgpu` behaviour** is out of scope here — those belong to the
  follow-on tickets.
- **`hex_cli gpu resource_set`** exists as of #827 and drives the same `hex_config` command.
  Using it instead of `hex_config` directly is fine, but its own behaviour is #827's to verify.

### Already verified — and what that did not cover

- The three PRs were each exercised on **sky150** by their author at review time, with the
  before/after `config.json` contents recorded in #898's and #1015's PR bodies.
- The `hex_config gpu_resource_set` path — including its refusal messages and the `-e` flag
  needed to see them — was re-verified on **cn13** during #827 and #1452 (2026-09-09).

**What that did not cover, and what the rows above are therefore load-bearing for:** the
reboot re-apply (§4) and the upgrade migration (§5) were never exercised end-to-end for a
pGPU assignment — every cn13 run used vGPU carves, and no upgrade was performed. §4 and §5
are the rows that decide this ticket's verdict.

- [x] **QA Status declared at Done** — this is a `User Story`, which carries a `QA Status`, and the
  one (#818) — **do not reopen this ticket**. §4 is where that is most likely.

steven-chiu-bigstack:
#### What type of PR is this?

Feat

- Add `gpu_resource_set` command to `sdk_gpu.sh` for pGPU.

Fixes #893

Try it out on 150.1 (with dedicated shell script)

1. Change directory to `/etc/cube/cos/gpu/`
2. Verify GPU UUID
   ```bash
   [sky150 /etc/cube/cos/gpu]# nvidia-smi -L
   GPU 0: NVIDIA RTX A2000 (UUID: GPU-28eca022-0c66-97c1-a527-17e900626924)
   [sky150 /etc/cube/cos/gpu]#
   ```
3. Inspect current `config.json`
   ```bash
   [sky150 /etc/cube/cos/gpu]# cat config.json
   []
   [sky150 /etc/cube/cos/gpu]#
   ```
4. Run script
   ```bash
   [sky150 /etc/cube/cos/gpu]# sh ~/893-gpu-resource-set-pgpu.sh GPU-28eca022-0c66-97c1-a527-17e900626924 -type pgpu
   Successfully update GPU GPU-28eca022-0c66-97c1-a527-17e900626924 to pgpu
   [sky150 /etc/cube/cos/gpu]#
   ```
5. Verify `config.json` is updated
   ```bash
   [sky150 /etc/cube/cos/gpu]# cat config.json
   [{"id":"GPU-28eca022-0c66-97c1-a527-17e900626924","type":"pgpu","profiles":null}]
   [sky150 /etc/cube/cos/gpu]#
   ```

#### Additional documentation

None

### DIFF
--- a/core/sdk_sh/modules/sdk_gpu.sh
+++ b/core/sdk_sh/modules/sdk_gpu.sh
@@ -549,29 +549,18 @@ gpu_pgpu_attached_instance_get()
 
 gpu_resource_set()
 {
-    local gpu_uuid="$1"
-    shift
+    local gpu_id="$1"
+    local new_type="$2"
+    local profiles="$3"
 
-    local new_type=""
-    local profiles=""
-    local has_profiles_option=0
-
-    while [ $# -gt 0 ]; do
-        case "$1" in
-            -type) new_type="$2"; shift 2 ;;
-            -profiles) profiles="$2"; has_profiles_option=1; shift 2 ;;
-            *) echo "Error: unknown option '$1'" >&2; exit 1 ;;
-        esac
-    done
-
-    if [ -z "$gpu_uuid" ]; then
-        echo "Error: GPU UUID is required" >&2
-        exit 1
+    if [ -z "$gpu_id" ]; then
+        echo "Error: gpu_id is required" >&2
+        return 1
     fi
 
     if [ -z "$new_type" ]; then
-        echo "Error: -type is required" >&2
-        exit 1
+        echo "Error: new_type is required" >&2
+        return 1
     fi
 
     case "$new_type" in
@@ -582,13 +571,25 @@ gpu_resource_set()
             ;;
     esac
 
-    if [ "$new_type" = "pgpu" ] && [ "$has_profiles_option" = "1" ]; then
-        echo "Error: -profiles is not allowed when type is pgpu" >&2
+    if [ "$new_type" = "pgpu" ] && [ -n "$profiles" ]; then
+        echo "Error: profiles is not allowed when new_type is pgpu" >&2
         exit 1
     fi
+    
+    if [[ "$new_type" == "sriovVgpu" || "$new_type" == "migBackedVgpu" ]]; then
+        if [ -z "$profiles" ]; then
+            echo "Error: profiles is required when new_type is sriovVgpu or migBackedVgpu" >&2
+            exit 1
+        fi
+        
+        echo "TODO: perform profiles argument validation."
+        # Schema: an array of profiles `{ id: number, count: number }[]`
+        # In addition to the format, we must also validate the existence
+        # and capacity constraints (for MIG-backed) of the provided profiles.
+    fi
 
-    if ! $NVIDIA_SMI --query-gpu=uuid --format=csv,noheader,nounits 2>/dev/null | grep -qF "$gpu_uuid"; then
-        echo "Error: GPU UUID '$gpu_uuid' not found" >&2
+    if ! $NVIDIA_SMI --query-gpu=uuid --format=csv,noheader,nounits 2>/dev/null | grep -qF "$gpu_id"; then
+        echo "Error: GPU UUID '$gpu_id' not found" >&2
         exit 1
     fi
 
@@ -598,21 +599,20 @@ gpu_resource_set()
     fi
 
     local gpu_config="[]"
-    local raw
-    raw=$(cat "$GPU_CONFIG_FILE_PATH" 2>/dev/null)
-    if echo "$raw" | jq -e 'type == "array"' >/dev/null 2>&1; then
-        gpu_config="$raw"
+    local raw_config
+    raw_config=$(cat "$GPU_CONFIG_FILE_PATH" 2>/dev/null)
+    if echo "$raw_config" | jq -e 'type == "array"' >/dev/null 2>&1; then
+        gpu_config="$raw_config"
     fi
 
     local current_type
-    current_type=$(echo "$gpu_config" | jq -r --arg id "$gpu_uuid" \
+    current_type=$(echo "$gpu_config" | jq -r --arg id "$gpu_id" \
         'map(select(.id == $id)) | if length > 0 then .[0].type else "unset" end')
 
     local pci_bus_id
-    pci_bus_id=$($NVIDIA_SMI --query-gpu=uuid,pci.bus_id --format=csv,noheader,nounits 2>/dev/null | \
-        awk -F',' -v uuid="$gpu_uuid" 'index($1, uuid) {print $2}' | xargs)
+    pci_bus_id=$($NVIDIA_SMI --query-gpu=pci.bus_id -i "$gpu_id" --format=csv,noheader,nounits 2>/dev/null)
     if [ -z "$pci_bus_id" ]; then
-        echo "Error: could not get PCI bus ID for GPU $gpu_uuid" >&2
+        echo "Error: could not get PCI bus ID for GPU $gpu_id" >&2
         exit 1
     fi
 
@@ -625,7 +625,7 @@ gpu_resource_set()
             for vm_id in $(virsh list --state-running --uuid 2>/dev/null); do
                 if virsh dumpxml "$vm_id" 2>/dev/null | \
                     grep -q "bus='0x${pci_bus}'.*slot='0x${pci_slot}'"; then
-                    echo "Error: GPU card $gpu_uuid is in-use" >&2
+                    echo "Error: GPU card $gpu_id is in-use" >&2
                     exit 1
                 fi
             done
@@ -634,16 +634,16 @@ gpu_resource_set()
         if [ "$current_type" = "sriovVgpu" ]; then
             unset_sriov_vgpu "$pci_bus_id"
         elif [ "$current_type" = "migBackedVgpu" ]; then
-            unset_mig_backed_vgpu "$gpu_uuid"
+            unset_mig_backed_vgpu "$gpu_id"
         fi
 
         local new_config
         new_config=$(echo "$gpu_config" | jq -c \
-            --arg id "$gpu_uuid" \
+            --arg id "$gpu_id" \
             'map(select(.id != $id)) + [{id:$id, type:"pgpu", profiles:null}]')
 
         echo "$new_config" > "$GPU_CONFIG_FILE_PATH"
-        echo "Successfully update GPU $gpu_uuid to pgpu"
+        echo "Successfully update GPU $gpu_id to pgpu"
     else
         echo "Error: type '$new_type' is not yet implemented" >&2
         exit 1
--- a/core/sdk_sh/modules/sdk_gpu.sh
+++ b/core/sdk_sh/modules/sdk_gpu.sh
@@ -547,6 +547,130 @@ gpu_pgpu_attached_instance_get()
     echo "null"
 }
 
+gpu_resource_set()
+{
+    local gpu_uuid="$1"
+    shift
+
+    local new_type=""
+    local profiles=""
+    local has_profiles_option=0
+
+    while [ $# -gt 0 ]; do
+        case "$1" in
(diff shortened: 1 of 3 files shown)


core/sdk_sh/modules/sdk_gpu.sh
```diff
@@ -367,6 +367,130 @@ gpu_vgpu_profile_list()
...
+            -type) new_type="$2"; shift 2 ;;
+            -profiles) profiles="$2"; has_profiles_option=1; shift 2 ;;
+            *) echo "Error: unknown option '$1'" >&2; exit 1 ;;
+        esac
+    done
+
+    if [ -z "$gpu_uuid" ]; then
+        echo "Error: GPU UUID is required" >&2
+        exit 1
+    fi
```
Eandalf-Bigstack: We could perform this check right after https://github.com/bigstack-oss/cubecos/pull/898/changes#diff-0c1a8c138b7181038b6cdb19c196ff56ba9f65a91f0fb5f955bc63e6b6f397e9R372

Eandalf-Bigstack (approved):

Eandalf-Bigstack (approved):

PR #900, opened 2026-06-05
feat: add `gpu_pgpu_attached_instance_get <pciAddress>` to `sdk_gpu`

steven-chiu-bigstack:
#### What type of PR is this?

Feat

Add `gpu_pgpu_attached_instance_get <pciAddress>` to `sdk_gpu`.

Related to #859

I couldn't find a better way to get the attached pGPU instance info (specifically the `id`) on GO server without relying on `hex_sdk`. We'll need this functionality to get the List GPUs API working correctly.

- Usage: `hex_sdk pgpu_attached_instance_get <pciAddress>`
- Returns: `{ id: string, name: string }` or `null`

#### Additional documentation

None

### DIFF
--- a/core/sdk_sh/modules/sdk_gpu.sh
+++ b/core/sdk_sh/modules/sdk_gpu.sh
@@ -367,6 +367,34 @@ gpu_vgpu_profile_list()
         'map(select(.id == $id)) | if length > 0 then .[0].profiles else null end'
 }
 
+gpu_pgpu_attached_instance_get()
+{
+    local pci_address="$1"
+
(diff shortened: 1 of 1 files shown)


steven-chiu-bigstack:
Hi @Eandalf-Bigstack 

In short, I couldn't find a way to get the attached pGPU instance ID using NVML alone, so I had to rely on `hex_sdk`.
Please let me know if you have a better solution in mind. 🙏

Eandalf-Bigstack (approved):
Great! This should work.

Issue #824, opened 2026-04-16
[User Story] Show the History of GPU Workload (Grafana)


Eandalf-Bigstack:
## Description

As a CubeCOS operator, I want to see the graph of GPU workload so that we could know the GPU workload history.

## Acceptance Criteria

> A detailed list of conditions that must be met for the story to be accepted. These statements must focus on functional behavior and constraints rather than specific code-level implementation.
  
- [x] Infra: Add a new Line Chart panel tracking GPU Utilization (compute workload) over time to the System section under `Home > Dashboards > Device`. 
   - [x] 1). Add a hidden `$GPU_HOST` variable to `device.json`
   - [x] 2). Create a new "GPU Utilization" Time Series panel using util_gpu filtered by that variable and grouped by pciid
- [x] Backend: Update the node API to accept a hostname and return a deep-link URL pointing to the new GPU Utilization panel ID
- [x] Frontend: Add a "View Workload History" button on the physical GPU table row that opens the backend-generated Grafana URL in a new tab
- [x] Collector: the `gpu.host` series must survive every GPU resource type, including `migBackedVgpu`
   - [x] 1). With all four resource types present on one node, `hex_sdk -f line gpu_host_stats` emits exactly one line per *visible* card, each line has exactly one tag/field separator, and telegraf logs no `metric parse error`
   - [x] 2). A card whose data cannot be read must never affect another card's line (verified by feeding a corrupted record)
   - [x] 3). On a MIG-backed card `mem_*` are the **device-level** values (not a GPU instance's framebuffer), and `util_*` are **absent** rather than `N/A`
   - [x] 4). On a healthy node the output is byte-for-byte identical to the previous implementation (no new InfluxDB series)

### Expected behaviour per GPU resource type

| Resource type | GPU Utilization panel | VRAM panel | Why |
| --- | --- | --- | --- |
| unset / native | line shown | Used (solid) + Total (dashed) | — |
| sriovVgpu | line shown | Used includes the vGPU allocation | — |
| **migBackedVgpu** | **no line** | line shown, Total is the whole card | NVIDIA reports no device-level utilization once MIG is enabled; per-GPU-instance metrics would need DCGM (out of scope, follow-up) |
| pgpu | no line | no line | the card is bound to vfio-pci and handed to a VM — no host-side visibility, by design |

Note: `mem_used + mem_free != mem_total`; the difference is `Reserved` (~2288 MiB on
[...]

PR #1017, opened 2026-06-26
fix(hex_sdk): match uppercase GPU when parsing util_gpu in gpu stats

SekiXu:
#### What type of PR is this?

Fix

`gpu_host_stats` and `gpu_vm_stats` parse GPU utilization from `nvidia-smi`, which prints the label as **`GPU`** (uppercase) under `Utilization`.
The collectors grepped for `Gpu`, which never matched, so `util_gpu` came out **empty**.

An empty field makes the emitted InfluxDB line protocol malformed (`...,util_gpu=,util_mem=0,...`). Telegraf rejects the whole point on parse, so **no `gpu.host` / `gpu.vm` data ever reaches InfluxDB** — the measurement is simply absent.

Fix: use `grep -iw Gpu` (case-insensitive, whole-word) so it matches `GPU` without also catching longer tokens.

None

Out of #824's scope, but required so the #824 panels have data.

#### Additional documentation

None


SekiXu:
Dashboards > Host > GPU row

image

leechienpang (approved):

PR #1018, opened 2026-06-26
fix: source Device dashboard Host variable from existing host tag

SekiXu:
#### What type of PR is this?

Fix

The Device dashboard's `$HOST` template variable queried `SHOW TAG VALUES FROM "ipmi_sensor" WITH KEY = "hostname"`. That tag does **not exist** — `ipmi_sensor` points are tagged with **`host`** (the Telegraf agent hostname), not `hostname`. The query returned nothing, so the **Host dropdown was empty** and could not filter the dashboard by node.

Fix: query the `host` tag instead, so the dropdown lists the node.

None

None

#### Additional documentation

None

### DIFF
--- a/core/grafana/provisioning/dashboards/device.json
+++ b/core/grafana/provisioning/dashboards/device.json
@@ -4285,7 +4285,7 @@
           "value": "DESKTOP-ENMHMJG"
         },
         "datasource": "influxdb-telegraf",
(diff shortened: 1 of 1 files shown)


SekiXu:
Dashboards > Device > System row

image

leechienpang (approved):

Issue #1029, opened 2026-06-27
[Bug] Error/warning log flooding on HA control-converged clusters


traviswu-bigstack:
## Description

On HA control-converged clusters several services flood error/warning logs at steady state, drowning real signal (observed in Graylog across the control nodes):

- **masakari hostmonitor** runs even when HA is off, and probes corosync by sniffing packets with `tcpdump` → continuous errors.
- **HAProxy galera listen** has no `timeout server`, so idle DB connections are cut → `MySQL server has gone away (2013)` / "connection aborted" floods.
- **monasca `ib_network`** check errors on NICs without a counters dir (e.g. irdma) → repeated tracebacks.
- **logstash** starts before influxdb/kapacitor → boot-time connection-error flood.

## How to Reproduce

1. Deploy a 3-node HA control-converged cluster.
2. Tail the service logs (or open Graylog).
3. Observe continuous masakari / haproxy-galera / monasca `ib_network` / logstash errors at idle.

## Expected behavior

Clean logs at steady state — no repeating error/warning floods from these services.

## Additional Context (Optional)

Root causes are config/policy-layer: masakari not gated on HA + tcpdump corosync probe; missing galera `timeout server`; `ib_network` not guarding counter-less NICs; logstash start ordering. Fixed in the linked PR.

## Output artifacts (Definition of Done)

> Beyond code, docs, and config changes, completing this issue must also deliver:

- [x] **Handbook knowledge update** — covered: [error-log-floods-corrections](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/error-log-floods-corrections.md) — land the durable, team-readable knowledge from this work into the bigstack-handbook cubecos kb (`kb/cubecos/…`) via `/bigstack-core:save-to-handbook` (Topic / Runbook / Known-issue / ADR as fits). _(drafted: `kb/cubecos/known-issues/error-log-floods-corrections.md`)_

PR #1053, opened 2026-07-01
fix: cut steady-state log flooding (ssh multiplexing, sdk_is self-check, pacemaker pe-error)

traviswu-bigstack:
#### What type of PR is this?

/kind bug

Cut steady-state log flooding found post-FTS on a single-node install:

- **ssh ControlMaster multiplexing** in the generated `/etc/ssh/ssh_config` — the
  health framework's per-service-per-node ssh checks reuse one session instead of
  flooding sshd/logind/sudo (a `cluster check` dropped from **118 self-ssh logins
  to ~1**, 0 within the persist window). `WriteSshConfig()` now runs on every
  commit so this also lands on reboot/upgrade, not just fresh install.
- **sdk_is** short-circuits `is_remote_running`/`is_remote_active` to the local
  `is_running`/`is_active` when the target is this node (skip ssh-to-self).
- **pacemaker**: reapply `stonith-enabled=false` + `pe-error-series-max` in
  `cube_cluster_start` — the `SetupCluster` set is lost when the CIB isn't up yet,
  leaving single-node STONITH-error flooding and unbounded `pe-error` file growth.

Refs #1029

> #1029 spans three repos — this is the **cubecos** half. The air-gapped
> `dnf-makecache` half ships in **bigstack-oss/hex#119**; the cos-api Debug
> poll-logging half in the companion **cube-cos-api** PR. Keep #1029 open until all
> three land.

> **No hex submodule bump here** — the `dnf-makecache` change is delivered by
> hex#119, not this PR (avoids a dangling submodule pointer on develop).
> Air-gapped safe: config-only, no new install/runtime fetches.

#### Additional documentation

```docs
kb/cubecos/known-issues/steady-state-log-floods-corrections.md  — in handbook PR #35
```

### DIFF
--- a/core/main/proj_functions
+++ b/core/main/proj_functions
@@ -273,6 +273,11 @@ cube_cluster_start()
         /usr/sbin/hex_config trigger cluster_start
     fi
 
+    # Reapply pacemaker scheduler props now the cluster is up; the set in
+    # config_pacemaker SetupCluster (pacemaker_last) is lost if the CIB isn't
+    # ready yet, leaving single-node flooding STONITH errors + unbounded pe-error.
+    Quiet -n pcs property set stonith-enabled=false pe-error-series-max="1000" pe-warn-series-max="1000" cluster-recheck-interval="5min"
+
     if $OPENSTACK image list -f value | grep -q "Cirros active" ; then
         cmd rm -f /etc/glance/cirros-0.4.0-x86_64-disk.qcow2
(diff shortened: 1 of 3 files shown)


Eandalf-Bigstack (commented):
LGTM, waiting for pipeline job runs on pipeline prod and accept.

Eandalf-Bigstack (approved):

PR #1028, opened 2026-06-27
fix(set_ready): drop redundant neutron/octavia/kafka work + align csi image tags

traviswu-bigstack:
#### What type of PR is this?

/kind cleanup
/kind bug

#### What this PR does / why we need it

Removes redundant (and partly destructive) work in `cluster set_ready` / cluster bring-up, and fixes the ceph-csi image version drift that breaks `k3s_last`:

- **neutron** (`sdk_health.sh`): gate the full `bootstrap neutron` in `health_neutron_repair()` on an OVN/neutron agent-alive check — `set_ready` invoked it unconditionally twice (~148s).
- **octavia** (`config_octavia.cpp`): drop the duplicate `os_octavia_init` from the `cluster_ready` trigger; it still runs via `cluster_start`→`ReinitMain`, so cluster recycle is unaffected.
- **kafka** (`config_kafka.cpp`): remove the destructive `--delete`; `--create --if-not-exists` + `--alter` is idempotent and keeps topic data on re-apply.
- **csi** (`csi-drivers.go`): align all 4 image tags with `core/k3s/ceph-csi/Makefile` (snapshotter v8.2.0, resizer v1.11.1, provisioner v5.0.1, attacher v4.6.1); stale tags made `k3s_last` fail with `No such image`.

Net effect: `set_ready` ~26 min → ~21 min, no topic data loss on re-apply, and `k3s_last` completes with all csi images synced.

#### Which issue(s) this PR fixes

Fixes #1027

#### Special notes for your reviewer

> Note

Verified end-to-end on a single-node R630 via PXE/SoL — baseline (`a8d3efc`) vs patched A/B:

| Metric | Baseline | Patched |
|---|---|---|
| neutron full bootstrap | 2× (~148s) | 0× (agents Alive) |
| `os_octavia_init` | 2× | 1× |
| kafka `--delete` | 11 | 0 (topics retained) |
| csi sync | fails @ snapshotter v7.0.0 | all 6 push, `k3s_last` exits 0 |
| `set_ready` wall-time | ~26 min | ~21 min |

Recycle was validated via the documented `cluster poweroff` → boot auto-bootstrap (not `cluster stop`/`start`): octavia re-inits on the `cluster_start` trigger, graceful poweroff is clean from a healthy state, and `cluster check` returns `ok` for all 27 service groups. No regression (ceph HEALTH_OK, tenant instance ACTIVE).

[...]

Issue #1027, opened 2026-06-27
[Bug] set_ready: redundant neutron/octavia/kafka work + csi image version drift


traviswu-bigstack:
## Description

> A clear and concise description of what the bug is.

`cluster set_ready` (and cluster bring-up) does redundant and partly destructive work, and the k3s ceph-csi image sync ships stale image tags that break `k3s_last`:

1. **neutron — full bootstrap, twice.** `health_neutron_repair()` always runs `cmd $HEX_CFG bootstrap neutron` (~75s OVS/OVN/neutron bring-up); `set_ready` calls it at steps `[2/6]` and `[6/6]`, even when neutron is already healthy (~148s wasted).
2. **octavia — `os_octavia_init`, twice.** Registered on both the `cluster_ready` and `cluster_start` triggers, which `set_ready` fires back-to-back (~27s wasted).
3. **kafka — destructive `--delete` on every apply.** `RecreateTopic()` deletes each topic before `--create --if-not-exists`, dropping existing topic data on every re-apply.
4. **csi — image version drift (hard failure).** `csi-drivers.go` pins tags that no longer match the bundled tars in `core/k3s/ceph-csi/Makefile`, so `k3s_last` fails with `No such image` and the csi images never sync to the local registry.

## How to Reproduce

> Steps to reproduce the behavior:

1. Install a node and run `cluster set_ready <cidr> <gw> <range>`.
2. Inspect `/var/log/hex_config.log`:
   - `grep -c "neutron commit(o) took"` → **2**
   - `grep -c os_octavia_init` → **2**
   - `grep -c "kafka-topics.sh --delete"` → **11**
   - csi: `Failed to push sig-storage/csi-snapshotter image ... No such image: registry.k8s.io/sig-storage/csi-snapshotter:v7.0.0` (and the same for resizer/provisioner/attacher).
3. `docker images | grep csi-snapshotter` shows `v8.2.0` present, but cubectl tries to push `v7.0.0`.

## Expected behavior

> A clear and concise description of what you expected to happen.

- neutron is re-bootstrapped only when its OVN/neutron agents are actually unhealthy.
- `os_octavia_init` runs once per `set_ready`.
- kafka topics are ensured idempotently (`--create --if-not-exists` + `--alter`) without deleting existing data.
- `k3s_last` syncs all six csi images to the local registry and exits 0.

## Additional Context

> Add any other context about the problem here.

**Fix** (PR #1028):
[...]

PR #1055, opened 2026-07-01
fix(octavia): recover octavia-hm0 on non-master nodes after reboot/rolling-restart

traviswu-bigstack:
#### What type of PR is this?

/kind bug

#1028 dropped the per-control-node `os_octavia_init` from the `cluster_ready`
trigger, assuming `cluster_start`→`ReinitMain` covers it. But `ReinitMain` runs
`os_octavia_init` only on the master (`IS_MASTER`), so after a reboot or rolling
restart the **non-master** control nodes' `octavia-hm0` is never brought up: the
health-manager can't receive amphora heartbeats, load balancers go OFFLINE and
amphorae to ERROR. It passed single-node validation because a lone node is always
the master.

- add `os_octavia_hm0_up`: lightweight, idempotent ovs/ip bring-up of
  `octavia-hm0` from the existing neutron port (~seconds; no port recreate, no
  ReconfigMain, no service churn) — single source of the bring-up logic.
- `os_octavia_node_init`: delegate its bring-up to `os_octavia_hm0_up` (dedup).
- `_health_octavia_auto_repair`: on codes 5/6/7 run the lightweight bring-up on
  the affected node — fast enough to be overlap-safe on the periodic check.
- `health_octavia_repair`: run the heavy `reinit_octavia` on ALL first-three
  compute nodes (master first, rest in parallel), not just master + invoker.

Refs #1027

> This repairs the octavia-hm0 regression introduced by **#1028** (which closed
> **#1027** — its bullet 2 was the octavia `os_octavia_init` dedup). #1027 is
> closed, so linking with `Refs` rather than reopening it. Switch to `Fixes` if
> you'd prefer to reopen/track there.

> Two SDK files — `sdk_os.sh` (bring-up) + `sdk_health.sh` (auto-repair wiring).
> No image/registry changes, air-gapped safe. Overlap-safe on the periodic health
> check (lightweight path; heavy `reinit_octavia` stays deliberate).

#### Additional documentation

```docs
(octavia-hm0 recovery not yet captured in the handbook kb — follow-up)
```

### DIFF
--- a/core/sdk_sh/modules/sdk_health.sh
+++ b/core/sdk_sh/modules/sdk_health.sh
@@ -2203,7 +2203,15 @@ _health_octavia_auto_repair()
 
     for node in "${CUBE_NODE_COMPUTE_HOSTNAMES[@]}" ; do
         if remote_run $node hex_sdk is_first_three_compute_node ; then
-            if ! is_remote_running $node octavia-worker ; then
+            # octavia-hm0 missing/down/unrouted (codes 5/6/7): lightweight bring-up
+            # on that node -- fast + idempotent (no port recreate, no reconfig, no
+            # service churn), so it's overlap-safe on the periodic auto_repair. The
+            # heavy reinit_octavia stays in health_octavia_repair (deliberate path).
+            if ! remote_run $node ovs-vsctl port-to-br octavia-hm0 >/dev/null 2>&1 \
+               || remote_run $node ip link show octavia-hm0 2>/dev/null | grep -q DOWN \
+               || ! remote_run $node route -n 2>/dev/null | grep -q octavia-hm0 ; then
+                Quiet -n remote_run $node hex_sdk os_octavia_hm0_up
+            elif ! is_remote_running $node octavia-worker ; then
                 remote_systemd_restart $node octavia-worker
             elif ! is_remote_running $node octavia-health-manager ; then
                 remote_systemd_restart $node octavia-health-manager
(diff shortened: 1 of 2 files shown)


core/sdk_sh/modules/sdk_health.sh
```diff
@@ -2214,10 +2222,21 @@ _health_octavia_auto_repair()
...
+    # other first-three-compute node runs in parallel. reinit_octavia is heavy
+    # (~2min/node incl. ReconfigMain), but this is the serialized, operator/roll
+    # -triggered path -- it can't overlap a periodic round, so the cost is fine.
+    local master=$CUBE_NODE_CONTROL_HOSTNAMES node pids=""
     Quiet -n remote_run $master $HEX_CFG reinit_octavia
-    cmd $HEX_CFG reinit_octavia
+    for node in "${CUBE_NODE_COMPUTE_HOSTNAMES[@]}" ; do
+        [ "$node" = "$master" ] && continue
+        if remote_run $node hex_sdk is_first_three_compute_node ; then
+            Quiet -n remote_run $node $HEX_CFG reinit_octavia & pids="$pids $!"
```
Eandalf-Bigstack: Maybe we could split this into two lines to improve readability.

traviswu-bigstack: Done — split into two lines in the amended commit (b65be4f); also rebased onto latest develop.

Eandalf-Bigstack (approved):

Issue #1056, opened 2026-07-02
[User Story] fc_link — external FC / multipath link health in cluster check


traviswu-bigstack:
## Description

As a **CubeCOS operator**, I want **`cluster check` to report external Fibre Channel / multipath link health as a first-class `fc_link` component** so that **I can detect a degraded or lost SAN path (single-path, dead path group, stopped multipathd, flapping link) before it silently takes down Cinder/Ceph-backed storage — especially after a reboot or rolling restart.**

Today FC/multipath health is invisible to `cluster check`: a node can be running on a single surviving path (redundancy already gone) and every service group still reads `ok`. The next path loss is then a hard storage outage with no prior warning.

## Acceptance Criteria

- [x] `cluster check` shows an `fc_link` component in the **Storage** service group on control + compute nodes.
- [x] Detector is **live-state gated**: a node with no FC backend / no HBA / no FC LUNs reports `na` and stays healthy (no false positives on non-FC deployments or fresh reboots).
- [x] Worst-of aggregation across nodes reports the right code: `1` redundancy degraded, `2` path group down, `3` multipathd not running, `4` link errors / flapping.
- [x] The health module is surfaced through the cos-api health registry and rendered as **"FC Link"** in the cos-ui health view.
- [x] `cluster health Storage` / repair path exercises the fc_link detail + repair without disturbing other Storage components.
- [x] Air-gapped safe: no new install/runtime network fetches.

## Scope (cross-repo)

| Repo | Change |
|---|---|
| **cubecos** | `sdk_health.sh` fc_link detector + fan-out + backend gate + detail/repair; `errcodes` fc_link codes; `cli_cluster` registers fc_link in Storage (`R_CONTROL\|R_COMPUTE`) |
| **cube-cos-api** | add `fc_link` (repairable) to the health module registry |
| **cube-cos-ui** | add `FC Link` display label for the `fc_link` module |
| **bigstack-handbook** | `kb/cubecos/architecture/fc-link-health-check.md` topic |

## Output artifacts (Definition of Done)

- [x] **Handbook knowledge update** — `kb/cubecos/architecture/fc-link-health-check.md` (landing via the handbook PR in this set).

chase-chou:
# Test Result: ✅ PASS (TC-1/TC-2/TC-3/TC-5/TC-6 — TC-4 partial, TC-4 codes 2/4 pending)

**Environment:**

| Item | Value |
| :--- | :--- |
[...]

PR #1057, opened 2026-07-02
feat(health): add fc_link Fibre Channel storage link health check

traviswu-bigstack:
#### What type of PR is this?

/kind feature

Adds an `fc_link` component to `cluster check` that reports external Fibre Channel / multipath link health in the **Storage** service group on control + compute nodes. Today FC/multipath health is invisible to `cluster check`: a node can be running on a single surviving path (redundancy already gone) and every group still reads `ok`, so the next path loss is a hard storage outage with no warning.

- `sdk_health.sh`: `health_fc_link_status` (per-node detector, **live-state gated** so it is correct after reboot and never false-positives a node with no FC LUN), `_fc_scan_all` (bounded-parallel fan-out with a short file cache), `health_fc_backend_configured` (cinder FC-backend gate), plus detail/repair.
- `errcodes`: fc_link codes — `1` redundancy degraded, `2` path group down, `3` multipathd not running, `4` link errors/flapping (worst-of aggregation).
- `cli_cluster`: register `fc_link` in the Storage service group (`R_CONTROL | R_COMPUTE`).

Fail-safe: no FC backend / no HBA / no LUNs → `na` (node stays healthy). Air-gapped safe — no new install/runtime fetches.

Pairs with cube-cos-api (registry) and cube-cos-ui (label) PRs; handbook topic under `kb/cubecos/architecture/`.

Refs #1056

> Detector is deliberately report-only for wiring advisories (no HBA / no reachable target) so a node legitimately not wired for FC doesn't flip cluster health. Code 4 (bad SFP/cable) is intentionally **not** software-repairable — repair skips it and leaves it reported for hardware attention.


core/sdk_sh/modules/sdk_health.sh
```diff
@@ -1590,6 +1590,393 @@ health_rbd_target_repair()
...
+    fi
+    local node tmp running=0
+    tmp="$(mktemp -d /tmp/fcscan.XXXXXX)" || return 1
+    for node in $(_fc_link_nodes) ; do
+        { printf '%s\t%s\n' "$node" "$(remote_run "$node" "$HEX_SDK health_fc_link_status" 2>/dev/null)" \
+            > "$tmp/$node" ; } &
+        running=$(( running + 1 ))
+        [ $(( running % FC_SCAN_POOL )) -eq 0 ] && wait
+    done
+    wait
```
Eandalf-Bigstack, line 1645: Not quite sure why we are waiting in the loop at `[ $(( running % FC_SCAN_POOL )) -eq 0 ] && wait`.

traviswu-bigstack: Good catch — that line was a batch-style throttle: launch `FC_SCAN_POOL` scans, `wait` for the whole batch, then start the next batch. It bounds concurrency (so a large cluster does not get hit with hundreds of simultaneous ssh sessions), but one slow node gates the entire batch, and as your comment shows the modulo idiom is not obvious.

Reworked in 6dac361 as a proper sliding-window pool using `wait -n` (bash ≥4.3; hex_sdk runs under bash 5.1): when the pool is full, wait for *any one* scan to finish, then launch the next — keeps `FC_SCAN_POOL` scans in flight continuously with no batch-boundary stalls. Also rebased onto latest develop.

Eandalf-Bigstack: Thanks for the detailed explanation.

Eandalf-Bigstack (approved):

PR #1085, opened 2026-07-07
fix: nfs-ganesha, cinder-backup & telegraf auto-recover after a maintenance cycle

### LINKED ISSUE #1084: [Bug] Services stay NG after a maintenance cycle instead of auto-recovering
## Description

After a `cluster powercycle`/reboot, several services stayed NG and needed a manual restart instead of self-healing via the periodic health auto-repair. Investigation found three per-service triggers and one systemic trap they all hit.

## How to Reproduce

1. Healthy cluster.
2. `cluster powercycle` (full-cluster reboot).
3. Observe services that come up racing their dependencies stay NG and never recover:
   - `Storage` → `ceph_mds(3 nfs-ganesha fail to mount)`
   - `BlockStor` → `cinder(5 backup down)`
   - `Metrics` → `telegraf(1 daemon down)`

## Expected behavior

All self-heal via the periodic auto-repair once their dependencies are up.

## Additional Context

Root causes:
- **nfs-ganesha**: 5.9 SEGVs on the SIGTERM stop → flaps → burns the auto-repair `maxerr` budget.
- **cinder-backup**: `health_cinder_check` fed the auto-repair `ceph health detail` instead of the down-service list → it matched nothing.
- **telegraf**: boots before kafka → `Connect()` fails → exits → trips systemd `StartLimitBurst` → burns `maxerr`.
- **systemic (the trap)**: `_health_fail_log` gave up auto-repair *permanently* once `count ≥ maxerr` — the count only resets on a passing check, impossible while the service is down. So anything transiently down at boot stayed stranded even after its dependency recovered.

Fixed in PR #1085 (SIGKILL drop-in, cinder ERR_MSG feed, telegraf restart drop-in, and `maxerr` backoff with a unit test). Verified across two full powercycles on the sky HA cluster.

A related finding — the master never marking bootup `done` — ships separately with the boot-time work (`fix/master-bootup-done`), since it touches code not in `develop`. See `kb/cubecos/known-issues/master-stuck-finalizing-no-done.md`.

- [x] **Handbook knowledge update** — covered: [cinder-backup-auto-repair-fed-wrong-errmsg](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/cinder-backup-auto-repair-fed-wrong-errmsg.md) + [nfs-ganesha-segv-on-stop](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/nfs-ganesha-segv-on-stop.md)

traviswu-bigstack:
#### What type of PR is this?

/kind bug

After a `cluster powercycle`/reboot, several services stayed NG and needed a manual restart instead of self-healing via the periodic health auto-repair. Four fixes so they recover on their own — three per-service *triggers* plus the systemic *trap* behind them:

**1. nfs-ganesha SIGKILL drop-in** (`core/ceph/nfs-ganesha-sigkill.conf` + `ceph.mk`)
ganesha 5.9 SEGVs on the SIGTERM stop, flapping it until it burns the auto-repair `maxerr` budget → `ceph_mds` stays NG. Hard-kill on stop → clean stop, no flap.

**2. cinder-backup auto-repair** (`sdk_health.sh`)
`health_cinder_check` fed `_health_cinder_auto_repair` `ceph health detail` (no `enabled … down` lines) instead of the down-service list, so it never restarted the down service. Feed it the service list (+ `Host` column).

**3. telegraf restart drop-in** (`core/telegraf/telegraf-restart.conf` + `telegraf.mk`)
telegraf's kafka output `Connect()` fails at init when it boots before kafka → exits 1 → trips `StartLimitBurst` → burns `maxerr`. A restart drop-in (`StartLimitIntervalSec=0`, `RestartSec=15`) retries in the background until kafka is up. `Type=simple`, so it does **not** block the `node_start` commit (an `ExecStartPre` wait would have).

**4. maxerr backoff — the systemic fix** (`sdk_health.sh` + unit test)
`_health_fail_log` gave up auto-repair permanently once `count ≥ maxerr` (the count only resets on a passing check, impossible while the service is down). So anything transiently down at boot stayed stranded even after its dependency recovered. Now it **backs off** past `maxerr` (retry every Nth round via the new `_health_repair_due` predicate, tunable `_HEALTH_REPAIR_BACKOFF`) instead of stopping. Unit-tested in `core/sdk_sh/tests/`.

Verified end-to-end on the sky HA cluster across two full powercycles: with all fixes, every service group returns to `ok` (bar an unrelated `fc_link` hardware baseline).

Fixes #1084

> Fixes 1–3 remove the per-service triggers; fix 4 removes the trap itself, so a service transiently down at boot recovers regardless of what stranded it. A **fifth** fix from the same investigation — the master never marking bootup `done` (`fix/master-bootup-done`) — is **not** in this PR: it edits `power_bootup_mark`/`cluster_check_repair_async`, which exist only on the boot-time branch, not `develop`. It ships with the boot-time work.

#### Additional documentation

```docs
kb/cubecos/known-issues/nfs-ganesha-segv-on-stop.md
kb/cubecos/known-issues/cinder-backup-auto-repair-fed-wrong-errmsg.md
kb/cubecos/known-issues/telegraf-crashloop-kafka-not-ready.md
kb/cubecos/known-issues/health-autorepair-gives-up-at-maxerr.md
```

### DIFF
--- a/core/ceph/ceph.mk
+++ b/core/ceph/ceph.mk
@@ -40,6 +40,8 @@ rootfs_install::
 	#$(Q)mv $(ROOTDIR)/lib/udev/rules.d/95-ceph-osd.rules $(ROOTDIR)/lib/udev/disabled/.
 	$(Q)rm -f $(ROOTDIR)/etc/ganesha/*
 	$(Q)$(INSTALL_DATA) $(ROOTDIR) $(COREDIR)/ceph/ganesha.conf ./etc/ganesha/
+	$(Q)chroot $(ROOTDIR) mkdir -p /etc/systemd/system/nfs-ganesha.service.d
+	$(Q)$(INSTALL_DATA) $(ROOTDIR) $(COREDIR)/ceph/nfs-ganesha-sigkill.conf ./etc/systemd/system/nfs-ganesha.service.d/
 	$(Q)chroot $(ROOTDIR) mkdir -p /etc/cube/cos/cron
 	$(Q)chroot $(ROOTDIR) mkdir -p /etc/cube/cos/ceph
 
--- a/core/ceph/nfs-ganesha-sigkill.conf
+++ b/core/ceph/nfs-ganesha-sigkill.conf
@@ -1,0 +1,10 @@
+# Upstream nfs-ganesha 5.9 SEGVs on the SIGTERM systemd sends to stop it, which
+# leaves the unit core-dumped/failed and flaps ganesha on every maintenance cycle
+# -- burning the health auto-repair maxerr budget so ceph_mds stays NG. Hard-kill
+# on stop (no catchable SIGTERM -> no crashing shutdown handler) and treat the
+# kill as a clean stop, so ganesha stops cleanly and the ~63s auto-repair recovers
+# it. See kb/cubecos/known-issues/nfs-ganesha-segv-on-stop.md.
+[Service]
+ExecStop=
+KillSignal=SIGKILL
+SuccessExitStatus=SIGKILL
--- a/core/ceph/nfs-ganesha-sigkill.conf
+++ b/core/ceph/nfs-ganesha-sigkill.conf
@@ -1,9 +1,5 @@
-# Upstream nfs-ganesha 5.9 SEGVs on the SIGTERM systemd sends to stop it, which
-# leaves the unit core-dumped/failed and flaps ganesha on every maintenance cycle
-# -- burning the health auto-repair maxerr budget so ceph_mds stays NG. Hard-kill
-# on stop (no catchable SIGTERM -> no crashing shutdown handler) and treat the
-# kill as a clean stop, so ganesha stops cleanly and the ~63s auto-repair recovers
-# it. See kb/cubecos/known-issues/nfs-ganesha-segv-on-stop.md.
+# Hard-kill on stop: nfs-ganesha 5.9 SEGVs on SIGTERM, flapping it and burning the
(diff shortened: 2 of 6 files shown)


traviswu-bigstack:
Added a third fix under the same theme (services not auto-recovering after a maintenance cycle):

**3. telegraf survives kafka-not-ready at boot** (`core/telegraf/telegraf-restart.conf` + `telegraf.mk`)
telegraf's kafka output `Connect()` fails at agent init if kafka's brokers aren't reachable yet — at boot its `node_start` commit can run before kafka, so it exits 1, trips systemd `StartLimitBurst`, and the health auto-repair then burns its `maxerr` budget and gives up → `Metrics NG (telegraf daemon down)`. A systemd drop-in (`StartLimitIntervalSec=0`, `RestartSec=15`) keeps restarting it in the background until kafka comes back. `Type=simple` → the `node_start` commit’s `systemctl restart` returns immediately, so this does **not** block the boot commit (an `ExecStartPre` wait-for-kafka would have — telegraf starts inside the commit sequence).

Verified on the sky HA cluster after a full powercycle: drop-in in effect on all 3 nodes, telegraf brought up and stable (`NRestarts=0`), `Metrics` green.

kb: `kb/cubecos/known-issues/telegraf-crashloop-kafka-not-ready.md`

Note: this fixes telegraf’s *trigger*; the systemic **maxerr permanent give-up** (the real reason a transiently-down service stays stranded) is a separate follow-up — auto-repair should back off past `maxerr`, not stop forever.

Eandalf-Bigstack (approved):

PR #1087, opened 2026-07-07
fix(health): check node clock time differences in parallel

### LINKED ISSUE #1086: [Bug]: health_clock_check Taking Too Long to Traverse 10+ Nodes
## Describe the bug

When we have more than 10 nodes in a cluster, health check `health_clock_check` would take too long to traverse through all nodes and report the latency as the time difference. Thus, it would create health check errors.

## Version 

What version of the product is affected [e.g. 3.x.x].

- v3.1.0

## To Reproduce

1. Have more than 10 nodes in a cluster
2. Health check item `clock` would keep reporting errors

## Expected behavior

1. Health check item `clock` should be OK if there is no real time differences between nodes

## Additional context

- Hardware [e.g. CPU, RAM, Disk Controller, Disks]
- Supporting logs

## QA Verification

> How QA verifies this issue is actually fixed — written as **checkable claims**: the exact
> command plus the expected outcome. Run everything from a control node as `root`.
> Fixed by #1087 (tip commit `fe385f8`, fast-forward merged to `develop` on 2026-07-08).

The fix is **not** in `health_clock_check` — that function is byte-identical to before. It is in
`cmd()`, which `health_clock_check` uses to collect each node's clock:

- `cmd()` called `is_sshable <node>` in the **foreground** of its fan-out loop and only
  backgrounded the `ssh` after it returned. On a cold cache `is_sshable` sleeps **≥1 s per
  node** (it polls a marker 3 × 1 s), so the fan-out cost ≥1 s × N *before* the last node's
  `date +%s` ran. At 10+ nodes the first and last readings were >10 s apart, and
  `health_clock_check` fails when `max − min > 10` — a false positive with perfectly
  synchronised clocks.
- `fe385f8` moves `is_sshable` **inside** the backgrounded block, so all nodes run
  concurrently. It also fixes two latent bugs in the same function: the per-node temp log used
  `${node}` (a stale outer variable) instead of `${_cmd_node}`, and `_cmd_node_log` was not
  declared `local` in the `-o`/one-by-one path.
- An interim commit in the same PR (`f759e38`) had swapped the check over to
  `cubectl node exec -np`; `fe385f8` **reverted** that and fixed `cmd()` instead. So
  `health_clock_check` still reads `cmd -v "date +%s"` — that is correct, not a missed revert.

### Preconditions

- A cluster with **10 or more nodes** for the headline case. Steps 3–5 are meaningful from 3
  nodes up, because they compare parallel against serial directly.
- **Clear the `is_sshable` cache before every timing measurement:**
  `rm -f /run/is_sshable_*`. The marker `/run/is_sshable_<node>` is treated as valid for 1
  minute; with warm markers `is_sshable` returns instantly and **both the old and the fixed
  code look fast**. A test that skips this will pass on unfixed code.

### Checks

|---|---|---|
| 1 | `grep -c 'Wrapping is_sshable' /usr/lib/hex_sdk/modules.pre/sdk_cmd.sh` | `1` — the build carries the fix. `0` means this node predates it; stop here. |
| 2 | `grep -n 'local times=' /usr/lib/hex_sdk/modules/sdk_health.sh` | `local times=$(cmd -v "date +%s" \| cut -d"\|" -f3 \| sed "/^$/d" \| sort)` — i.e. `cmd`, **not** `cubectl node exec` |
| 3 | `hex_sdk cmd -av "hostname" \| wc -l` | the cluster's node count (record it as `N` for step 4) |
| 4 | `rm -f /run/is_sshable_*; time hex_sdk cmd -av "date +%s"` | one `<node>\|0\|<epoch>` line per node, and `real` **roughly constant, ~2–4 s regardless of N** — emphatically not ~1 s × N. On a 12-node cluster: ~3 s, where the old code took ~13 s. |
| 5 | `rm -f /run/is_sshable_*; time hex_sdk cmd -aov "date +%s"` | `real` **scales with N** (≳1 s per node). `-o` is one-by-one and still serial by design — this is the control that proves step 4 is measuring parallelism and not a warm cache. |
| 6 | `rm -f /run/is_sshable_*; hex_sdk cmd -av "date +%s" \| cut -d'\|' -f3 \| sed '/^$/d' \| sort -n \| awk 'NR==1{m=$1} {M=$1} END{print "min="m, "max="M, "delta="M-m}'` | `delta=0` or `1`; **must be ≤ 10**. This is the exact quantity the check thresholds. |
| 7 | `hex_sdk health_clock_report` | a table reading `status healthy`, `error_code 0`, `description ok` |
| 8 | `hex_sdk -f json health_clock_report` | `{"status" : "healthy","error_code" : "0","description" : "ok"}` |
| 9 | `cat /run/health_clock_check` | `{"SERVICE" : "clock","ERROR_CODE" : "0","DESCRIPTION" : "ok"}` |
| 10 | `ls /tmp/cmd_* 2>/dev/null \| wc -l` | `0` — `cmd()` removes each per-node temp log; covers the `${node}` → `${_cmd_node}` fix |

### Sustained run — the original symptom was "keeps reporting errors"

```bash
rm -f /tmp/health_clock_error.count
for i in $(seq 5) ; do hex_sdk health_clock_check ; echo "run $i rc=$?" ; sleep 60 ; done
ls /tmp/health_clock_error.count 2>/dev/null || echo "no failure counter -> all runs passed"
```

Expected: `rc=0` on all five runs, and no `/tmp/health_clock_error.count`. The `sleep 60` is
deliberate — it lets the `is_sshable` markers expire so every round runs the cold-cache worst
case, which is what the periodic health check actually hits.

Also confirm nothing new was appended:

```bash
tail -20 /var/log/health_clock_error.log
```

Expected: no block dated during this test window. (Blocks from before the fix are fine and
expected — do not treat old entries as a failure.)

### Negative test — real clock drift must still be caught

The fix removes a false positive; it must not remove the true positive. On a **non-control**
node:

```bash
ssh root@<node> 'systemctl stop chronyd ; date -s "+30 seconds"'
hex_sdk -v health_clock_check ; echo "rc=$?"
```

Expected: `rc=1` and `Biggest time gap between nodes is 30s` (±1 s).

Then repair and re-check:

```bash
hex_sdk health_clock_repair          # runs ntp_makestep (restart chronyd + hwclock -w) on every node
ssh root@<node> 'systemctl start chronyd'
hex_sdk -v health_clock_check ; echo "rc=$?"
```

Expected: `rc=0`, and `/run/health_clock_check` back to `"ERROR_CODE" : "0"`.

- [x] **QA Status declared at Done** — set the board's `QA Status` field to `Not needed` or
      `Ready to QA`. Never leave it empty on a Done ticket; `In QA` / `Verified` are QA's own
      transitions.

Eandalf-Bigstack:
#### What type of PR is this?

- fix

- Check node clock time differences in parallel
  - The original implementation leveraged `cmd()`. However, the underlying SSH handshakes for calling to each remote node took around 1 second. This could pile up and exceed the threshold 10 seconds if we have more than 10 nodes in a cluster.

- Fixes #1086

### DIFF
--- a/core/sdk_sh/modules/sdk_health.sh
+++ b/core/sdk_sh/modules/sdk_health.sh
@@ -276,7 +276,9 @@ health_clock_report()
 
 health_clock_check()
 {
-    local times=$(cmd -v "date +%s" | cut -d"|" -f3 | sed "/^$/d" | sort)
+    # this is a temporary fix since we are moving away cubectl in favor of cmd()
+    # however, we do not have real parallelism in cmd()
+    local times=$(cubectl node exec -np date "+%s" | sort)
     local timea=$(echo "$times" | head -1)
     local timeb=$(echo "$times" | tail -1)
     local delta=$(expr ${timeb:-0} - ${timea:-0})
--- a/core/sdk_sh/modules/sdk_health.sh
+++ b/core/sdk_sh/modules/sdk_health.sh
@@ -276,9 +276,7 @@ health_clock_report()
 
 health_clock_check()
 {
-    # this is a temporary fix since we are moving away cubectl in favor of cmd()
-    # however, we do not have real parallelism in cmd()
-    local times=$(cubectl node exec -np date "+%s" | sort)
+    local times=$(cmd -v "date +%s" | cut -d"|" -f3 | sed "/^$/d" | sort)
     local timea=$(echo "$times" | head -1)
     local timeb=$(echo "$times" | tail -1)
     local delta=$(expr ${timeb:-0} - ${timea:-0})
--- a/core/sdk_sh/modules.pre/sdk_cmd.sh
+++ b/core/sdk_sh/modules.pre/sdk_cmd.sh
@@ -88,13 +88,18 @@ cmd()
     fi
     if [ "x$onebyone" = "xfalse" ] ; then
         for _cmd_node in "${_cmd_nodes[@]}" ; do
-            _cmd_nodes_logs[$_cmd_node]="$(mktemp -u /tmp/cmd_${node}.XXXX)"
+            # Minor fix: changed ${node} to ${_cmd_node} to correctly map the loop variable
+            _cmd_nodes_logs[$_cmd_node]="$(mktemp -u /tmp/cmd_${_cmd_node}.XXXX)"
+
             if [ "x$dryrun" = "xfalse" ] ; then
-                if is_sshable $_cmd_node ; then
-                    timeout $_ssh_to ssh $_ssh_opts $_cmd_node "$(typeset -f ${1%% *} || echo true ); VERBOSE=$VERBOSE; $@" >${_cmd_nodes_logs[$_cmd_node]} 2>&1 &
-                else
-                    false &
-                fi
+                # Wrapping is_sshable into the backgrounded execution block
+                {
+                    if is_sshable $_cmd_node ; then
+                        timeout $_ssh_to ssh $_ssh_opts $_cmd_node "$(typeset -f ${1%% *} || echo true ); VERBOSE=$VERBOSE; $@"
+                    else
+                        false
+                    fi
+                } >"${_cmd_nodes_logs[$_cmd_node]}" 2>&1 &
             else
                 echo "ssh -o LogLevel=quiet $_cmd_node $@" &
             fi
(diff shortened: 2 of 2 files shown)


core/sdk_sh/modules/sdk_health.sh
```diff
@@ -265,7 +265,9 @@ health_clock_report()
 
 health_clock_check()
 {
-    local times=$(cmd -v "date +%s" | cut -d"|" -f3 | sed "/^$/d" | sort)
+    # this is a temporary fix since we are moving away cubectl in favor of cmd()
+    # however, we do not have real parallelism in cmd()
+    local times=$(cubectl node exec -np date "+%s" | sort)
```
traviswu-bigstack: cmd() is using a parallel path by default. What makes it behave serially here is the per-node is_sshable pre-flight in the launch loop (sdk_cmd.sh): it runs in the foreground before each node's job is spawned. move is_sshable inside the backgrounded subshell — ( is_sshable $n && timeout $_ssh_to ssh … ) & — so cmd() is parallel end-to-end; then this check can return to cmd -v and the cubectl exception goes away. 

What do you think?

Eandalf-Bigstack: Yes, the is_sshable is the bottleneck. Let me try to move it into background also.

traviswu-bigstack (approved):

PR #1076, opened 2026-07-06
fix(cluster): make rolling-restart status resilient to failover, wedge, and phase loss

traviswu-bigstack:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it

Makes rolling-restart status tracking survive real-world conditions (stacked on #1058, first of a 4-PR series):

- **Locked, retried job.json writes** — `_power_roll_write` wraps every update in `flock` with bounded retry (30×2 s) across a cephfs/MDS failover window; unique temp names prevent lost updates from concurrent writers. All setters route through it.
- **Truthful phases + per-phase timing** — `phase_ts` stamped on every status change; `status` gains a per-phase duration breakdown table; new `status_watch` (30 s redraw). A node only becomes "rebooting" once it actually drops off the network; "bootstrapping" is inferred when it pings again; "finalizing" (status advanced by `bootstrap_cube_config` right before `cube_cluster_start` — the single, traceable place for boot-phase transitions) is reported with its own progress percent in the JSON status consumed by the UI.
- **Abort that works** — `abort` now also applies to a *running* roll (wedged/zombie driver), re-enables the in-flight node's nova-compute and clears `inflight`; `continue` still requires paused.

Part of the cluster boot-time reduction initiative (~45 → ~20 min); lab-validated on the R630 cluster.

#### Which issue(s) this PR fixes

Fixes #1072

#### Special notes for your reviewer

> Stack: #1058 ← **this** ← bootup-status ← vm-restore ← autorepair-robustness. Review/merge in that order; each later PR's diff assumes the earlier ones.

#### Additional documentation

```docs
Handbook: kb/cubecos/architecture/rolling-restart-*.md (resilience notes; tracking link added).
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01So64xm2uahGXMhX7JLT8M1

PR #1092, opened 2026-07-07
fix(cluster): record the rolling-restart 'bootstrapping' phase timestamp

### LINKED ISSUE #1091: [Bug] rolling_restart status shows a blank 'bootstrapping' time
## Description

`hex_cli -c cluster -c rolling_restart -c status` shows an empty `bootstrapping` column for every node, and an inflated, never-ending `rebooting` time (e.g. `rebooting 33m40s*` on a node whose total is only `12m37s`). `cluster bootup status` shows the same phase correctly.

## How to Reproduce

1. `hex_cli -c cluster -c rolling_restart` on an HA cluster.
2. `... rolling_restart -c status` while/after it runs — `bootstrapping` is blank, `rebooting` grows unbounded with a `*`.

## Expected behavior

`bootstrapping` shows a real duration, and `rebooting` ends when bootstrapping begins.

## Additional Context

Root cause: the phase state (`job.json`) lives on **cephfs** (`/mnt/cephfs/rolling_restart/job.json`), and `_power_roll_set_phase_ts` guards on `[ -e "$ROLLING_RESTART_JOB" ] || return 0`. The `bootstrapping` stamp fires **before** the recovering node’s config commit mounts cephfs, so the write is **silently dropped**; `finalizing`/`done` run after the mount and record fine. The display computes `rebooting = bootstrapping_ts − rebooting_ts`, so a missing `bootstrapping_ts` leaves `rebooting` open-ended. `cluster bootup status` is unaffected because it uses a local `/run` file. Latent since the phase-timing feature was introduced (26310cf / #1076); not a regression.

Fixed in PR: capture the timestamp, then write directly if cephfs is up, else delegate to an up control peer, else spool to `/run` and replay on the next cephfs-up call (guarded so a peer failure can’t drop the stamp or kill boot). Unit-tested.

- [x] **Handbook knowledge update** — covered: [rolling-restart-bootstrapping-time-blank](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/rolling-restart-bootstrapping-time-blank.md)

traviswu-bigstack:
#### What type of PR is this?

/kind bug

`rolling_restart status` showed a blank `bootstrapping` column and an inflated, never-ending `rebooting` time.

**Root cause.** The phase state (`job.json`) lives on cephfs, and `_power_roll_set_phase_ts` guards on `[ -e "$ROLLING_RESTART_JOB" ] || return 0`. The `bootstrapping` stamp fires early in a recovering node’s boot — **before its config commit mounts cephfs** — so the write is silently dropped. `finalizing`/`done` run after the mount and record fine. Because the display derives `rebooting = bootstrapping_ts − rebooting_ts`, a missing `bootstrapping_ts` leaves `rebooting` open-ended. (`cluster bootup status` is unaffected — it uses a local `/run` file.)

**Fix.** Capture the real timestamp, then:
1. write it directly if cephfs is up (also flushing any buffered stamps);
2. else delegate to an up control peer that has cephfs — **in a subshell**, so a peer that Errors/`exit`s via `remote_run` can’t kill the boot script;
3. else spool locally to `/run` and replay on the next cephfs-up call.

The stamp is never dropped and never fatal. Unit-tested in `core/sdk_sh/tests/test_power_roll_phase_ts.sh` (direct / delegated / buffered / delegate-failure branches).

Latent since the phase-timing feature (26310cf / #1076) — not a regression.

Fixes #1091

> Surfaced while validating the boot-time work on the sky HA cluster. Independent of the storage-recovery (#1085) and master-bootup-done (#1090) fixes. The delegate path deliberately avoids the bare `remote_run → Error → exit` hazard (the same one #1090 addresses) by isolating it in a subshell.

#### Additional documentation

```docs
kb/cubecos/known-issues/rolling-restart-bootstrapping-time-blank.md
```

### DIFF
--- a/core/sdk_sh/modules/sdk_power.sh
+++ b/core/sdk_sh/modules/sdk_power.sh
@@ -46,12 +46,56 @@ _power_roll_set_node_status()
         '(.nodes[]|select(.hostname==$h)) |= (.status=$s | .phase_ts=((.phase_ts//{}) + {($s):$ts}))'
 }
 
-# Stamp a phase timestamp without changing .status. No-op if no job exists.
+# The raw phase_ts write. Also the entry point a peer or a spool-replay uses:
+# `hex_sdk _power_roll_write_phase_ts <host> <phase> <epoch>`.
+_power_roll_write_phase_ts()
+{
+    [ -e "$ROLLING_RESTART_JOB" ] || return 1
+    _power_roll_write --arg h "$1" --arg p "$2" --argjson ts "$3" \
+        '(.nodes[]|select(.hostname==$h)) |= (.phase_ts=((.phase_ts//{}) + {($p):$ts}))'
+}
+
+# Replay locally-buffered stamps to job.json once cephfs is reachable.
+_power_roll_flush_phase_ts_spool()
+{
+    local spool=$1 bh bp bts
+    [ -e "$spool" ] || return 0
+    while read -r bh bp bts ; do
+        [ -n "$bh" ] && _power_roll_write_phase_ts "$bh" "$bp" "$bts"
+    done < "$spool"
+    rm -f "$spool"
+}
+
+# First reachable control peer (not self) -- it has cephfs, so it can stamp for us.
+_power_roll_ts_peer()
+{
+    local self=$(hostname) n
+    for n in ${MASTER_CONTROL:-} $(cubectl node list -r control -j 2>/dev/null | jq -r '.[].hostname' 2>/dev/null) ; do
+        [ "$n" = "$self" ] && continue
+        is_sshable "$n" 2>/dev/null && { echo "$n" ; return 0 ; }
+    done
+}
+
+# Stamp a phase timestamp without changing .status. job.json lives on cephfs, which
+# isn't mounted early in a node's own boot -- so capture the real ts now, then write
+# it directly (cephfs up), else delegate to an up peer that has cephfs, else spool
+# locally and flush on a later cephfs-up call. Never fatal: the delegate runs in a
+# subshell so a peer that Errors/exits (remote_run) can't kill the boot script.
 _power_roll_set_phase_ts()
 {
-    [ -e "$ROLLING_RESTART_JOB" ] || return 0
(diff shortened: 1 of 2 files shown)


core/sdk_sh/modules/sdk_power.sh
```diff
@@ -46,14 +46,58 @@ _power_roll_set_node_status()
...
+
+# First reachable control peer (not self) -- it has cephfs, so it can stamp for us.
+_power_roll_ts_peer()
+{
+    local self=$(hostname) n
+    for n in ${MASTER_CONTROL:-} $(cubectl node list -r control -j 2>/dev/null | jq -r '.[].hostname' 2>/dev/null) ; do
+        [ "$n" = "$self" ] && continue
+        is_sshable "$n" 2>/dev/null && { echo "$n" ; return 0 ; }
+    done
+}
```
Eandalf-Bigstack, line 77: It might not guarantee CephFS working if we follow the list of control nodes and test if it is reachable solely.

traviswu-bigstack: 1. This is the rolling-restart path — one node reboots at a time while the other control nodes stay up and intact. So a peer that answers SSH almost always has a healthy cephfs too; is_sshable is good enough for this case in practice.
 2. Why delegate at all: the stamp is taken at the very start of the recovering node's bootstrap, before its config commit mounts cephfs — so the node can't write job.json itself yet. A peer that already has cephfs records the timestamp on its behalf.
 3. If SSH really doesn't work out (no reachable peer, or the delegate fails), the stamp isn't lost: it's buffered to a local spool file and written to cephfs once this node's cephfs comes up after bootstrapping. So the peer path is best-effort "write it sooner"; the spool is the backstop.

Eandalf-Bigstack: Got it. Thanks for the explanation. That would provide a guard to ensure CephFS is working for writing the shared progress file.

Eandalf-Bigstack (approved):

PR #1100, opened 2026-07-10
fix(ceph): render mon host from the live monmap on every node's boot

### LINKED ISSUE #1099: [Bug] ceph.conf mon host lists only the master mon — master death blackholes all new ceph clients and blocks host-failure evacuation
## Description

Every node's `/etc/ceph/ceph.conf` carries `mon host = <master mon IP>` — a **single address** — so when the master control node goes down, every *new* ceph client on the surviving nodes hangs bootstrapping against the dead IP even though the mon daemons keep perfect quorum. The blast radius is total loss of host-failure recovery:

- `ceph -s` / `ceph df` hang (~45s+ per attempt, up to `client_mount_timeout`).
- nova-compute blocks inside `rbd_utils.get_pool_info()` (`ceph df` subprocess) while holding the resource-tracker lock → service heartbeats stop → both survivors go `down` for the scheduler → `ComputeFilter returned 0 hosts` → `NoValidHost` → **masakari host-failure evacuation is structurally impossible cluster-wide**.
- Long-running qemu processes are immune (they learned the full monmap at connect time), which is why this stayed hidden until a real host-failure test.

Root cause is a path regression, not a missing feature:
- The per-commit render (`config_ceph.cpp` `Commit()` → `UpdateConfig`) writes only the master's mon IP — by design for first bootstrap, but it runs on **every** config commit/boot and regresses the file each time.
- `hex_config sync_ceph_config` renders the correct full monmap list, and used to run on every node (pre-refactor `cube_cluster_start` fired the `cluster_start` trigger unconditionally per node). Since the workload-preserving rolling-restart split, the trigger is **master-only, run-once** (`cube_cluster_start_cluster_if_master`), so non-master nodes never converge. The only other caller is the manual CLI `storage > sync_config`.
- Bonus hazard: `ceph_mon_map_iplist` returns the literal string `1` on failure; an unguarded sync would write `mon host = 1`.

## How to Reproduce

1. Healthy 3-control cluster; check any non-master node: `grep 'mon host' /etc/ceph/ceph.conf` → single master IP (any number of power cycles / rolling restarts does not change this).
2. Power off the master (chassis off).
3. On a survivor: `timeout 45 ceph -s` → hangs/nothing, while `ceph daemon mon.<host> mon_status` shows healthy quorum of the remaining mons.
4. Within minutes nova-compute heartbeats stop (`openstack compute service list` → survivors `down`); any evacuation attempt → `NoValidHost`.

Observed live on the sky lab 2026-07-10 during a masakari host-failure test (VIP-holder master powered off): evacuation was blocked ~35 minutes until `mon host` was hand-edited to the full list, after which ceph clients answered instantly (HEALTH_WARN 1/3 mons down, quorum 2/3) and the same 9-VM evacuation completed in under a minute.

## Expected behavior

Ceph clients on every node can reach the mon quorum whenever quorum exists — regardless of which node is down. `mon host` lists all mons on every node, from first boot onward, surviving every reboot.

Fix (validated on sky, all 3 nodes): run `sync_ceph_config` in the per-node boot half (`cube_cluster_start_node`) so each node re-renders `mon host` from the live monmap on every boot (the sdk already falls back to extracting the local mon store when quorum is unreachable), and guard `sync_ceph_config` against writing an empty/invalid list. The `[mds.<host>] mon host` line rendered from `ceph_mds_map_iplist` is pre-existing scoped behavior and untouched.

Related (separate issues): VIP-failover half-dead client sockets (keystoneauth/oslo.db timeouts + `vaw` sweep; `vaw` pacemaker resource is also missing on clusters set up before it existed); FC fabric degradation on sky.

- [x] **Handbook knowledge update** (landed: `kb/cubecos/known-issues/ceph-mon-host-single-ip-master-death.md`, bigstack-handbook#70) — land the durable, team-readable knowledge from this

traviswu-bigstack:
#### What type of PR is this?

/kind bug

Every node's config commit renders `mon host = <master mon IP>` — one address. The full-monmap renderer (`sync_ceph_config`) exists, but since the rolling-restart split it is reached only via the master-only, run-once `cluster_start` trigger (or the manual `storage > sync_config` CLI), so non-master nodes never converge and every boot regresses their conf back to the single master IP.

Consequence (hit live during a host-failure test): with the master chassis off, every **new** ceph client on the survivors hangs bootstrapping against the dead IP despite healthy 2/3 mon quorum — `ceph df` blocks nova-compute's resource tracker, heartbeats stop, both survivors go scheduler-`down`, and host-failure evacuation returns `NoValidHost` cluster-wide. Running qemu processes are unaffected (they hold the learned monmap), which kept this invisible until a real host death.

Fix:
- `cube_cluster_start_node`: run `sync_ceph_config` on **every node's boot**, restoring the pre-refactor per-node convergence (the sdk falls back to extracting the local mon store when quorum is unreachable).
- `ceph_mon_map_iplist` / `ceph_mon_map_hosts`: return empty output on failure instead of the literal `1` — a bash `${var:-1}` default-expansion copied from the file's arithmetic contexts; every caller (sync, the rejoin path, `ceph_mds_map_hosts`) consumed the sentinel as real data, e.g. writing `mon host = 1`.
- `sync_ceph_config`: additionally skip the rewrite unless the monmap yields something address-shaped, so a failed monmap fetch never clobbers a working conf (defense in depth over the sdk fix).

Validated on sky141/142/143: `sync_ceph_config` renders the full v2/v1 monmap list per node and ceph clients answer instantly with a mon down; during the incident, the single-IP → full-list edit turned a ~35-minute evacuation blackout into a sub-minute evacuation.

Fixes #1099

- First-bootstrap behavior is unchanged: the commit path still seeds the single master IP before a monmap exists; the per-node sync then converges the conf as soon as a monmap is available and keeps it converged on every boot.
- Transition note for existing clusters: the first post-upgrade boot syncs from a conf that still lists only the master — fine while the master is alive; on control nodes the local-store fallback covers the master-down case.
- The `[mds.<host>] mon host` line from `ceph_mds_map_iplist` is pre-existing, section-scoped behavior and deliberately untouched.
- The `CONTROL_REJOIN` fetch (`config_ceph.cpp:1508`) also consumed the old sentinel unvalidated; with the sdk fix it now gets empty output on failure — still a failing rejoin, but no longer writes garbage. A validation there is a possible follow-up.

#### Additional documentation

```docs
Each node re-renders ceph.conf's `mon host` from the live monmap at every boot
(cube_cluster_start_node -> hex_config sync_ceph_config); clients can reach the
mon quorum regardless of which control node is down.
```

### DIFF
--- a/core/main/proj_functions
+++ b/core/main/proj_functions
@@ -287,6 +287,8 @@ cube_cluster_start_node()
     _cube_cluster_start_env
     Quiet -n ip neigh flush all
     Quiet -n $HEX_SDK ceph_leave_maintenance # ensure storage is working before other actions
+    # re-render mon host from the live monmap (boot commit writes only the master mon IP)
+    Quiet -n /usr/sbin/hex_config sync_ceph_config
     if [ "$VERBOSE" == "1" ] ; then
         /usr/sbin/hex_config -p trigger node_start
     else
--- a/core/modules/config_ceph.cpp
+++ b/core/modules/config_ceph.cpp
@@ -1744,6 +1744,13 @@ SyncConfigMain(int argc, char* argv[])
     std::string monIplist = HexUtilPOpen(HEX_SDK " ceph_mon_map_iplist %s", CONF);
     std::string mdsIplist = HexUtilPOpen(HEX_SDK " ceph_mds_map_iplist %s", CONF);
     std::string adminCliPass = GetSaltKey(s_saltkey, s_adminCliPass.newValue(), s_seed.newValue());
+
+    // no usable monmap (sdk returns "1" on failure): keep the existing conf
+    if (monHosts.length() == 0 || monIplist.find('.') == std::string::npos) {
+        HexLogWarning("sync_ceph_config: no monmap available, keeping %s", CONF);
+        return EXIT_SUCCESS;
+    }
+
     UpdateConfig(fsid, ctrl, ctrlIp, sharedId, monHosts, monIplist, pubNet, clstrNet, adminCliPass, s_perfTuned, mdsIplist);
 
     return EXIT_SUCCESS;
--- a/core/sdk_sh/modules/sdk_ceph.sh
+++ b/core/sdk_sh/modules/sdk_ceph.sh
@@ -296,14 +296,14 @@ ceph_mon_map_iplist()
     elif echo $iplist | grep -q 'v1' ; then
         iplist=$(monmaptool --print $MAPFILE | grep 6789 | awk -F' ' '{print $2}' | awk -F':' '{print $2}' | tr '\n' ',')
     fi
-    echo -n ${iplist:-1}
+    echo -n $iplist
 }
 
 ceph_mon_map_hosts()
 {
     ceph_mon_map_create $1
     local hosts=$(monmaptool --print $MAPFILE | grep 6789 | awk -F' ' '{print $3}' | cut -c 5- | tr '\n' ',')
-    echo -n ${hosts:-1}
+    echo -n $hosts
 }
 
 ceph_mds_map_iplist()


Eandalf-Bigstack:
Summary — good fix for the regression; restores per-node convergence and removes the harmful "1" sentinel. A few critical issues / suggestions:

1. shell: unquoted expansions
   - echo -n $iplist and echo -n $hosts must be quoted (echo -n "$iplist") to avoid word-splitting/globbing.

2. validation too narrow (IPv6 / hostnames)
   - monIplist.find('.') rejects valid IPv6 (":"), hostname-only monitors, and comma-separated lists. Replace with robust validation: trim whitespace, accept presence of '.' or ':' or ',' and/or validate against an IP/hostname regex.

3. trim and sanitize monIplist
   - strip newlines/trailing commas before validating to avoid false positives from e.g. log noise.

4. callers & backwards-compatibility
   - removing the sentinel may break other callers (CONTROL_REJOIN and any scripts). Audit callers and add defensive validation where they consume monmap output.

5. failure handling / timing
   - calling sync_ceph_config at boot synchronously could delay boot. Consider non-blocking run or a bounded timeout/retry, and log outcomes.

6. logging & testing
   - keep the HexLogWarning but add more context (node id). Add unit/integration tests for IPv4/IPv6/hostname/empty outputs and for the boot path.

Implementation notes (short):
- Change echo lines to echo -n "$iplist" / "$hosts".
- In config_ceph.cpp: trim monIplist, then check (contains '.' || contains ':' || contains ',') and/or run a regex IP/hostname validator; only then UpdateConfig.
- Audit and harden other consumers (e.g., CONTROL_REJOIN).

Overall: accept with fixes above — the behavioral change is correct but needs input sanitization, quoting, broader validation, and caller hardening.

Eandalf-Bigstack:
This is a solid catch. The root cause analysis in the PR description is spot-on—relying solely on the master IP after a rolling restart is a severe availability risk, and fixing the per-node convergence is the right architectural move.

However, looking at the proposed diff and the prior discussion, there are a few critical flaws that need to be addressed before this can be merged, particularly around IPv6 support and boot reliability.

Here is my review:

### 🚨 High Priority / Blockers

**1. IPv6 and Ceph v2 Protocol Breakage (config_ceph.cpp)**
The validation logic `monIplist.find('.') == std::string::npos` is overly strict and makes dangerous assumptions about network topologies:

* **IPv6:** If a cluster uses IPv6, the monitor IPs will look like `[fd00::10]:6789`. This contains no dots, causing the check to fail and silently skip the config sync, breaking IPv6 deployments.
* **Fix:** Replace the string-dot search with a more robust check. Validate that the string is not empty after trimming whitespace, and consider a regex that permits valid IPv4, IPv6, and hostname patterns, or simply rely on `monHosts.length() > 0 && monIplist.length() > 0` if `UpdateConfig` handles the sanitization.

**2. Synchronous Boot Blockers (proj_functions)**
You added `Quiet -n /usr/sbin/hex_config sync_ceph_config` to the node boot sequence.

* **Risk:** `sync_ceph_config` shells out to Ceph SDK commands. If the cluster is in a degraded state and those CLI commands hang, this synchronous call could completely stall the node's boot sequence.
* **Fix:** Wrap the `hex_config` call in a bounded timeout (e.g., `timeout 30s`) or run it asynchronously if the boot process doesn't strictly depend on its immediate completion to proceed.

**3. Shell Word Splitting / Globbing (sdk_ceph.sh)**
Removing the `:-1` sentinel is correct, but leaving the variables unquoted is risky.

* **Risk:** `echo -n $iplist` and `echo -n $hosts` will be subject to word splitting and globbing by the shell. If the `monmaptool` output ever includes unexpected characters or spaces, it will break the formatting.
* **Fix:** Double-quote the variables: `echo -n "$iplist"` and `echo -n "$hosts"`.

---

### ⚠️ Code Quality / Maintainability

**1. Outdated / Confusing Comments (config_ceph.cpp)**
The new comment in the C++ file states:
`// no usable monmap (sdk returns "1" on failure): keep the existing conf`

* **Issue:** This PR literally removes the `"1"` fallback from the SDK scripts in the same diff. The comment contradicts the new code.
* **Fix:** Update the comment to reflect the new reality (e.g., `// no usable monmap (sdk returns empty string on failure): keep the existing conf`).

**2. Trailing Commas in Shell Output (sdk_ceph.sh)**
The shell scripts use `tr '\n' ','` to format the IP and host lists.

* **Issue:** This inherently leaves a trailing comma at the end of the string (e.g., `IP1,IP2,IP3,`).
* **Fix:** Ensure that `UpdateConfig` in C++ strips trailing commas, or better yet, fix it in the shell script using `paste -sd,` instead of `tr` to cleanly join the lines without a trailing delimiter.

**3. Caller Auditing for Sentinel Removal**
As noted in the PR discussion, `CONTROL_REJOIN` and potentially other consumers previously expected `"1"` as the failure sentinel. You need to ensure that returning an empty string `""` from `ceph_mon_map_iplist` doesn't cause out-of-bounds segfaults or misconfigurations in those unpatched callers.

### Summary

**Status: Request Changes**
The behavioral intent is perfectly aligned with fixing the bug, but the implementation needs to be hardened. Please update the C++ validation to accommodate IPv6, quote your shell variables, protect the boot sequence from potential hangs, and tidy up the comments.

Eandalf-Bigstack (approved):

PR #1030, opened 2026-06-27
fix: stop error/warning log flooding on HA clusters

traviswu-bigstack:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it

Config/policy-layer corrections for the error/warning log floods on HA control-converged clusters:

- **masakari**: probe corosync via `corosync-cfgtool -s` (not a `tcpdump` sniff); gate both hostmonitor and processmonitor to run only under HA (robust HA read); SDK health check gated on HA.
- **haproxy**: add `timeout server` to the galera listen so idle DB connections aren't cut (stops `MySQL server has gone away (2013)` / aborted-connection floods).
- **monasca `ib_network`**: skip NICs without a counters dir (e.g. irdma) instead of raising.
- **logstash**: start after influxdb/kapacitor to stop the boot-time error flood.

#### Which issue(s) this PR fixes

Fixes #1029

#### Special notes for your reviewer

> Purely config/policy modules + the masakari `handle_host.py` and monasca `ib_network.py` patches. 6 commits, 6 files, CI green.

#### Additional documentation

```docs
kb/cubecos/known-issues/error-log-floods-corrections.md (drafted)
```

PR #129, opened 2025-05-29
fix: SDK mongodb_check syntax errors

leechienpang:
#### Which issue(s) this PR fixes

Fixes #

#### Special notes for your reviewer

> Note

#### Additional documentation

```docs

```

PR #1104, opened 2026-07-12
fix: cut steady-state log flooding on healthy clusters, phase 2

### LINKED ISSUE #1103: [Bug] Steady-state log flooding on healthy clusters (influx flux errors, DumpInterface link-bounce, monasca sudo churn, license probes, beats noise)
## Description

On a healthy, idle 3-node HA control-converged cluster, `/var/log/messages` grows at **~250 lines/min (~15k lines/hour)** of pure noise, burying real signals exactly when operators need them (a degraded peer makes it *worse* — kafka/etcd retry spam adds another ~100/min). Root-caused on the sky lab cluster (2026-07-11/12); this is the steady-state phase-2 companion to the FTS-phase floods of #1029.

Seven distinct sources, largest first:

1. **influxd flux "Invalid column reader received from predecessor"** (~42% of syslog, ~70/min) — the UI/API events polling (httpd → VIP:8086) queries the `events` measurement where some rows lack the `detail` field; the flux engine logs ~30 `lvl=info` lines per query.
2. **kernel `8021q: adding VLAN 0 to HW filter` bursts** (~14%) — `hex_sdk DumpInterface` **up-probes every admin-down NIC** (then downs it) to read link speed; cube-cos-api node-sync calls it every 30s from every node, bouncing 6 links each time. Also made each API node sync take up to ~48s (8s speed-wait × down NICs).
3. **systemd/logind session churn** (~13%) — 75% from monasca `ipmi_sensors` running `sudo ipmitool sensor reading <sensor>` once **per sensor** per 30s cycle (5.8k sudo sessions/12h); the rest from peer health-check ssh whose `ControlPersist 30s` expires between ~40s poll cycles.
4. **hex_config license probes** (~5%) — every `license_check` invocation logs 3 lines at Notice/Info ("Checking license…" / "License files are not installed" / "Command … exited with status"), driven by the ~40s health cycle. A *licensed* cluster spams equivalently ("still valid for N days").
5. **filebeat filestream scanner warnings** (~7%) — `horizon_access.log` fingerprint-collides with `access.log` (byte-identical duplicate: two CustomLog targets), plus "N files too small to be ingested" for one-shot tool logs (privsep/manage/dbsync) that never reach the 1KB fingerprint minimum.
6. **beats "Non-zero metrics" snapshots** — filebeat+auditbeat each dump a full-page JSON metrics blob to syslog every 30s at info.
7. Down-node amplification: with one control node off, filebeat/auditbeat/neutron rdkafka retry spam + etcd raft prober warnings push the rate to ~250+/min.

## How to Reproduce

1. Install a 3-node HA control-converged cluster; let it reach steady state (`cluster check` all ok).
2. `tail -f /var/log/messages` on any control node, or count: `tail -2000 /var/log/messages | awk '{print $1,$2,substr($3,1,5)}' | uniq -c`.
3. Observe ~150–250 lines/min matching the patterns above with zero operator activity.

## Expected behavior

A healthy idle node's syslog should be quiet (tens of lines/min at most), so that real warnings/errors are visible. Monitoring/polling machinery must not bounce NIC links, spawn a sudo session per sensor, or log per-probe status at Notice.

All fixes are implemented and **verified live on the sky cluster** (hot-deploy, runtime-equivalent): quiet-node rate dropped **~250 → ~22–26 lines/min (~90%)** with each pattern independently verified at zero. Fix inventory (PRs to follow, one per repo):

- cubecos: influxdb template `[logging] level=warn`; filebeat template: native (inode) file identity — kills the 1KB fingerprint floor and the silent sub-1KB-never-ships gap — plus exclude list + `logging.metrics.enabled: false` (both beats); `ControlPersist 30s→120s`; monasca `ipmi_sensors` batched into one multi-sensor ipmitool call (note: ipmitool exits 1 when any named sensor is missing while still printing the rest).
- hex: `DumpInterface` no longer up-probes down NICs on the `-f json` path (reports capability speed from advertised modes — readable while down); license probe logging demoted to Debug for expected states (alarm states stay at Notice); generic per-command "exited with status" dispatcher line demoted to Debug.
- Deferred: cube-cos-api NIC-spec sync 30s→hourly (static data); influx `events` schema normalization (`detail` field) as the root fix for (1).

Handbook knowledge update (DoD): `kb/cubecos/known-issues/steady-state-log-floods-corrections.md` in bigstack-handbook — full symptom→root-cause→fix→verification chain for every item, including the measurement pitfalls (dmesg ring-buffer saturation, rsyslog out-of-order blocks).

Related: #1029 (FTS-phase floods, phase 1).

- [x] **Handbook knowledge update** (landed: `kb/cubecos/known-issues/steady-state-log-floods-corrections.md`, bigstack-handbook#57)

traviswu-bigstack:
#### What type of PR is this?

/kind bug

Phase 2 of the log-noise work (phase 1 = #1030, FTS-time floods): a healthy **idle** 3-node HA control node logs ~250 lines/min of noise. This PR (plus its hex companion) takes that to ~25 lines/min — ~90% — with every source root-caused and independently verified on the sky lab cluster:

| Fix | Flood it kills | Share |
|---|---|---|
| influxdb template: `[logging] level=warn` | flux "Invalid column reader" per UI/API events query | ~42% |
| hex bump (bigstack-oss/hex#129): DumpInterface json path stops up-probing down NICs | kernel `8021q: adding VLAN 0` bursts; also un-stalls API node sync (≤8s per down NIC) | ~14% |
| monasca `ipmi_sensors`: one batched ipmitool call/cycle instead of one sudo per sensor | 75% of root session churn (5.8k sudo/12h) | ~10% |
| ssh `ControlPersist 30s→120s` | peer health-ssh re-login every ~40s poll | ~3% |
| hex bump: license probe logging → Debug (alarm states stay Notice) | license triplets every health cycle | ~5% |
| filebeat template: **native (inode) file identity** (kills the 1KB fingerprint floor — quiet copytruncate-rotated service logs warned "too small" every 10s AND sub-1KB error bursts never shipped); exclude `horizon_access.log` only (byte-identical CustomLog duplicate — with native identity both copies would ship; the small one-shot tool logs no longer need excluding since native identity has no size floor); `logging.metrics.enabled: false` for both beats | filestream scanner warnings + per-30s JSON metrics blobs + a silent observability gap | ~9% |

Fixes #1103

> - `ipmi_sensors` batching: ipmitool exits 1 when any named sensor is missing **while still printing the rest**, so the check parses stdout regardless of return code (verified against missing-sensor hardware).
> - influxd restart briefly errors cube-cos-api health queries ("failed to get query cursor") — roll one node at a time when this lands.
> - Native-identity migration verified live on sky142: no history re-ingest (offset delta ~= baseline; modest startup catch-up), and a canary appended to a 0-byte octavia log reached kafka in 2s. Safe here because cube logs rotate via copytruncate (inode-stable).
> - Remaining known noise is deliberate and documented: octavia/masakari 0-byte service logs stay watchable (copytruncate + journald-bound request logs = empty on healthy days), cube-cos-api NIC-spec sync 30s→hourly deferred to that repo.

#### Additional documentation

```docs
Handbook: kb/cubecos/known-issues/steady-state-log-floods-corrections.md (bigstack-handbook) — full
symptom → root cause → fix → verification chain per item, incl. measurement pitfalls
(dmesg ring-buffer saturation, rsyslog out-of-order blocks).
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,7 @@
 TAGS
 jail/ubi9.tar.gz
 triangle
+
+# Python bytecode
+__pycache__/
+*.pyc
--- a/core/modules/config_cubesys.cpp
+++ b/core/modules/config_cubesys.cpp
@@ -117,8 +117,11 @@ WriteSshConfig(void)
     // Multiplex repeated connections so health checks (one ssh per service per
     // node) reuse a single session instead of flooding sshd/logind/sudo.
     fprintf(fout, "ControlMaster auto\n");
+    // control socket is root-only in /run (tmpfs, cleared on reboot) on a
+    // single-tenant appliance; 120s outlives the ~40s health poll so masters
+    // don't expire between cycles.
     fprintf(fout, "ControlPath /run/cube-ssh-%%C\n");
-    fprintf(fout, "ControlPersist 30s\n");
+    fprintf(fout, "ControlPersist 120s\n");
     fclose(fout);
 
     return true;
--- a/core/modules/config_logstash.cpp
+++ b/core/modules/config_logstash.cpp
@@ -83,6 +83,13 @@ WriteConfigs(const std::string &kafkaHosts)
 
     fprintf(fout, "filebeat.inputs:\n");
     fprintf(fout, "- type: filestream\n");
+    // native (inode) identity: no 1KB fingerprint floor, so sub-1KB writes
+    // ship; safe here -- these logs rotate via copytruncate (inode-stable)
+    fprintf(fout, "  file_identity.native: ~\n");
+    fprintf(fout, "  prospector.scanner.fingerprint.enabled: false\n");
+    // horizon_access is a byte-identical duplicate of access.log (both
+    // CustomLog targets); with native identity both copies would ship
+    fprintf(fout, "  prospector.scanner.exclude_files: ['horizon_access\\.log$']\n");
     fprintf(fout, "  paths:\n");
     fprintf(fout, "    - /var/log/messages\n");
     fprintf(fout, "    - /var/log/nova/*.log\n");
@@ -123,6 +130,10 @@ WriteConfigs(const std::string &kafkaHosts)
     fprintf(fout, "  max_message_bytes: 1000000\n");
     fprintf(fout, "\n");
 
+    // keep the per-30s metrics snapshots out of syslog
+    fprintf(fout, "logging.metrics.enabled: false\n");
+    fprintf(fout, "\n");
+
     fclose(fout);
 
     fout = fopen(AB_CONF, "w");
@@ -158,6 +169,10 @@ WriteConfigs(const std::string &kafkaHosts)
     fprintf(fout, "    reachable_only: false\n");
     fprintf(fout, "  required_acks: 1\n");
     fprintf(fout, "  max_message_bytes: 1000000\n");
+    fprintf(fout, "\n");
+
+    // keep the per-30s metrics snapshots out of syslog
+    fprintf(fout, "logging.metrics.enabled: false\n");
     fprintf(fout, "\n");
 
     fclose(fout);
--- a/core/monasca/yoga_patch/monasca_agent/collector/checks_d/ipmi_sensors.py
+++ b/core/monasca/yoga_patch/monasca_agent/collector/checks_d/ipmi_sensors.py
@@ -1,8 +1,13 @@
 # Monasca agent check: host hardware sensors via IPMI.
 # Feeds Watcher thermal_optimization / airflow_optimization / saving_energy.
+import re
 import subprocess
 
 import monasca_agent.collector.checks as checks
+
+# first numeric token in a `sensor reading` value cell ("45", "36.500",
+# or a unit-bearing "12.3 degrees C" from other firmware)
+_NUM = re.compile(r'[-+]?[0-9]*\.?[0-9]+')
 
 
 class IpmiSensors(checks.AgentCheck):
@@ -24,13 +29,27 @@ class IpmiSensors(checks.AgentCheck):
     def check(self, instance):
         dimensions = self._set_dimensions(None, instance)
         sensors = instance.get('sensors') or self.DEFAULT_SENSORS
-        for metric, sensor in sensors.items():
-            try:
-                out = subprocess.check_output(
-                    ['sudo', '-n', 'ipmitool', 'sensor', 'reading', sensor],
-                    timeout=15, stderr=subprocess.DEVNULL).decode()
-                value = float(out.split('|')[1].strip())
-            except (OSError, subprocess.SubprocessError, IndexError, ValueError) as e:
-                self.log.debug('ipmi sensor %s unavailable: %s', sensor, e)
+        # one batched ipmitool call (one sudo) per cycle; a missing sensor is
+        # omitted from output and forces exit 1 while the rest still print, so
+        # parse stdout regardless of return code (do NOT gate on it)
+        try:
+            proc = subprocess.run(
+                ['sudo', '-n', 'ipmitool', 'sensor', 'reading'] + list(sensors.values()),
+                timeout=15, stdout=subprocess.PIPE, stderr=subprocess.DEVNULL)
+            out = proc.stdout.decode()
+        except (OSError, subprocess.SubprocessError) as e:
+            self.log.debug('ipmi sensors unavailable: %s', e)
+            return
+        readings = {}
+        for line in out.splitlines():
+            name, sep, value = line.partition('|')
+            if not sep:
                 continue
-            self.gauge(metric, value, dimensions=dimensions)
+            m = _NUM.search(value)
+            if m:
+                readings[name.strip()] = float(m.group())
+        for metric, sensor in sensors.items():
+            if sensor in readings:
+                self.gauge(metric, readings[sensor], dimensions=dimensions)
+            else:
+                self.log.debug('ipmi sensor %s unavailable', sensor)


Eandalf-Bigstack:
Summary
- Good set of targeted noise fixes; measurable impact. A few critical issues need attention before merge.

Findings & action items
- Critical: remove committed binary __pycache__/ipmi_sensors.cpython-36.pyc — don’t check .pyc into git; add to .gitignore and rebase.
- Filebeat/YAML risk: keys like file_identity.native and prospector.scanner.* look version-specific/deprecated. Validate generated YAML against the deployed Filebeat version (yaml lint + start/validate) — mis‑keys can crash logging and create bigger outages than the noise.
- ipmi_sensors parsing: float(value.strip()) will fail when ipmitool prints units (e.g., "12.3 degrees C"). Parse numeric token via regex ([-+]?[0-9]*\.?[0-9]+) and normalize sensor names (strip/case) before matching. Consider checking proc.returncode and logging stderr for diagnostics.
- ControlPersist 120s: acceptable for small clusters but extends socket lifetime — document security implications, ensure /run path perms and monitor for stale sockets/resource growth under scale.
- Submodule bump (hex): ensure companion hex PR is merged and deployed together; verify kernel/VLAN issues resolved in that version.
- Tests: add/execute YAML validation and a rolling-deploy smoke test (filebeat, influxd, monasca) to avoid health-query regressions.

Approve after: remove pyc, YAML validation, robust ipmi parsing, and confirming hex companion deploy.

Eandalf-Bigstack:
This is a solid initiative to tackle log noise, and the measurable reductions you’ve outlined in the PR description are highly impactful for the cluster's observability health. The batching approach for `ipmitool` is a particularly great optimization.

However, I have reviewed the implementation against the prior discussion and codebase context, and there are a few critical syntax and logic issues that will break logging and metric collection if merged as-is.

Here is my review of the necessary corrections before approval.

## Critical Issues & Bugs

### 1. Filebeat YAML Syntax Mismatch (Logging Outage Risk)

The C++ config generator is injecting obsolete `prospector` keys into a modern `filestream` input block.

* **The Flaw:** In `core/modules/config_logstash.cpp`, you define `- type: filestream`, but immediately follow it with `prospector.scanner.fingerprint.enabled: false` and `prospector.scanner.exclude_files: ...`. The `prospector` namespace was deprecated years ago and is completely invalid under the `filestream` input schema. This configuration will cause Filebeat to fail startup validation, resulting in a total logging outage for the node.
* **The Fix:** For `filestream` inputs, `file_identity.native: ~` implicitly replaces the fingerprinting requirement. You can safely remove the `prospector.scanner.fingerprint.enabled` line entirely. The exclusion rule should be moved directly under the input configuration as `exclude_files: ['horizon_access\.log$']`.

### 2. IPMI Sensor Value Parsing

The strict float casting will fail in many hardware environments.

* **The Flaw:** In `ipmi_sensors.py`, `float(value.strip())` assumes the raw value from `ipmitool` is purely numeric. If the hardware returns units (e.g., `12.5 Volts` or `45 degrees C`), the `float()` cast will throw a `ValueError`, causing the loop to swallow the error and skip reporting healthy sensors.
* **The Fix:** Import the `re` module and use regex to extract the numeric portion before casting. For example: `match = re.search(r'[-+]?\d*\.?\d+', value); parsed_val = float(match.group()) if match else None`.
* **Sensor Name Matching:** Ensure you normalize both the config `sensor` name and the output `name` string (e.g., `.strip().lower()`) before comparing them in the dictionary to prevent subtle case-sensitivity mismatches across different server vendors.

### 3. Committed Binary Artifacts

* **The Flaw:** The PR includes `core/monasca/yoga_patch/monasca_agent/collector/checks_d/__pycache__/ipmi_sensors.cpython-36.pyc`.
* **The Fix:** Compiled python byte-code should never be tracked in version control. Remove it from the branch (`git rm --cached ...`) and ensure `__pycache__/` is listed in your `.gitignore` file.

## Architectural & Security Considerations

### Subprocess Error Visibility

In the `ipmi_sensors.py` exception block, you are redirecting `stderr` to `subprocess.DEVNULL`. Because `ipmitool` is known to exit 1 if *any* sensor is missing, completely suppressing standard error makes diagnosing legitimate IPMI subsystem failures (e.g., daemon crashes, permission drops) incredibly difficult. Capture `stderr` using `subprocess.PIPE` and log it at a `DEBUG` or `WARNING` level when a subprocess error is caught.

### SSH Socket Persistence Risk

Bumping `ControlPersist` to `120s` in `config_cubesys.cpp` solves the health-poll expiration problem but extends the lifecycle of multiplexed SSH sockets in `/run/`. This increases the temporal attack surface for socket hijacking if a bad actor gains access to the node. Ensure the generated sockets (`/run/cube-ssh-%C`) are restricted via directory permissions (e.g., `0700` owned by the executing user) to mitigate lateral movement risks.

## Required Action Items for Approval

1. Remove the tracked `.pyc` file and update `.gitignore`.
2. Correct the Filebeat YAML generation to strictly use `filestream` compatible schemas and validate the output against the target Filebeat binary (`filebeat test config`).
3. Implement regex-based float extraction and string normalization in the IPMI python script.
4. Capture and log `stderr` from the `ipmitool` subprocess instead of nullifying it.
5. Confirm the companion `hex` submodule PR (#129) is merged and green before we merge this phase.

core/monasca/yoga_patch/monasca_agent/collector/checks_d/__pycache__/ipmi_sensors.cpython-36.pyc
Eandalf-Bigstack, line 1: We might need to check-in the source code instead. Otherwise, we would have no clue to maintain this part.

traviswu-bigstack:
Thanks for the review — addressed in the amended commit (force-pushed). Point by point:

**✅ Committed `.pyc` — fixed.** Removed `__pycache__/ipmi_sensors.cpython-36.pyc` from the tree and added `__pycache__/` + `*.pyc` to `.gitignore` (it slipped in from a local `py_compile` during editing). Good catch.

**❌ Filebeat "prospector keys invalid → logging outage" — this is a false alarm; the keys are correct.** The `prospector.scanner.*` namespace is the **current filestream schema** (`prospector.scanner.exclude_files`, `prospector.scanner.fingerprint.*`, `prospector.scanner.symlinks`, …) — not the deprecated top-level `filebeat.prospectors`, which is what that namespace was years ago. Verified on the deployed version:
```
# filebeat version
filebeat version 9.2.3
# filebeat test config -c /etc/filebeat/filebeat.yml
Config OK
```
No deprecation/unknown-field warnings in the filebeat log, and it has been running `active` and shipping on all three sky nodes since the hot-deploy (a canary in a 0-byte log reached kafka in 2s). Note the suggested fix — a bare `exclude_files:` under the `filestream` input — is **not** valid there; it must stay under `prospector.scanner.`.

**⚠️ ipmi parsing — hardened anyway.** `ipmitool sensor reading <s>` emits a **bare** number (`Outlet-TMP | 45`, `PSU1_Inlet_TEMP | 36.500`) — no units — so `float()` was fine for this call (units appear with `ipmitool sensor`, the full listing, which we don't use). Still, I switched value extraction to a numeric-token regex `[-+]?[0-9]*\.?[0-9]+` so a unit-bearing reading from other firmware degrades gracefully. Unit-tested against bare / `12.3 degrees C` / non-numeric / no-pipe inputs. Re: `returncode` — deliberately **not** gated: a missing sensor makes ipmitool exit 1 while still printing the rest, so we parse stdout regardless (documented in the comment). Re-deployed live; `host_power` still flowing for all 3 nodes.

**ℹ️ ControlPersist 120s — noted in code.** Added a comment: the control socket is root-only in `/run` (tmpfs, cleared on reboot) on a single-tenant appliance, so the socket-lifetime concern is bounded.

**ℹ️ hex submodule** — companion is bigstack-oss/hex#129 (ready, green); this bump points at its head and should merge/deploy together.

Eandalf-Bigstack (approved):

Issue #651, opened 2026-01-27
[Story] Upgrade MariaDB from 10.6.x to 10.11.x


jdjgya:
## Description

As a CubeCOS operator, I want to have MariaDB compatible with OpenStack Caracal so that we could upgrade to OpenStack Caracal.

## Acceptance Criteria

> A detailed list of conditions that must be met for the story to be accepted. These statements must focus on functional behavior and constraints rather than specific code-level implementation.

- [x] Have a OpenStack Caracal compatible MariaDB installed, which should be 10.11.x.

### Rolling-upgrade / mixed-version DoD (added 2026-07-12)

Must be safe to apply **one node at a time on a live 3-node HA cluster** (control tier), leaving the cluster serving throughout — CubeCOS drains VMs by **live migration**, which itself needs a healthy Galera DB. Verify on a real HA cluster:

- [ ] **Quorum preserved** — rolling one node at a time keeps Galera in a Primary component (never below 2/3); the roll halts safely rather than dropping to 1/3 if a node cannot rejoin.
- [x] **Mixed-version cluster operates** — with nodes on both the old and new MariaDB versions at once, every member reaches `wsrep_local_state_comment = Synced`, `wsrep_cluster_size` returns to 3, and OpenStack read+write DB ops succeed against every node. Hold this mixed state ≥10 min (a realistic firmware-reboot window).
- [x] **Cross-version rejoin** — a new-version node rejoins from an old-version donor via SST/IST (`wsrep_sst_method = rsync`) with no error.
- [x] **`mariadb-upgrade` deferred to finalization** — system-table upgrade runs **once, only after all nodes are new**, never per-node mid-roll (no system-table DDL replicated into a mixed cluster).
- [ ] **Health-gated advance** — the node roll advances only when the just-upgraded node reports `Synced` + cluster_size restored (not on config-done markers alone).
- [x] **Live-migration smoke test passes** while the cluster is mixed-version (nova live migration completes end-to-end — it writes migration/instance-host rows to the DB).
- [x] Source-release OpenStack is validated to run on the new MariaDB, so the bridge can be delivered (e.g. as a no-reboot fixpack) **before** the OpenStack firmware upgrade.

### Verification status (2026-08-19)

[...]

Issue #614, opened 2026-01-27
[Story] Upgrade RabbitMQ from 3.9.21 to 3.10.x


jdjgya:
## Description

As a CubeCOS operator, I want RabbitMQ to be compatible with OpenStack Antelope so that we could upgrade to Antelope.

~We would also need to upgrade Erlang from 24 to 25 for RabbitMQ 3.12 to run.~

We would need to upgrade to RabbitMQ 3.10 to run and provide the rolling upgrade capability.

## Acceptance Criteria

> A detailed list of conditions that must be met for the story to be accepted. These statements must focus on functional behavior and constraints rather than specific code-level implementation.

- [x] ~Upgrade Erlang from 24 to 25~
- [x] Upgrade RabbitMQ from 3.9.21 to ~3.12.x~ 3.10.x
  - [x] Ensure the queue is set up and exchanges for OpenStack are shown up.

### Rolling-upgrade / mixed-version DoD (added 2026-07-12)

Must be safe to apply **one node at a time on a live 3-node HA cluster** (control tier) — CubeCOS drains VMs by **live migration**, which needs the message bus for source↔dest RPC. Verify on a real HA cluster:

- [x] **Mixed-version cluster operates** — with nodes on both the old and new RabbitMQ/Erlang versions at once, `rabbitmqctl cluster_status` shows all nodes running with no partition, queues stay `synchronised`, and OpenStack RPC (nova/neutron/cinder) succeeds. Hold ≥10 min (a realistic firmware-reboot window).
- [x] **Feature flags** — no new feature flags are enabled until **all** nodes are new; any flag the target *requires* is enabled at the correct point per RabbitMQ's rules. Document the feature-flag sequence.
[...]

PR #1105, opened 2026-07-12
feat(upgrade): mixed-version rolling-upgrade safety (gcache, live-migration gate, version-uniformity gates)

traviswu-bigstack:
#### What type of PR is this?

/kind feature

Bundles the **mixed-version rolling-upgrade safety** changes for the Yoga→Epoxy SLURP path (established by the DB/MQ mixed-version spike). Five self-contained fixes, one per commit-worth of concern:

1. **gcache 300M → 4G** (`config_mysql.cpp`) — a rejoining node after a firmware-reboot-length downtime must catch up via **IST**; SST is broken for the MariaDB 10.6→10.11 jump (proven: the joiner crashes). Sizing the gcache guarantees IST. On-disk ring buffer, negligible cost. Ships from the Antelope image so the 10.6 **donors** carry it into the A→C roll.

2. **Live-migration gate on the rolling-upgrade relay** (`proj_functions` + `bootstrap_cube_config`) — the relay advanced to draining the next node on config-done **markers only**. `live_migration_gate` asserts the drain path is *functional* (Galera Synced+Primary, RabbitMQ clustered, ceph PGs servable, nova conductor/scheduler up + ≥1 live target, cinder-volume up, neutron alive) **without asserting version uniformity**, so a mixed N/N+1 SLURP window passes. keystone/haproxy/vip/memcached are covered transitively by the authenticated `openstack` calls; glance/cyborg are not on the live path. Pauses (never advances into a broken cluster) on timeout. Runs **after** `hex_config -p bootstrap standalone-done`.

3. **Masakari maintenance on planned reboot** (`sdk_os.sh`) — `os_segment_host_maintenance` marks a host in masakari maintenance **before** the planned drain (hooked into `os_pre_failure_host_evacuation[_sequential]`), so hostmonitor doesn't cold-evacuate (rebuild) a deliberately-rebooting node — a hard kill that also forces a Galera SST.

4. **`mysql_upgrade` gated on version uniformity** (`config_mysql.cpp` + `os_mariadb_version_uniform`) — defer the system-table upgrade until every control node reports the same MariaDB version, so its DDL never replicates into not-yet-upgraded peers. Version uniformity, not firmware, is the correct gate for web-scale.

5. **RabbitMQ feature-flag enable gated on version uniformity** (`config_rabbitmq.cpp` + `os_rabbitmq_version_uniform`) — enabling a flag is irreversible and strands nodes that don't support it; only enable when the cluster is version-uniform (pre-roll all-old = prep; post-roll all-new = graduate; never mid-roll).

Unifying principle for 4 & 5: cluster-wide graduations (`mysql_upgrade`, feature-flag enable) are gated on **per-service version uniformity**, never run mid-roll.

Relates to #651 (10.6→10.11 IST), #614/#785/#796 (RabbitMQ feature-flags/Erlang), and the traced rolling-upgrade gaps (marker-based advance, masakari-not-suppressed, mysql_upgrade mid-roll).

> **Needs lab validation** — items 2/3 sit in the boot/relay path. Verify: a healthy same-version rolling restart still advances; a mixed-version rolling upgrade passes the gate and does NOT cold-evacuate; the version-uniformity gates defer during the mixed window and fire once converged. All C++ compiles (`config_mysql.o`, `config_rabbitmq.o`); shell syntax-checked. `cinder-volume` check assumes the pacemaker active/passive resource. Follow-up: unify the restart path onto `live_migration_gate` and drop the `if upgrading` fork.

#### Additional documentation

```docs
Handbook: kb/cubecos/architecture/db-mq-mixed-version-upgrade-spike.md and
slurp-upgrade-yoga-to-epoxy.md (spike, empirical IST/SST, the from-Antelope rule).
```

---
**Update:** folded in `feat(ha): always create the vaw resource in SetupCluster` — drops the `is_node_rolling_upgrade` guard so the vaw pacemaker resource (with its `vaw with vip` colocation) is created at every cluster setup, keeping FTS and upgrade paths consistent.

### DIFF
--- a/core/main/bootstrap_cube_config
+++ b/core/main/bootstrap_cube_config
@@ -68,6 +68,17 @@ RollingEvacuateAndReboot()
                 if [ "x$n_lastboot" = "xlast_boot = Never" -a "x$n_fwver" = "x$fwver" ] ; then
                     upgrading=true
                     echo "Starting auto rolling upgrade" >> $BOOTSTRAP_CUBE_MODE
+                    # Don't drain+reboot the next node until this node has rejoined
+                    # and the live-migration path is functional. Version-robust, so
+                    # a mixed N/N+1 SLURP window passes; pauses (never advances into
+                    # a broken cluster) if it can't confirm within the timeout.
+                    if ! hex_sdk live_migration_gate 300 "$N" ; then
+                        msg="Auto rolling upgrade paused: live-migration path not ready before draining $N"
+                        hex_sdk remote_run $n_ip "echo -n \"$msg\" >$BOOTSTRAP_CUBE_LOG"
+                        echo -n "$space_filler">$(hex_sdk GetKernelConsoleDevices) ; echo -n "$msg" $'\r' | tee $BOOTSTRAP_CUBE_LOG $(hex_sdk GetKernelConsoleDevices)
+                        hex_sdk influx_event "${query},severity=\"W\" message=\"$msg\""
+                        break
+                    fi
                     if [ "x$n_role" = "xcontrol-converged" -o "x$n_role" = "xedge-core" -o "x$n_role" = "xcompute" ] ; then
                         msg="Evacuating VMs on $N"
                         echo -n "$space_filler">$(hex_sdk GetKernelConsoleDevices) ; echo -n "$msg" $'\r' | tee $BOOTSTRAP_CUBE_LOG $(hex_sdk GetKernelConsoleDevices)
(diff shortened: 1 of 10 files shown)


traviswu-bigstack:
**Update:** `live_migration_gate` now also gates the rolling-**restart** relay (`_power_roll_kick`, right before `power_node_evacuate`), not just the upgrade relay — so **one gate protects both cases**, per the discussion that they share the same live-migration-must-work requirement. The restart relay previously gated the drain on `cube_cluster_ready` (config-done markers) only. `cluster_check_repair_essential` is retained as the repair primitive (also used by the `cluster check_repair` CLI) — a distinct role from the pre-drain gate.

PR #1109, opened 2026-07-13
feat(upgrade): gate the rolling-upgrade relay on a functional live-migration path

traviswu-bigstack:
#### What type of PR is this?

/kind feature

Adds a **version-robust live-migration health gate** to the rolling-**upgrade** relay, which today advances to draining the next node on **config-done markers only** — no cluster-health check. That can take the next node down while live migration (the drain mechanism) isn't actually functional.

Rolling **restart** already gates via `cluster_check_repair_essential`. Rolling **upgrade** had no gate, because the full essential check reports NG on mixed-version skew and was deliberately skipped. But both modes have the *same* requirement — live migration must work to drain each node — so the gate belongs to both.

`live_migration_gate` (`proj_functions`) asserts the live-migration path is **functional**, never version-uniform, so a mixed N/N+1 SLURP window passes:
- **version-safe membership**: Galera `Synced`+`Primary`, RabbitMQ clustered, ceph PGs servable (`min_size`)
- **functional backends**: nova conductor/scheduler up + ≥1 live target (excluding the draining node), `cinder-volume` up, neutron alive
- **transitively covered** by the authenticated `openstack` calls: keystone, haproxy, vip, memcached
- **not gated** (not on the live-migration path): glance (live migration ships running state, no re-image), cyborg (accelerator VMs can't live-migrate)

Wired into `RollingEvacuateAndReboot` right **before** the next node's drain (`os_pre_failure_host_evacuation_sequential`). If the path can't be confirmed within the timeout, the roll **pauses** rather than draining into a broken cluster. Invoked **after** `hex_config -p bootstrap standalone-done` (which is what rejoins this node), so it can actually assert the rejoin.

Relates to the traced rolling-upgrade gap: advance is marker-based, not health-based.

> **Needs lab validation** — this sits in the boot/relay path. Verify: (a) a healthy same-version rolling restart still advances promptly; (b) a mixed-version rolling upgrade passes the gate mid-roll; (c) an intentionally-degraded cluster pauses the roll instead of draining. The `cinder-volume` check assumes the pacemaker active/passive cinder-volume resource ("up somewhere"). Follow-up: replace the restart path's `cluster_check_repair_essential` gate with this same function so both modes share one gate, and drop the `if upgrading` fork.

#### Additional documentation

```docs
Handbook kb/cubecos/architecture/db-mq-mixed-version-upgrade-spike.md and
slurp-upgrade-yoga-to-epoxy.md cover why mixed-version-safe DB/MQ makes the
version-safe membership signals reliable mid-upgrade (the basis for this gate).
```

### DIFF
--- a/core/main/bootstrap_cube_config
+++ b/core/main/bootstrap_cube_config
@@ -68,6 +68,17 @@ RollingEvacuateAndReboot()
                 if [ "x$n_lastboot" = "xlast_boot = Never" -a "x$n_fwver" = "x$fwver" ] ; then
                     upgrading=true
                     echo "Starting auto rolling upgrade" >> $BOOTSTRAP_CUBE_MODE
+                    # Don't drain+reboot the next node until this node has rejoined
+                    # and the live-migration path is functional. Version-robust, so
+                    # a mixed N/N+1 SLURP window passes; pauses (never advances into
+                    # a broken cluster) if it can't confirm within the timeout.
+                    if ! hex_sdk live_migration_gate 300 "$N" ; then
+                        msg="Auto rolling upgrade paused: live-migration path not ready before draining $N"
+                        hex_sdk remote_run $n_ip "echo -n \"$msg\" >$BOOTSTRAP_CUBE_LOG"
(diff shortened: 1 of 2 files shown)


traviswu-bigstack:
Folded into #1105 (combined into one mixed-version rolling-upgrade-safety PR). Closing.

PR #1113, opened 2026-07-14
fix(apache): remove default welcome page on port 8080

raven-pan:
#### What type of PR is this?

/kind bug

Fixes the Apache default "Test Page" served on port 8080 of the management IP (Tenable plugin 112351). Default service pages fail CIS Benchmark, PCI-DSS 2.2.1, and NIST 800-53 CM-6/CM-7, which surfaces in enterprise procurement scans.

Two layers of defense:

- **Bind Apache's 8080 listener to loopback** (`config_apache2.cpp`). Port 8080 is httpd's main listener (Horizon, Keystone `/identity`, `/server-status`) so it can't be removed — but its only two consumers reach it over loopback (HAProxy's `openstack_horizon` backend, the Monasca agent's check). Binding `127.0.0.1` removes it from off-box scans.
- **Stop serving default content** even on-box: drop the RPM's `welcome.conf` + `/usr/share/httpd/noindex` from the image (`keystone.mk`), and deny the empty DocumentRoot with `Require all denied` / `Options -Indexes` (`config_apache2.cpp`). The deny block is required — without it, removing `welcome.conf` would expose a browsable directory listing, since stock httpd.conf grants `Options Indexes` on `/var/www/html` and `welcome.conf` carried the only `-Indexes` covering `/`.
- **Pin the Monasca check to `127.0.0.1`** (`config_monasca.cpp`), since `localhost` can resolve to `::1` where httpd no longer listens.

**Why literal `127.0.0.1` and not `localhost`**

Both the listener bind (`config_apache2.cpp`) and every consumer (the Monasca check, HAProxy's `openstack_horizon` backend) use the literal IPv4 address, never the hostname `localhost`. `localhost` resolves via `/etc/hosts` to **both** `127.0.0.1` and `::1`, and the resolver commonly returns `::1` (IPv6) first. That matters on both sides:

- **Bind side** — `Listen localhost:8080` would bind **both** `127.0.0.1:8080` *and* `[::1]:8080`, widening the surface this PR is trying to shrink, and making socket creation depend on name resolution at boot (a fragile dependency for an early-boot service). The literal binds exactly one IPv4 loopback socket, with no resolver step.
- **Connect side** — a client using `localhost` may hit `[::1]:8080`, where httpd (bound to IPv4 loopback only) is *not* listening → `connection refused`. This is exactly the failure the `fix(monasca)` commit prevents by pinning the check URL to `127.0.0.1`.

Using the literal keeps the whole path single address-family (IPv4 loopback), resolver-independent, and symmetric — matching what HAProxy already does.

Commits:
- `fix(monasca): pin the apache server-status check to 127.0.0.1`
- `fix(apache): bind 8080 to loopback and deny the empty docroot`
- `fix(apache): stop shipping the default welcome page`

Fixes bigstack-oss/cubecos-private#76

> ⚠️ **Not yet verified on a node.** These commits were written and reviewed without a build environment (Docker/jail unavailable). The C++ passed `g++ -fsyntax-only`; a whole-branch review confirmed the sed expansion, generated Apache config, Makefile recipe, and — the load-bearing assumption — that no service depends on cross-node `:8080`, so the loopback bind does not break HA.
>
> **Do not merge until the checklist below passes in the jail against a test node.**

**Pre-merge verification (run in the jail)**

Build:
- [ ] `make rootfs` succeeds
- [ ] Staging tree has `welcome.conf.orig` and **no** `welcome.conf` in `etc/httpd/conf.d/`
- [ ] Staging tree has **no** `usr/share/httpd/noindex` directory

On a booted node:
- [ ] `ss -lntp | grep 8080` shows `127.0.0.1:8080` only — no `0.0.0.0` / `*` / `[::1]`
- [ ] `curl -I http://<mgmt-ip>:8080/` from another host → **connection refused** (the acceptance criterion)
- [ ] `curl -sI http://127.0.0.1:8080/` → `403`, `Server: Apache` with no version, no test page, no `Index of /`
- [ ] Horizon still loads through the VIP (`https://<vip>/horizon` → 200)
- [ ] `/server-status` still answers on loopback; `/root/.apache.cnf` reads `http://127.0.0.1:8080/server-status?auto`; monasca-agent logs show no apache connection errors

HA (multi-node):
- [ ] From control node B: `curl --max-time 5 http://<node-a-mgmt-ip>:8080/server-status?auto` → refused, **and** cluster health stays green with no Apache/Monasca alarms

If the HA check breaks something, that invalidates the design's core premise — fall back to keeping `0.0.0.0` and relying on the welcome-page removal alone, which still satisfies the ticket's acceptance criterion.

#### Additional documentation

The reasoning here is recorded as a handbook kb Topic — [bigstack-handbook#210](https://github.com/bigstack-oss/bigstack-handbook/pull/210), landing `kb/cubecos/architecture/httpd-8080-loopback-hardening.md`. It covers why 8080 can't be removed, why the `welcome.conf` deletion and the `Require all denied` docroot block are one change, and the load-bearing "no cross-node `:8080`" assumption plus its fallback.

```docs

```

### DIFF
--- a/core/keystone/keystone.mk
+++ b/core/keystone/keystone.mk
@@ -14,5 +14,9 @@ rootfs_install::
 	$(Q)cp -f $(ROOTDIR)/etc/keystone/keystone.conf $(ROOTDIR)/etc/keystone/keystone.conf.def
 	$(Q)[ -f $(ROOTDIR)/etc/httpd/conf.d/ssl.conf ] && mv $(ROOTDIR)/etc/httpd/conf.d/ssl.conf $(ROOTDIR)/etc/httpd/conf.d/ssl.conf.orig || /bin/true
 	$(Q)$(INSTALL_DATA) $(ROOTDIR) $(COREDIR)/keystone/ssl.conf ./etc/httpd/conf.d/
+# httpd ships a default welcome page; drop the config and its html so port 8080
+# serves no default content (cubecos-private#76)
+	$(Q)[ -f $(ROOTDIR)/etc/httpd/conf.d/welcome.conf ] && mv $(ROOTDIR)/etc/httpd/conf.d/welcome.conf $(ROOTDIR)/etc/httpd/conf.d/welcome.conf.orig || /bin/true
+	$(Q)rm -rf $(ROOTDIR)/usr/share/httpd/noindex
 # openssl 3.x requires no /dev/urandom for the written openssl command
 	$(Q)sed -i '/RANDFILE/d' $(ROOTDIR)/usr/libexec/mod_auth_mellon/mellon_create_metadata.sh
--- a/core/modules/config_apache2.cpp
+++ b/core/modules/config_apache2.cpp
@@ -14,6 +14,10 @@
 #include <hex/process.h>
 #include <hex/process_util.h>
 
+// httpd's only consumers are local: haproxy's openstack_horizon backend and the
+// monasca agent's server-status check. Binding loopback keeps 8080 off the
+// management network, where vulnerability scanners were reaching it.
+#define HTTP_ADDR "127.0.0.1"
 #define HTTP_PORT 8080
 
 static const char NAME[] = "httpd";
@@ -65,6 +69,24 @@ WriteSiteConf(const char* hostname, const bool debug)
         "ErrorLog /var/log/httpd/error.log\n",
         "CustomLog /var/log/httpd/access.log combined\n",
         std::string("LogLevel ") + (debug ? "debug" : "info") + "\n",
+
+        // Suppress the version banner here rather than relying on
+        // keystone's ssl.conf, which also sets these. Same values, so the
+        // duplication is inert.
+        "ServerTokens Prod\n",
+        "ServerSignature Off\n",
+        "TraceEnable Off\n",
+
+        // Stock httpd.conf grants Indexes on the DocumentRoot, and welcome.conf
+        // (removed from the image) carried the only "Options -Indexes" covering
+        // "/". DocumentRoot holds nothing -- /var/www is used for certs only --
+        // so deny it outright and serve a bare 403.
+        "<Directory \"/var/www/html\">\n",
+        "    Options -Indexes\n",
+        "    AllowOverride None\n",
+        "    Require all denied\n",
+        "</Directory>\n",
+
         std::string("<VirtualHost *:") + std::to_string(HTTP_PORT) + ">\n",
         "</VirtualHost>\n",
     };
@@ -151,7 +173,8 @@ Init()
 {
     if (HexSystemF(
             0,
-            "sed -e \"s/Listen 80/Listen %d/\" %s > %s",
+            "sed -e \"s/Listen 80/Listen %s:%d/\" %s > %s",
+            HTTP_ADDR,
             HTTP_PORT,
             ORIG,
             CONF)
--- a/core/modules/config_monasca.cpp
+++ b/core/modules/config_monasca.cpp
@@ -147,7 +147,9 @@ WriteAgentSetupDetectionConfigs()
     }
 
     fprintf(fout, "[client]\n");
-    fprintf(fout, "url=http://localhost:8080/server-status?auto\n");
+    // explicit IPv4 loopback: httpd binds 127.0.0.1 only, and "localhost" can
+    // resolve to ::1 first
+    fprintf(fout, "url=http://127.0.0.1:8080/server-status?auto\n");
     fclose(fout);
 
     fout = fopen(AGENT_MYSQL_CONF, "w");


Eandalf-Bigstack:
# Code Review: Apache Default Welcome Page Removal PR

## Critical Issues

### 1. **Sed Command Escaping Risk** ⚠️
**Location:** `config_apache2.cpp:176-177`

```cpp
"sed -e \"s/Listen 80/Listen %s:%d/\" %s > %s",
HTTP_ADDR,
```

**Issue:** The replacement string `%s:%d` is interpolated directly into the sed command without escaping. While `127.0.0.1:8080` is safe, if `HTTP_ADDR` ever contained a forward slash or backslash (e.g., from config changes), the sed command would break. Consider using a safer sed delimiter or proper escaping:

```cpp
"sed -e 's|Listen 80|Listen %s:%d|' %s > %s",  // Using | as delimiter
```

**Severity:** Medium — Low risk with current value, but fragile pattern.

---

### 2. **Directory Block Syntax/Placement Uncertainty**
**Location:** `config_apache2.cpp:72-80`

```cpp
"<Directory \"/var/www/html\">\n",
"    Options -Indexes\n",
"    AllowOverride None\n",
"    Require all denied\n",
"</Directory>\n",
```

**Issues:**
- Placed in `WriteSiteConf()` output before the `<VirtualHost>` block. Verify this is valid Apache config scope — Directory blocks inside VirtualHost are valid, but this block appears *before* VirtualHost declaration, which may be invalid.
- No confirmation the actual DocumentRoot is `/var/www/html`. If the default httpd.conf uses a different path, this directive won't protect the real DocumentRoot.
- The PR states "/var/www is used for certs only," but the protected path is `/var/www/html` — confirm this mismatch is intentional.

**Severity:** High — If syntax is invalid or path is wrong, the security goal fails silently.

---

### 3. **Duplicate Configuration (Version Suppression)**
**Location:** `config_apache2.cpp:73-76`

```cpp
"ServerTokens Prod\n",
"ServerSignature Off\n",
"TraceEnable Off\n",
```

**Issue:** PR description states "ssl.conf also sets these." If `ssl.conf` is loaded after this config and uses different values, the last one wins. Verify:
- Load order of configs (does `ssl.conf` come after this generated file?)
- Values match exactly
- No conflicting production security headers elsewhere

**Severity:** Low — Likely harmless, but should be consolidated to one location.

---

### 4. **Makefile Race Condition**
**Location:** `keystone.mk:17-18`

```makefile
$(Q)[ -f $(ROOTDIR)/etc/httpd/conf.d/welcome.conf ] && mv $(ROOTDIR)/etc/httpd/conf.d/welcome.conf $(ROOTDIR)/etc/httpd/conf.d/welcome.conf.orig || /bin/true
$(Q)rm -rf $(ROOTDIR)/usr/share/httpd/noindex
```

**Issue:** The `rm -rf` uses a glob pattern implicitly (entire directory). If `/usr/share/httpd/` is a symlink or mount point in some build scenarios, this could delete unintended files. Consider:
- Explicit path check before removal
- Use `rm -rf` safely with validation

**Severity:** Low — Unlikely in normal build, but risky pattern.

---

## Non-Critical Observations

✅ **Monasca fix is sound:** Pinning to explicit `127.0.0.1` is correct (avoiding IPv6 resolution race).

✅ **Loopback-only binding:** Reduces attack surface appropriately.

⚠️ **Unverified in real environment:** PR acknowledges no node testing. Pre-merge verification checklist in PR description should **must** pass before merge.

---

## Verification Blockers

Before merge, confirm:
1. **Syntax check:** Verify generated Apache config passes `apachectl configtest`
2. **DocumentRoot path:** Confirm actual default DocumentRoot matches `/var/www/html`
3. **Sed expansion:** Run with real values and verify output
4. **No regression:** Horizon and server-status still functional on loopback

Eandalf-Bigstack:
This is a solid approach to closing the port 8080 exposure. The two-layer defense (binding to loopback and actively denying the empty docroot) is exactly the right strategy here, and pinning Monasca to the IPv4 loopback is a smart catch to avoid the `::1` resolution race condition.

I’ve reviewed the PR diff against the codebase context and the prior reviewer’s comments. Here is my assessment:

## Clarifications on Prior Review Comments

Let's quickly address the previous reviewer's concerns, as some of their premises need slight adjustments:

* **Sed Command Escaping:** While using `|` as a delimiter is a good defensive practice, `HTTP_ADDR` is defined locally as a macro (`#define HTTP_ADDR "127.0.0.1"`). Since it's hardcoded and controlled within this file, the risk of a rogue slash is virtually zero.
* **Directory Block Syntax:** The reviewer was concerned about the `<Directory>` block sitting outside the `<VirtualHost>`. This is **perfectly valid** Apache configuration. Placing it in the global server context ensures the protection applies globally to `/var/www/html` regardless of how virtual hosts are resolved.
* **Duplicate Configuration:** Re-declaring `ServerTokens Prod` here is actually safer. If `ssl.conf` is ever modified or fails to load, this file guarantees the information leak is still plugged on port 8080.
* **Makefile `rm -rf`:** Since this executes within a controlled `ROOTDIR` build staging environment, standard glob removal of the `noindex` directory is standard practice and safe.

---

## New Architectural & Logic Findings

However, there are two subtle but important flaws in the current diff that should be fixed before merge:

### 1. Incomplete Loopback Pinning in VirtualHost

**Location:** `config_apache2.cpp:88` (approx)

```cpp
std::string("<VirtualHost *:") + std::to_string(HTTP_PORT) + ">\n",

```

While the `Listen` directive correctly restricts the socket to `127.0.0.1`, leaving the VirtualHost bound to `*` creates a fragile configuration. If another service or future commit ever re-adds a wildcard `Listen` directive (e.g., `Listen 8080`), this VirtualHost will instantly become exposed on all interfaces again.

**Fix:** Make the VirtualHost declaration symmetric with the bind address:

```cpp
std::string("<VirtualHost ") + HTTP_ADDR + ":" + std::to_string(HTTP_PORT) + ">\n",

```

### 2. Regex Boundary Flaw in the `sed` Command

**Location:** `config_apache2.cpp:176`

```cpp
"sed -e \"s/Listen 80/Listen %s:%d/\" %s > %s",

```

The regex `s/Listen 80/` is unanchored. If the stock `httpd.conf` ever ships with `Listen 8080` (or if this script runs idempotently on an already modified file), the replacement will yield `Listen 127.0.0.1:808080`, breaking Apache completely.

**Fix:** Add boundaries or anchors to ensure you are only matching exactly port 80.

```cpp
"sed -e 's/^Listen 80\\b/Listen %s:%d/' %s > %s",

```

*(Note the switch to single quotes for the sed script to make escaping the boundary `\b` cleaner in C++).*

Eandalf-Bigstack:
## Review: `fix(apache): remove default welcome page on port 8080`

### Overview

Closes a compliance-scan finding (Tenable 112351 / CIS / PCI-DSS 2.2.1 / NIST CM-6/CM-7): Apache's stock "Test Page" is reachable on port 8080 of the management IP. Three coordinated changes across 3 files (+31/-2):

1. **`config_apache2.cpp`** — binds httpd's 8080 listener to `127.0.0.1` only (via the `sed` that rewrites `Listen 80` → `Listen 127.0.0.1:8080`), adds a `<Directory "/var/www/html">` block denying access and disabling directory indexing, and adds `ServerTokens Prod` / `ServerSignature Off` / `TraceEnable Off`.
2. **`keystone.mk`** — removes `welcome.conf` (renamed to `.orig`, following the existing `ssl.conf.orig` pattern) and deletes `/usr/share/httpd/noindex` from the built rootfs.
3. **`config_monasca.cpp`** — repoints the Monasca `server-status` health check from `http://localhost:8080` to `http://127.0.0.1:8080`, since `localhost` can resolve to `::1` where httpd no longer listens.

### Correctness

- Verified against surrounding code, not just the PR's own claims: `config_haproxy.cpp:113` already has `server localhost 127.0.0.1:8080 check` for the `openstack_horizon` backend — HAProxy was already loopback-only, so the httpd bind change doesn't break Horizon routing.
- The `<Directory>` deny block is genuinely necessary, not just belt-and-suspenders: stock `httpd.conf` grants `Options Indexes` on `DocumentRoot`, and `welcome.conf` was the only file carrying `-Indexes` for `/`. Removing `welcome.conf` without this block would trade a version-disclosing test page for a full directory listing — worse. Correctly reasoned in the code comment.
- `ServerTokens`/`ServerSignature`/`TraceEnable` are already set identically in `core/keystone/ssl.conf` (verified) — the PR's own comment calls this out as inert duplication rather than pretending it's new.
- The `sed -e "s/Listen 80/Listen %s:%d/"` pattern is pre-existing (only the `%s:` prefix is new); not a regression introduced here.
- `/identity` (Keystone) is exposed via a global `Alias`/`Location`, not scoped to a specific `Listen`/vhost, per `keystone-wsgi.conf.in` — so it rides on whatever `Listen 80`→8080 resolves to, consistent with the PR's claim that Keystone `/identity` is one of the two live 8080 consumers.

### Risk — the big one

The PR body itself flags this: **not verified on a real node**, with an explicit pre-merge checklist and a documented fallback plan. That's the right way to submit an unverified infra change, but:
- The HA claim ("no service depends on cross-node :8080") is architecturally plausible from static analysis (HAProxy already uses loopback) but unconfirmed for actual multi-node topologies — e.g., if any deployment runs HAProxy and httpd on *different* hosts (not colocated with a VIP), the loopback bind would break Horizon for that topology. Worth confirming this colocation assumption holds for all supported cluster layouts.
- `Require local` in `writeStatusConf` (existing code, unchanged) already permits loopback — so after this change the explicit `Require ip <ctrlIp>` entries for `/server-status` become mostly redundant for the *local* check, though they may still matter if some other design intends cross-node status polling that this PR silently forecloses. Worth a sentence confirming no code path polls `/server-status` cross-node.

**This should not be merged before the checklist in the description passes on a real jail/node** — already explicit in the description; I'd treat it as a hard gate, not a suggestion.

### Style / conventions

- Comments are precise and justify *why*, not *what* — matches the codebase's existing terse style.
- The `welcome.conf` → `welcome.conf.orig` rename mirrors the existing `ssl.conf` → `ssl.conf.orig` handling in the same Makefile target, reusing the `[ -f ... ] && mv ... || /bin/true` idiom rather than inventing a new pattern.
- `HTTP_ADDR` as a `#define` next to `HTTP_PORT` is consistent with the file's existing style.

### Test coverage

No unit/integration tests exist for this class of config-generation module in the repo, consistent with project norms; the PR compensates with a detailed manual verification checklist, appropriate given the infra nature of the change.

### Security

Net positive: closes an actual scan finding, removes a directory-listing exposure that would otherwise be introduced, and uses consistent IPv4-only reasoning end-to-end (no resolver dependency at boot) — binding via hostname `localhost` would have silently widened the surface back open on `::1`. No secrets or auth logic changed; `Require all denied` + `Options -Indexes` is a correct, minimal-footprint deny-by-default posture for an unused DocumentRoot.

### Suggestions

1. Before merge: run the checklist in the PR body against a real node, particularly the multi-node HA check — the one item that can't be validated by static review.
2. Consider a one-line comment near `writeStatusConf`'s `Require ip <ctrlIp>` noting it's now belt-and-suspenders given loopback-only access, so a future reader doesn't wonder if it's load-bearing for cross-node access.

---
*Review generated with [Claude Code](https://claude.com/claude-code)*

Eandalf-Bigstack (approved):

Issue #960, opened 2026-06-15
[Epic] CubeCOS Deployment Matrix — x86/ARM × Full-IaaS/edge-k3s (four use cases)


traviswu-bigstack:
## Vision

One CubeCOS codebase, **two reusable building blocks** — **architecture support** (#962) and a
**service profile** (#963) — unlock a 2×2 matrix of **four distinct market use cases**. We invest
once in each building block and harvest four products. v1 targets the highest-growth cell
(Jetson edge-AI); the other cells fall out of the same two investments.

## Why CubeCOS — the core thesis

CubeCOS is an **opinionated, batteries-included OS for cloud infra**: it solves the *complexity*
problem of open infrastructure. Even the lean edge-k3s profile ships **k3s + Ceph + observability +
simplified day-2 ops** pre-integrated as one product under one support throat — the operator never
assembles or babysits the stack (iPhone, not a parts bin). The differentiator is **one operating
model end to end** — the same opinionated platform from datacenter (UC-1/UC-3) to edge (UC-2/UC-4) —
which the pure-edge vendors can't match.

**Where the thesis is strongest vs. contested:**
- **Strongest in the Full-IaaS column (UC-1/UC-3):** OpenStack complexity is brutal and credible
  "opinionated IaaS appliance" rivals are few (Platform9, Mirantis, Canonical Sunbeam).
- **Most contested in the edge-k3s column (UC-2/UC-4):** baseline complexity is lower and the field
  of simplifiers is dense and "good enough" — Canonical (MicroCloud / MicroK8s; note L4T *is*
  Ubuntu), SUSE Edge (RKE2 / Longhorn / Elemental / Fleet), Talos + Omni, Spectro Cloud Palette,
  Red Hat MicroShift / OpenShift+ODF. So at the edge we win on the **end-to-end continuity** above,
  not on feature-for-feature simplicity.
- **Watch-outs:** *opinionated = rigid* — edge-AI buyers live deep in NVIDIA's specific stack, so our
  opinions must match theirs or "simple" reads as "limiting"; **bundling Ceph at the edge** is in
  tension with the footprint go/no-go (rivals chose lighter storage deliberately); and **integration
  is a treadmill, not a moat** — every upstream bump (k3s/Ceph/JetPack) is our test burden, which a
  small vendor re-absorbs the very complexity it sells away.

## The matrix

| | Full IaaS profile | edge-k3s profile (#963) |
|---|---|---|
[...]

PR #1114, opened 2026-07-15
feat(manila): multi-backend share types (CephFS-native + NetApp stub)

steven-chiu-bigstack:
Close bigstack-oss/cubecmp#957

## Summary

COS-side implementation backing bigstack-oss/cubecmp#957 (spike: bigstack-oss/cubecmp#960). Adds two new optional Manila share backends, both disabled by default:

- **CephFS-native** (fully wired): dedicated `manila_cephfsnative` Ceph filesystem + `client.manila` cephx identity, `[cephfsnative]` manila.conf driver stanza, `CEPHFS` added to `enabled_share_protocols` (a real bug spike #960 caught — Manila rejects CephFS shares without it), and a bound `cephfs_share_type` created on every cluster start.
- **NetApp** (config-surface stub only): `[netapp]` stanza + `netapp_share_type`, not functionally validated — no real array available to test against.

Both toggle via `manila.cephfsnative.enabled` / `manila.netapp.enabled` (default `false`) — zero behavior change for existing clusters until an admin opts in.

## What changed

- `core/modules/config_manila.cpp` — new tunings, `SetDriverCephFSNative`/`SetDriverNetapp`, dynamic `enabled_share_backends`/`enabled_share_protocols`, `CONFIG_REQUIRES(manila, ceph)`, `ClusterStartMain` now also runs share-type binding on every start (not just first boot).
- `core/modules/config_ceph.cpp` — cross-reads the manila toggle, provisions the dedicated filesystem + cephx client, gates the pre-existing (previously unconditional, `FIXME`-marked) `[client.manila]` ceph.conf stanza.
- `core/sdk_sh/modules/sdk_os.sh` — new `os_manila_backend_types_init`, idempotent, creates+binds a share type per enabled backend via the `share_backend_name` extra-spec (the exact mechanism spike #960 verified end-to-end).

## Notable fixes caught during review

- A cross-module "modified" tracking variable in `config_ceph.cpp` was initially left unwired from `CommitCheck()`, which would have silently no-op'd a toggle flip on an already-running cluster — fixed and reconfirmed.
- Final whole-branch review flagged the `client.manila` cephx `mds` cap as `allow *` (unscoped, defeating the point of a dedicated filesystem) and the NetApp password tuning as publicly readable — both fixed and reconfirmed.

## Test plan

No automated test harness exists for hex config C++ modules in this repo (confirmed: only `config_etcd` has a bespoke rig). Verification is manual, against a live dev cluster, per the plan's Task 6:

- [ ] `cubectl config commit` with both toggles at default `false` — confirm zero diff to `manila.conf`/`ceph.conf`/`openstack share type list` vs. pre-branch baseline
- [ ] Enable `manila.cephfsnative.enabled`, commit, confirm `openstack share type list` shows `cephfs_share_type`, `openstack share pool list` shows the `cephfsnative` backend
- [ ] Create + mount a share on `cephfs_share_type`, write/read to confirm the data path
- [ ] Toggle back off, confirm revert to baseline

## Out of scope

CMP-side UI/API/billing (separate `cubecmp` repo/story) and the handbook knowledge update (DoD, lands in `bigstack-handbook`) are tracked separately, not part of this PR.

### DIFF
--- a/core/modules/config_ceph.cpp
+++ b/core/modules/config_ceph.cpp
@@ -689,9 +689,9 @@ SetupManilaCephFS()
     HexUtilSystemF(0, 0, "timeout 10 ceph fs set %s allow_new_snaps true", MANILA_CEPHFS_NAME);
 
     HexUtilSystemF(0, 0, "timeout 10 ceph auth get-or-create client.manila "
-                         "mon 'allow r' mgr 'allow rw' mds 'allow *' "
+                         "mon 'allow r' mgr 'allow rw' mds 'allow rw fsname=%s' "
                          "osd 'allow rw pool=%s, allow rw pool=%s' -o %s",
-        MANILA_CEPHFS_DATA_POOL, MANILA_CEPHFS_METADATA_POOL, MANILA_CEPH_KEYRING);
+        MANILA_CEPHFS_NAME, MANILA_CEPHFS_DATA_POOL, MANILA_CEPHFS_METADATA_POOL, MANILA_CEPH_KEYRING);
     HexUtilSystemF(0, 0, "chmod 0600 %s", MANILA_CEPH_KEYRING);
 
     HexSystemF(0, "touch " MAKRER_MANILA_CEPHFS);
--- a/core/modules/config_ceph.cpp
+++ b/core/modules/config_ceph.cpp
@@ -1471,6 +1471,7 @@ CommitCheck(bool modified, int dryLevel)
         | s_bNetModified
         | s_bCubeModified
         | s_bKeystoneModified
+        | s_bManilaModified
         | G_MOD(IS_MASTER)
         | G_MOD(CTRL_IP)
         | G_MOD(SHARED_ID)
--- a/core/modules/config_ceph.cpp
+++ b/core/modules/config_ceph.cpp
@@ -33,6 +33,7 @@
 
 #define MAKRER_POOL "/etc/appliance/state/ceph_osd_pool_done"
 #define MAKRER_CEPHFS "/etc/appliance/state/cephfs_done"
+#define MAKRER_MANILA_CEPHFS "/etc/appliance/state/manila_cephfs_done"
 #define MAKRER_RESTFUL "/etc/appliance/state/ceph_restful_done"
 #define MAKRER_AUTOSCALE "/etc/appliance/state/ceph_autoscale_done"
 #define MAKRER_MSGR2 "/etc/appliance/state/ceph_msgr2_done"
@@ -66,6 +67,7 @@ const static char DUMMY_KEYRING[] = "/etc/ceph/ceph.client.None.keyring";
 const static char ADMIN_KEYRING[] = "/etc/ceph/ceph.client.admin.keyring";
 const static char K8S_KEYRING[] = "/etc/ceph/ceph.client.k8s.keyring";
 const static char CEPHFS_CLIENT_AUTHKEY[] = "/etc/ceph/admin.key";
+const static char MANILA_CEPH_KEYRING[] = "/etc/ceph/ceph.client.manila.keyring";
 
 static const char FSID[] = "c6e64c49-09cf-463b-9d1c-b6645b4b3b85";
 
@@ -79,6 +81,10 @@ static const char CEPHFS_METADATA_POOL[] = "cephfs_metadata";
 static const char CEPHFS_NAME[] = "cephfs";
 static const char CEPHFS_DEFAULT_SUB_VOLUME_GROUP[] = "cephfs_default_sub_volume_group";
 static const char CEPHFS_STORE_DIR[] = "/mnt/cephfs";
+
+static const char MANILA_CEPHFS_DATA_POOL[] = "manila_cephfs_data";
+static const char MANILA_CEPHFS_METADATA_POOL[] = "manila_cephfs_metadata";
+static const char MANILA_CEPHFS_NAME[] = "manila_cephfsnative";
 
 static const char SRV_KEY[] = "/var/www/certs/server.key";
 static const char SRV_CRT[] = "/var/www/certs/server.cert";
@@ -136,6 +142,7 @@ CONFIG_TUNING_SPEC_STR(CUBESYS_SEED);
 CONFIG_TUNING_SPEC_BOOL(CUBESYS_HA);
 CONFIG_TUNING_SPEC_BOOL(CUBESYS_SALTKEY);
 CONFIG_TUNING_SPEC_STR(KEYSTONE_ADMIN_CLI_PASS);
+CONFIG_TUNING_SPEC_BOOL(MANILA_CEPHFSNATIVE_ENABLED);
 
 // parse tunings
 PARSE_TUNING_BOOL(s_debugEnabled, CEPH_DEBUG_ENABLED);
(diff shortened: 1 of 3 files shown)


traviswu-bigstack (commented):
The way of implement is not fully reviewed, but the reason of why this story is requied should be provided. Making netapp be the offical supported manila backen breaks the baseline we are following. and why do we need cephfs as another manila backend

Eandalf-Bigstack:
I'll analyze this PR systematically, examining the code changes against the stated goals and architecture.

● Check repo context (shell)
  │ cd /Users/jimlin/Documents/bigstack/repo/cubecos && git log --oneline -1
  └ 2 lines…

## Code Review: Multi-Backend Manila Share Types

### ✅ Strengths
- **Security fixes confirmed**: NetApp password marked `TUNING_UNPUB`, CephFS cephx mds cap properly scoped to filesystem (addressing spike #960)
- **Idempotency patterns solid**: Marker files (`MAKRER_MANILA_CEPHFS`), type-existence checks prevent double-creation
- **Clean separation**: Backend drivers elegantly parameterized; toggles default to `false`
- **Dependency tracking**: `CONFIG_REQUIRES(manila, ceph)` and `s_bManilaModified` wiring ensures config change detection

---

### 🔴 Critical Issues

**1. Error Handling in `SetupManilaCephFS()` – Silent Failures**
```cpp
HexUtilSystemF(0, 0, HEX_SDK " ceph_create_pool %s cephfs", MANILA_CEPHFS_DATA_POOL);
HexUtilSystemF(0, 0, HEX_SDK " ceph_create_pool %s cephfs", MANILA_CEPHFS_METADATA_POOL);
HexUtilSystemF(0, 0, "timeout 10 ceph fs new %s %s %s", ...);  // Could fail if pools don't exist
```
- **Risk**: Pool creation fails silently; filesystem creation proceeds with missing pools → cluster starts degraded, operator unaware
- **Fix**: Check return values; fail fast if any step fails

**2. Protocol Replacement Bug in `config_manila.cpp`**
```cpp
if (s_cephfsnativeEnabled) {
    config["DEFAULT"]["enabled_share_protocols"] = "NFS,CIFS,CEPHFS";  // Overwrites, doesn't append
}
```
- **Risk**: If admin configured custom protocols elsewhere, enabling CephFS silently discards them
- **Fix**: Append CEPHFS to existing value or merge sets

**3. NetApp Validation Absent**
- No checks that `s_netappServer`, `s_netappUsername`, `s_netappPassword`, `s_netappVserver`, `s_netappAggregate` are all set
- **Risk**: Enabling NetApp without required credentials → Manila fails to start, operator sees cryptic error
- **Fix**: In `Validate()` or early in `Commit()`, require all NetApp fields if `s_netappEnabled == true`

**4. Fragile Manila Type Detection in `sdk_os.sh`**
```bash
local cephfs_type_id=$(manila type-list | awk '/ cephfs_share_type /{print $2}')
```
- **Risk**: CLI output format can change; whitespace-sensitive awk pattern fails silently, type silently never created
- **Fix**: Use `manila type-show cephfs_share_type 2>/dev/null | grep -q id` or similar error-aware check

**5. No Error Propagation in Shell Script**
```bash
manila type-create ... 2>&1
manila type-key ... 2>&1
```
- **Risk**: If either command fails, script continues silently; operator unaware type wasn't bound
- **Fix**: Add `|| return 1` or `set -e` within the function

---

### ⚠️ Minor Issues

- **Unscoped credentials in memory**: `SetDriverNetapp()` stores plaintext password in config dict written to disk. Already marked `TUNING_UNPUB`, but verify `WriteConfig()` doesn't log/debug-dump the dict.
- **Marker file placement**: `MAKRER_MANILA_CEPHFS` touched only after auth key created; if `chmod` fails, marker file not written, but pools + fs already exist → idempotency broken.

---

### 🧪 Test Coverage Gap
PR acknowledges no automated tests exist. Manual test plan in description is thorough, but **recommend**:
- Test failure scenario: pool creation fails mid-setup (simulate with `ceph fs new` against nonexistent pool)
- Test toggle flip: disable CephFS after enabling, confirm clean revert (check filesystem doesn't stay bound to shares)
- Test NetApp with missing credentials: enable without setting server/vserver, confirm clear error message during commit

---

**Recommendation**: Address error handling (issues 1–2) and validation (issue 3) before merge. Issues 4–5 are higher risk given OpenStack CLI brittleness.

Eandalf-Bigstack:
Eandalf-Bigstack's preliminary analysis is incredibly thorough and spot on. I agree with all the points raised in the prior discussion and have found an additional architectural oversight regarding state management during this diff review.

Here is the consolidated senior review:

## Pull Request Review: Multi-Backend Manila Share Types

### 🚨 Critical Issues & Bugs

* **State Inconsistency on Disable (New Finding):** If an admin toggles `manila.cephfsnative.enabled` or `manila.netapp.enabled` from `true` back to `false`, the backend drivers are cleanly removed from `manila.conf`. However, `os_manila_backend_types_init` does not delete the OpenStack share types. End-users will still see `cephfs_share_type` as available, but provisioning against it will fail with an opaque `No valid host` scheduling error.
* **Protocol Overwrite Bug:** In `config_manila.cpp`, executing `config["DEFAULT"]["enabled_share_protocols"] = "NFS,CIFS,CEPHFS";` hardcodes the protocols. If a cluster administrator has configured custom protocols, this assignment will silently obliterate them.
* **Silent Failures in Ceph Setup:** `SetupManilaCephFS()` executes `ceph fs new` and pool creations via `HexUtilSystemF` without evaluating the return codes. If a timeout occurs, the C++ module still returns `true`, writes the completion marker file, and continues—leaving the storage layer broken but marked as "healthy" by the configuration engine.
* **Missing NetApp Validation:** `SetDriverNetapp()` blindly accepts empty strings for addresses and credentials. Flipping the NetApp toggle without populated tunings will cause the Manila volume service to crash loop on startup.
* **Fragile Shell Scripting:** The `awk '/ cephfs_share_type /{print $2}'` parsing in `sdk_os.sh` is whitespace-dependent. OpenStack CLI table formatting changes or truncation will break this, causing duplicate creation attempts that fail the script.

---

### ⚠️ Minor & Security Observations

* **Plaintext Passwords in Memory:** `s_netappPassword` is passed as a standard `std::string` and stored in the `Configs& config` map. While correctly flagged as `TUNING_UNPUB`, ensure `WriteConfig()` or `config.dump()` does not output the `[netapp]` dict to syslog or stdout during high-verbosity debug modes.
* **Premature Marker File Creation:** `touch MAKRER_MANILA_CEPHFS` is executed regardless of the success of the preceding auth/chmod commands. If `chmod 0600` fails, the marker is written anyway, breaking idempotency for correcting file permissions on subsequent cluster starts.
* **Missing Shell Error Propagation:** OpenStack CLI commands in `os_manila_backend_types_init` lack error handling. Failure to bind the extra-specs (`manila type-key ...`) will go unnoticed by the deployment pipeline, resulting in a share type that isn't actually mapped to the backend.

---

### 📋 Recommended Action Items

1. Implement a teardown block in `os_manila_backend_types_init` to execute `manila type-delete` when enablement toggles are flipped to `0`.
2. Update `config_manila.cpp` to read the existing `enabled_share_protocols` string, check for `CEPHFS`, and safely append it rather than overwriting.
3. Wrap the `HexUtilSystemF` calls in `SetupManilaCephFS()` with conditionals that return `false` on non-zero exit codes to trigger proper config failure states.
4. Add a `Validate()` condition in `config_manila.cpp` that enforces `s_netappServer`, `s_netappUsername`, and `s_netappPassword` are populated if `s_netappEnabled` is true.
5. Refactor the `awk` CLI parsing to use native OpenStack formatting hooks (e.g., `manila type-show cephfs_share_type -c id -f value 2>/dev/null`).

Eandalf-Bigstack:
Please ignore the previous two review comments. They are AI-assisted PR review by using GitHub Copilot and Gemini Pro. What mentioned are either legacy issues, which are not related to this PR, or tedious grammar preferences, and should have very little to do with this PR.

Only two stuff we need to consider, one is the Manila architecture design decision we have discussed in the meeting and pointed out by @traviswu-bigstack in the previous comment. Another one is that Ceph pool creation could happen in `config_manila.cpp` instead of `config_ceph.cpp` to keep Manila service logics away from Ceph, where `config_ceph.cpp` could only take care of the Ceph services.

Issue #689, opened 2026-01-29
[Bug]: Fixpack list didn't show fixpack


chase-chou:
## Describe the bug

Via web update fixpack, but didn't displayed on Fixpack page.

web AP (have total but didn't have info) 

<img width="2531" height="898" alt="Image" src="https://github.com/user-attachments/assets/0f96db21-d380-4db7-ba52-085888462621" />

```
{
    "code": 200,
    "data": {
        "fixpacks": [],
        "page": {
            "total": 1,
            "number": 1,
            "size": 0,
            "totalItemCount": 0
        }
    },
    "msg": "List of fixpacks",
    "status": "ok"
}
```

Check path `/var/fixpack/`
```
[qa200 ~]# ls /var/fixpack/
v3.1.0-rc1_norollback.fixpack
```

## Version 
CUBE_3.1.0_20260128-0651_2bdd69c

## Environment
[qa200](https://10.32.10.200)

## To Reproduce
1. Via web update fixpack
2. Check /var/fixpack/
3. See error

## Expected behavior
Fixpack show on web

jdjgya:
Hi @chase-chou , Thank you for reporting this. 

During the investigation, I realized API can't parse them because

1). The fixpacks you uploaded to the QA200 are `0` size, which does not seem like a valid fixpack


```sh
[qa200 /var/fixpack]# ll /var/fixpack/
total 8
drwxr-xr-x  2 root root 4096 Jan 29 22:12 ./
drwxr-xr-x 29 root root 4096 Jan 26 15:09 ../
-rw-r--r--  1 root root    0 Jan 29 18:14 v3.1.0-rc1_norollback.fixpack
-rw-r--r--  1 root root    0 Jan 29 18:21 v3.1.0-rc1_rollback.fixpack
```

<br/>

2). More importantly, they can't be mounted by

(It's an important part for the API to parse the detailed info and show it on the UI by mounting)

```sh
[qa200 /var/fixpack]# mount v3.1.0-rc1_norollback.fixpack test
mount: /var/fixpack/test: /var/fixpack/v3.1.0-rc1_norollback.fixpack is not a block device; try "-o loop".
```

<br/>

3). I've tried the fixpack we used for the last test, and it parses well.

<br/>

Could you help me check why the fixpacks are all 0 size? 

Thanks a lot 🙏

chase-chou:
Now can sure, it happen on via web import fixpacks

jdjgya:
[...]

PR #1133, opened 2026-07-21
fix(influxdb): require ceph before enabling the influx mgr plugin

### LINKED ISSUE #1132: [Bug] influxdb commits concurrently with ceph: influx plugin never enabled, ceph usage spew breaks e2e
## Description

Under hex's dataflow commit scheduler (hex `15df706`, default-on), `config_influxdb` commits **concurrently with `config_ceph`** at FTS — its only declared prerequisite is `ceph_dashboard_idp`, not `ceph`. `EnableCephInfluxPlugin` then runs while ceph-mgr is still bootstrapping:

1. `ceph mgr module enable influx 2>/dev/null` fails silently — **and the marker is touched anyway**, so the enable is never retried: ceph stats silently never reach influxdb.
2. The three `ceph influx config-set …` calls (interval/hostname/port) hit an unknown command; ceph prints its closest-matches usage (the `pg …` family) to **stderr**, which those lines don't redirect — spraying the serial console mid-FTS.

The console spew also broke the e2e accept job (#689, all four topologies): triangle's fts-done expect fallback `"*> "` matched inside `pg ls-by-pool <poolstr> [<states>…]` and aborted the wait mid-commit (fixed separately in triangle#124).

## How to Reproduce

1. Build develop (with hex ≥ `15df706`) and run the accept e2e, or FTS any topology while watching the serial console.
2. At `committing ▸ ceph influxdb …` the console shows three `no valid command found; 10 closest matches: pg stat …` dumps.
3. After FTS: `ceph mgr module ls | grep influx` → not enabled; influxdb has no `ceph` measurements.

## Expected behavior

influxdb's ceph-plugin setup runs only after the ceph module has committed (mgr up): the enable succeeds, config-set commands are valid, nothing is printed to the console, and a transient enable failure is retried on the next commit instead of being masked by the marker.

## Screenshots (Optional)

None — see accept build #689 console (`centos9_cubecos_accept`).

Root-caused from Jenkins accept #689: last green build #681 predates the hex scheduler bump; builds 682–688 failed in the build stage and never reached e2e. Reproduced the exact spew on the sky lab: any unknown ceph command family prints the `pg` closest-matches list; `config_influxdb` is the only FTS-path caller with unredirected stderr and exactly three calls.

N/A — all topologies (1ec, 2ec_1m, 3cc, 3c_3p_5s) in the QEMU e2e rig.

- [x] **Handbook knowledge update** — bigstack-oss/bigstack-handbook#107 — land the durable, team-readable knowledge from this
      (Candidate: known-issue on module-commit concurrency — modules must declare every
      cross-module runtime dependency via CONFIG_REQUIRES now that commits parallelize.)

traviswu-bigstack:
#### What type of PR is this?

/kind bug

Fixes the FTS regression that has failed the accept e2e since the hex dataflow commit scheduler went default-on (first e2e run to hit it: accept #689, all four topologies).

- **`CONFIG_REQUIRES(influxdb, ceph)`** — the influx ceph-mgr plugin needs a functioning ceph mgr. influxdb only declared `ceph_dashboard_idp`, so the DAG scheduler (correctly, per its declared inputs) committed influxdb while ceph was still bootstrapping. The old level-barrier walk masked the missing edge via link order.
- **Marker only on success** — `ceph mgr module enable influx` failing (mgr not ready) no longer touches the done-marker, so the enable is retried on the next commit instead of being skipped forever (previously: ceph stats silently never reached influxdb after a racy FTS).
- **Silence stderr on the `config-set` calls** — an unknown/failing ceph command prints its closest-matches usage (the `pg …` family) to stderr; unredirected, it sprayed the serial console mid-FTS, which is what tripped triangle's fts-done expect.

Test-side hardening (loose `"*> "` expect fallback) is bigstack-oss/triangle#124 — independent, both should land.

Fixes #1132

> Root cause chain verified empirically: the exact `pg …` closest-matches spew reproduces for any unknown ceph command family (sky lab), three dumps = the three `config-set` calls, stderr-only (matches the unredirected lines), and #681 (pre-scheduler) shows zero spew. `config_influxdb.o` compiles. Worth a follow-up audit: any other module that shells out to `ceph`/`rbd` in Commit but doesn't `CONFIG_REQUIRES` ceph has the same latent race.

#### Additional documentation

```docs
Handbook: candidate known-issue on module-commit concurrency (modules must
declare every cross-module runtime dependency now that commits parallelize).
```

### DIFF
--- a/core/modules/config_influxdb.cpp
+++ b/core/modules/config_influxdb.cpp
@@ -99,7 +99,12 @@ EnableCephInfluxPlugin(bool update, const std::string sharedId)
     HexLogInfo("updating influxdb ceph plugin");
 
     if (stat(MAKRER, &ms) != 0) {
-        HexSystemF(0, "timeout 10 ceph mgr module enable influx 2>/dev/null");
+        // mark done only when the enable actually landed, so a transient
+        // mgr-not-ready failure is retried on the next commit
+        if (HexSystemF(0, "timeout 10 ceph mgr module enable influx 2>/dev/null") != 0) {
+            HexLogWarning("ceph mgr not ready to enable influx plugin; will retry on next commit");
+            return true;
+        }
         HexSystemF(0, "touch " MAKRER);
     }
 
@@ -107,9 +112,9 @@ EnableCephInfluxPlugin(bool update, const std::string sharedId)
     // or master node runs restart when other nodes are down (cluster start)
     // NOTE: timeout of HexSystemF doesn't work in this case
     //       since parent doesn't kill child process when timeout occurs
-    HexSystemF(0, "timeout 10 ceph influx config-set interval %d >/dev/null", INFLUX_INTERVAL);
-    HexSystemF(0, "timeout 10 ceph influx config-set hostname %s >/dev/null", sharedId.c_str());
-    HexSystemF(0, "timeout 10 ceph influx config-set port 8086 >/dev/null");
+    HexSystemF(0, "timeout 10 ceph influx config-set interval %d >/dev/null 2>&1", INFLUX_INTERVAL);
+    HexSystemF(0, "timeout 10 ceph influx config-set hostname %s >/dev/null 2>&1", sharedId.c_str());
+    HexSystemF(0, "timeout 10 ceph influx config-set port 8086 >/dev/null 2>&1");
 
     return true;
 }
@@ -228,6 +233,9 @@ ClusterReadyMain(int argc, char **argv)
 
 CONFIG_MODULE(influxdb, 0, Parse, 0, 0, Commit);
 CONFIG_REQUIRES(influxdb, ceph_dashboard_idp);
+// the influx mgr plugin needs a functioning ceph mgr; without this the
+// dataflow scheduler commits influxdb while ceph is still bootstrapping
+CONFIG_REQUIRES(influxdb, ceph);
 
 // extra tunings
 CONFIG_OBSERVES(influxdb, cubesys, ParseCube, NotifyCube);


Eandalf-Bigstack:
## Code Review: fix(influxdb) - Ceph Dependency and Error Handling

### Summary
This PR addresses a race condition in the influxdb module where the ceph-mgr plugin was being enabled before ceph fully bootstrapped. Changes include proper dependency declaration, conditional marker creation for transient failures, and stderr suppression.

---

### ✅ Strengths

1. **Correct Dependency Declaration** — `CONFIG_REQUIRES(influxdb, ceph)` properly enforces the scheduling constraint. The comment clearly explains why this is needed.

2. **Transient Failure Handling** — The marker is now only created on successful enable, allowing retries on the next commit. This is the right approach for temporary mgr unavailability.

3. **Stderr Redirection** — Changing to `2>&1` prevents stderr noise from the ceph command from polluting the console during parallel commits.

---

### ⚠️ Issues & Observations

1. **Pre-existing Typo Propagated** — The variable `MAKRER` (line 101) appears to be a typo for `MARKER`. This is pre-existing code, but the PR perpetuates it. While not a blocker for this PR, consider filing a follow-up cleanup issue.

2. **Return Value Consistency** — The function returns `true` on both:
   - Success (after marker creation)
   - Transient failure (early return with warning)
   
   Verify that the caller correctly interprets `true` as "continue, don't block" rather than "success." The behavior seems intentional (allowing the module to partially function), but this should be validated against the calling context in `Commit()`.

3. **Submodule Update** — The hex submodule commit hash changed (`5bbd3db...` → `78bde56...`). This should be intentional and unrelated to this fix. Confirm that hex changes are compatible and not pulling in unexpected modifications.

4. **Error Logging Granularity** — The warning is logged on mgr-not-ready, but the three `config-set` failures after a successful enable would fail silently (commands return to `/dev/null 2>&1`). Document whether this is acceptable or if you need to log config-set failures separately.

---

### ✓ Verification Checklist
- [ ] `config_influxdb.o` compiles successfully (mentioned in PR notes—confirm build passes)
- [ ] Caller of `EnableCephInfluxPlugin()` handles `true` return correctly for retries
- [ ] hex submodule changes are intentional and tested
- [ ] FTS regression #1132 is resolved in triangle#124 or depends on it

---

**Recommendation:** Approve with minor concerns noted. The core fixes are sound; verify the submodule and return-value semantics before merge.

Eandalf-Bigstack:
Here is a review of the PR, building on the previous reviewer’s notes and identifying critical new issues.

## Critical Issues

### 1. Command Injection Vulnerability (Security)

In the updated `config-set` commands, `sharedId.c_str()` is injected directly into an unquoted shell execution string:

```cpp
HexSystemF(0, "timeout 10 ceph influx config-set hostname %s >/dev/null 2>&1", sharedId.c_str());

```

If `sharedId` can be influenced by user input, hostnames, or contains any shell metacharacters (e.g., spaces, `;`, `&`, `|`, `$()`), this will result in arbitrary command execution.
**Fix:** Wrap the format specifier in single quotes to safely delimit the string in the shell: `hostname '%s'`.

### 2. Accidental Submodule Bump

The PR includes a commit hash change for the `hex` submodule (`5bbd3dba...` to `78bde560...`).
Unless this PR explicitly relies on a concurrent framework change in the `hex` scheduler, this is likely an accidental inclusion from running `git commit -a`.
**Fix:** Revert the submodule change unless strictly intentional, as it risks pulling in untested framework regressions.

---

## Logic & Architectural Flaws

### 1. False Success in DAG Scheduling

Building on the previous reviewer's observation: returning `true` on a transient failure is dangerous in a DAG-based scheduler.

```cpp
if (HexSystemF(...) != 0) {
    HexLogWarning("ceph mgr not ready...");
    return true; // <-- Tells the DAG "I successfully finished"
}

```

If `influxdb` tells the scheduler it succeeded, the scheduler will immediately unblock any downstream modules that depend on `influxdb`. If those dependents expect the influx plugin to be running, they will crash or fail.
**Fix:** If the module failed to reach its target state, it should return `false` (or the equivalent error signal in your framework) so the DAG scheduler correctly registers a failure/pending state and pauses downstream execution.

### 2. Blind Executions (Silent Failures)

While redirecting `2>&1` to `/dev/null` successfully cleans up the console output, the code entirely ignores the return value of the three `config-set` commands.

```cpp
HexSystemF(0, "timeout 10 ceph influx config-set interval %d >/dev/null 2>&1", INFLUX_INTERVAL);

```

If the Ceph manager restarts or these commands fail, they will fail silently. Furthermore, because these commands sit *outside* the `MAKRER` conditional block, they are blindly executed on every single commit pass, wasting cycles if the state is already reached.
**Fix:** Check the return codes. If they fail, log a localized warning. Consider bringing these configuration commands inside a state-check block so they only run when genuinely necessary.

---

## Minor Observations

* **The `MAKRER` Typo:** As noted in the prior discussion, `MAKRER` is a legacy typo for `MARKER`. Since you are modifying the logic heavily reliant on this macro, this PR is the perfect time to rename it and clean up the technical debt.

Eandalf-Bigstack (approved):

PR #1125, opened 2026-07-21
feat(rolling): consolidated rolling upgrade/restart -- mixed-version gates, one kind-aware roll, drain correctness

traviswu-bigstack:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it

Everything needed to make a **rolling upgrade (Yoga→Epoxy SLURP) and rolling restart safe and operable**, consolidated from the lab-validated working branch. Supersedes #1105 (its 13 commits are the first 7 here, squashed; content byte-identical). 21 commits in 4 layers:

**1. Mixed-version safety gates** (commits 1–7, ex-#1105)
- **gcache 300M→4G** — a rejoining node must catch up via IST; SST is broken across the MariaDB 10.6→10.11 jump. Ships from the Antelope image so 10.6 donors carry it into the A→C roll.
- **`live_migration_gate`** — version-robust functional gate (Galera Synced+Primary, RabbitMQ clustered, ceph PGs servable, nova/cinder/neutron up, ≥1 live target) on **both** relays; never asserts version uniformity, so the mixed N/N+1 window passes. `repair` mode = check-first surgical repairs (formerly `_os_pre_failure_host_evacuation`).
- **Version-uniformity gates** — `mysql_upgrade` and RabbitMQ feature-flag enable deferred until every control node reports the same version (fail-safe: unreachable ⇒ not uniform); masakari maintenance set before a planned drain so hostmonitor doesn't cold-evacuate a deliberately-rebooting node.
- **HA repairs** — vaw always follows vip (新北 RCA: reversed colocation pinned the VIP → 12h outage), quorum-guarded etcd member reseed.
- **Ceph roll guard** — `mon_osd_down_out_interval=3600` during a roll so a rebooting node's OSDs stay `in` (no re-replication churn); OSD-rejoin gate before the next drain.

**2. Boot-path discipline & telemetry** (commits 8–13)
- Don't disruptively repair pacemaker while the cluster is mid-transition (wait for quorum instead); abort a node's bring-up when bootstrap fails instead of marching into a whole-cluster hang.
- Structured `hex_log_event` events (BSP/RUG/CLU) across bootstrap and the roll paths, replacing hand-built influx line protocol (which had real defects: RUG leak, blank UI messages, malformed metadata).
- `admin.key` carried across PPU so cephfs mounts on upgraded nodes; drain retries a rejected migration; running VMs recorded up front so poweroff/powercycle restores them.

**3. Drain correctness** (commits 14–16)
[...]

PR #1136, opened 2026-07-21
feat(rolling): stall watchdog pauses a wedged roll from the VIP holder

### LINKED ISSUE #1135: [Feature] Stall watchdog for rolling update/restart
## Description

A roll whose orchestrator dies freezes silently: the per-node deadline existed in job.json but only the status display read it, and nothing ran periodically — tripwire run 1 wedged at `draining` for 60+ minutes with zero signal, and recovery cost an afternoon (stale job, 8 phantom-ERROR VMs). This feature makes a wedged roll surface in minutes and self-clean:

- every job.json write stamps a **heartbeat** (single choke point), so a live orchestrator can't help but prove liveness;
- a systemd-timer **watchdog** — armed only during a roll, only on control-capable nodes, acting only from the VIP holder — pauses the job past a **load-aware node deadline** (`max(ROLLING_NODE_TIMEOUT, 1500s + 120s × VMs)`) or on heartbeat silence in write-constant phases;
- a trip also runs the per-node **stale-ERROR reconcile** while the drain record is fresh;
- pause/continue move the roll's cluster-wide guards (auto-repair marker, ceph osd-out pin, watchdog) **symmetrically**, closing PR #1125 review issue #2 (orphaned `mon_osd_down_out_interval=3600` after a crash), with a watchdog stale-sweep as backstop.

## Definition of Done (DoD)

- [x] Heartbeat stamped by every job write; watchdog pauses on deadline/heartbeat-silence from the VIP holder only
- [x] Timer ships disabled; armed at roll start on control nodes, re-armed by mid-roll boots, disarmed at completion/abort, self-stops when orphaned
- [x] Deadline scales with VMs on the host; `ROLLING_NODE_TIMEOUT` remains the operator floor
- [x] Pause/continue/abort release/retake all three cluster guards symmetrically; stale osd-out pin swept
- [x] Lab-validated on sky 3cc: synthetic wedge trips (deadline + heartbeat) pause correctly; healthy 3× rolling_update + 1× rolling_restart with watchdog armed stay untripped; reconcile reset branch validated on a real drain failure
- [x] Pause reason survives the event pipeline (comma sanitize — was silently dropped)

## Technical Notes

- **Feature Owner:** travis.wu
- **Frontend:** none (pause events reach the existing UI event stream)
- **Backend:** cubecos sdk_power.sh + systemd units
- **Infra:** none — no new services outside a roll window

## UI/UX Mockups (Optional)

None

- [x] **Handbook knowledge update** — bigstack-oss/bigstack-handbook#108 (design doc: watchdog + guard bracket + reconcile sections; runbook: enforced stalls + pause recovery; deck refreshed)

traviswu-bigstack:
#### What type of PR is this?

/kind feature

A roll whose orchestrator dies freezes silently — the per-node deadline was only read by the status display, and nothing ran periodically to look (tripwire run 1 wedged at `draining` with zero signal; recovery cost an afternoon). Two commits:

1. **Stall watchdog** — every `job.json` write stamps `.heartbeat` (one choke point in `_power_roll_write`); a systemd-timer watchdog enforces the node deadline (any phase) and heartbeat silence (write-constant phases only: draining/staging) from the VIP holder, pausing the job with a reason and running the per-node stale-ERROR reconcile while the drain record is fresh. The deadline is load-aware (`max(ROLLING_NODE_TIMEOUT, 1500s + 120s × VMs on the host)`, floor 2700s) so a big drain isn't false-paused. The timer ships **disabled** and exists only during a roll, only on control-capable nodes: armed at start (runtime start, never enable), re-armed by a mid-roll control-node boot, disarmed on completion/abort, and every tick self-stops its own timer when no running/paused roll exists — before any VIP gating, so non-VIP leftovers can't tick forever. Pause/continue also now release/retake the roll's cluster guards **symmetrically** (auto-repair marker, `mon_osd_down_out_interval` pin, watchdog) — closing PR #1125 review issue #2 (orphaned ceph config after a crash) — with a VIP-holder stale-sweep as the backstop.
2. **Pause-reason sanitize** — `hex_log_event` attrs are comma-separated k=v; a comma in the pause reason (they essentially all have one) sheared the record and the pipeline dropped it silently — proven on the sky deploy roll, where the emitted `CLU00004W` never reached the events DB. Same `${var//[,$'\n']/ }` treatment as `CLU00006W`.

3. **Bootstrap console/status cleanup** (`refactor(bootstrap)`) — resolve the kernel console list once (was ~26 `hex_sdk` spawns per boot, two per status line) and fold the repeated status line into a `bootstrap_status` helper. Fixes it also carried: 79-col filler + home-print-home so statuses update one console line in place on real 80-col consoles (previously each status wrapped and walked down a row, text indented at the wrap point); `tee` instead of an ambiguous `>$(...)` redirect (silently failed with >1 `console=` device); `/run/cube_bootstrap.log` becomes a plain per-boot transcript (was truncated per write, one CR-riddled line); `HOSTNAME` from `$(hostname)` once (the old `ssh localhost hostname` yielded an empty hostname whenever sshd wasn't up — blank `host=` on boot events) with stray `$(hostname)` subshells unified onto it; and the early influxdb restart dropped — events moved to `hex_log_event` (journal + filebeat backfill tolerate a late sink) and influxdb comes up via its own config module during the commit.

Lab validation (sky 3cc, 2026-07-21): synthetic deadline + heartbeat wedges pause correctly with proper reasons; non-VIP nodes and done-state no-op; 3× rolling_update + 1× rolling_restart ran with the watchdog armed and untripped; the reconcile's reset branch had its first production execution on a real drain failure (neutron blip → phantom ERROR → reset → VM live-migrated cleanly 90s later).

Fixes #1135

- The watchdog acts only from the VIP holder — reusing the cluster's existing leader election rather than inventing a second one. Trade-off discussed: if the VIP is unheld the watchdog is blind, but that state is a cluster-wide control-plane outage already loudly visible elsewhere; a reach-the-VIP fallback was deliberately deferred.
- The reconcile reset requires "domain running on exactly the host nova records" — the signature of a migration that died pre-switchover. Post-copy failures leave both sides paused (excluded by `--state-running`) and split-brain shows two hosts (excluded by host-match), so the damaged-VM modes classify themselves out.
- `Quiet -n systemctl stop` inside the watchdog (vs the cluster-wide `_power_roll_watchdog_disarm` used by lifecycle events) is deliberate: each tick self-stops its own timer — no N×N remote_run from a 60s tick path.

#### Additional documentation

```docs
bigstack-oss/bigstack-handbook#108 — design doc (watchdog, guard bracket, reconcile,
one-roll-two-kinds), runbook (enforced stalls + pause recovery), deck refresh.
```

### DIFF
--- a/core/main/bootstrap_cube_config
+++ b/core/main/bootstrap_cube_config
@@ -12,31 +12,38 @@ fi
 
 systemctl mask openvswitch # openvswitch shouldn't be brought up by pacemaker commit
 
-export HOSTNAME=$(ssh localhost hostname 2>/dev/null)
-# Bring up influxdb early so boot-time health/metrics have a sink
-hex_sdk cmd -c "systemctl show -p SubState influxdb | grep -q -i running || systemctl restart influxdb"
-space_filler="                                                                                                    "
+# from the kernel: ssh-localhost returned empty whenever sshd wasn't up yet
+export HOSTNAME=$(hostname)
+# 79 cols fits an 80-col console row, so the status rewrites in place
+space_filler=$(printf '%79s' '')
+# console list is fixed at boot -- resolve once
+KERNEL_CONSOLES=$(hex_sdk GetKernelConsoleDevices)
+# console: blank, home, print, home (one write); log: plain line per status.
+# tee, not >$(...): a multi-device redirect target is ambiguous and fails.
+bootstrap_status()
+{
+    local msg="$1"
+    echo "$msg" >> $BOOTSTRAP_CUBE_LOG
+    echo -n "$space_filler"$'\r'"$msg"$'\r' | tee $KERNEL_CONSOLES >/dev/null 2>&1
+}
 BootstrapAndClusterstart()
 {
     if [ ! -e "$BOOTSTRAP_CUBE_MARKER" ] ; then
-        msg="Bootstrapping CubeCOS"
-        echo -n "$space_filler">$(hex_sdk GetKernelConsoleDevices) ; echo -n "$msg" $'\r' | tee $BOOTSTRAP_CUBE_LOG $(hex_sdk GetKernelConsoleDevices)
+        bootstrap_status "Bootstrapping CubeCOS"
         hex_log_event -e BSP00001I "interface=system,host=$HOSTNAME,service=bootstrap,phase=bootstrapping"
         hex_sdk power_bootup_mark bootstrapping   # cluster bootup status: entering the config commit
-        [ -e /store/rolling_recover ] && hex_sdk _power_roll_set_phase_ts "$(hostname)" bootstrapping   # rolling_restart phase timing
-        bash -c "hex_config -p bootstrap standalone-done | tee $BOOTSTRAP_CUBE_LOG $(hex_sdk GetKernelConsoleDevices)"
+        [ -e /store/rolling_recover ] && hex_sdk _power_roll_set_phase_ts "$HOSTNAME" bootstrapping   # rolling_restart phase timing
+        bash -c "hex_config -p bootstrap standalone-done | tee -a $BOOTSTRAP_CUBE_LOG $KERNEL_CONSOLES"
         if [ ! -e "$BOOTSTRAP_CUBE_MARKER" ] ; then
-            msg="Bootstrap error: $(journalctl | grep hex | egrep -i 'failed to commit')"
-            echo -n "$space_filler">$(hex_sdk GetKernelConsoleDevices) ; echo -n "$msg" $'\r' | tee $BOOTSTRAP_CUBE_LOG $(hex_sdk GetKernelConsoleDevices)
+            bootstrap_status "Bootstrap error: $(journalctl | grep hex | egrep -i 'failed to commit')"
             hex_log_event -e BSP00003W "interface=system,host=$HOSTNAME,service=bootstrap,detail=$(journalctl | grep hex | egrep -i 'failed to commit|took' | tail -2 | tr -d ',\"=|' | tr '\n' ' ')"
             return 1
         else
-            msg="Starting cluster"
-            echo -n "$space_filler">$(hex_sdk GetKernelConsoleDevices) ; echo -n "$msg" $'\r' | tee $BOOTSTRAP_CUBE_LOG $(hex_sdk GetKernelConsoleDevices)
+            bootstrap_status "Starting cluster"
             hex_log_event -e BSP00002I "interface=system,host=$HOSTNAME,service=bootstrap,phase=cluster_start"
             hex_sdk power_bootup_mark finalizing   # cluster bootup status: entering cluster_start (+ master check_repair)
-            [ -e /store/rolling_recover ] && hex_sdk _power_roll_set_node_status "$(hostname)" finalizing   # rolling_restart: status + phase ts (until power_roll_advance -> done)
-            bash -c "hex_sdk -v cube_cluster_start | tee $BOOTSTRAP_CUBE_LOG $(hex_sdk GetKernelConsoleDevices)"
+            [ -e /store/rolling_recover ] && hex_sdk _power_roll_set_node_status "$HOSTNAME" finalizing   # rolling_restart: status + phase ts (until power_roll_advance -> done)
+            bash -c "hex_sdk -v cube_cluster_start | tee -a $BOOTSTRAP_CUBE_LOG $KERNEL_CONSOLES"
             return 0
         fi
     fi
@@ -46,8 +53,7 @@ BootstrapAndClusterstart()
 # turns one failed module into a whole-cluster hang on the ready-wait
 AbortBootstrap()
 {
-    msg="Bootstrap failed on $(hostname); aborting cluster bring-up. Fix the failing module, then re-run."
-    echo -n "$space_filler">$(hex_sdk GetKernelConsoleDevices) ; echo -n "$msg" $'\r' | tee $BOOTSTRAP_CUBE_LOG $(hex_sdk GetKernelConsoleDevices)
+    bootstrap_status "Bootstrap failed on $HOSTNAME; aborting cluster bring-up. Fix the failing module, then re-run."
     hex_log_event -e BSP00004E "interface=system,host=$HOSTNAME,service=bootstrap"
 }
 
@@ -68,8 +74,7 @@ PostBootRecovery()
     fi
 
     if hex_sdk cube_cluster_ready ; then
-        msg="Cluster checking and repairing"
-        echo -n "$space_filler">$(hex_sdk GetKernelConsoleDevices) ; echo -n "$msg" $'\r' | tee $BOOTSTRAP_CUBE_LOG $(hex_sdk GetKernelConsoleDevices)
+        bootstrap_status "Cluster checking and repairing"
         hex_log_event -e BSP00005I "interface=system,host=$HOSTNAME,service=bootstrap"
         hex_sdk cmd "rm -f /tmp/health_*error.count"
         # normal boots only: 'cluster check' kicks auto_repair, which on a roll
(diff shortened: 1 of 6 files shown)


Eandalf-Bigstack:
# Code Review: Rolling Update Stall Watchdog

## Overview
This PR implements a stall detection and recovery mechanism for rolling updates via a systemd watchdog timer. The implementation is thoughtfully designed, with good defensive patterns and appropriate scope limitations (VIP-holder-only, control nodes only).

## Critical Issues
**None identified.**

## High-Priority Concerns

### 1. **VIP Lookup May Silently Fail**
```bash
source hex_tuning $SETTINGS_TXT cubesys.control.vip
_vip=$T_cubesys_control_vip
[ -n "$_vip" ] || return 0
```
- If `hex_tuning` fails (missing file, syntax error), `$T_cubesys_control_vip` won't be set and the watchdog silently exits.
- **Impact**: Watchdog becomes inert on nodes where tuning load fails.
- **Recommendation**: Add error handling or log the failure for observability.

### 2. **Heartbeat Injection in `_power_roll_write()` is Brittle**
```bash
jq "$@" --argjson __hb "$(date +%s)" "($filter) | .heartbeat=\$__hb"
```
- The filter re-wrapping assumes the final argument is always a valid jq filter expression. If a caller passes an invalid filter or positional args in unexpected order, the heartbeat injection fails silently or corrupts the filter.
- **Impact**: Misaligned heartbeat timestamps or failed job.json updates.
- **Recommendation**: Add input validation or document the strict contract (final arg must be filter, no positional varargs).

### 3. **Remote VM Count Query Lacks Error Recovery**
```bash
local _nvms=$(remote_run $host "virsh list --uuid --state-running 2>/dev/null | grep -c ." ... | tr -dc 0-9)
```
- If `remote_run` fails or returns non-numeric output, `tr -dc 0-9` leaves an empty string, which bash arithmetic handles as 0. This masks remote failures.
- **Impact**: Nodes with many VMs get the default 2700s timeout instead of scaled timeout if remote_run fails.
- **Recommendation**: Explicitly check `remote_run` success before parsing output; log failures.

## Medium-Priority Issues

### 4. **Bootstrap Status Refactoring: No Validation of Device List**
```bash
KERNEL_CONSOLES=$(hex_sdk GetKernelConsoleDevices)
# later:
echo -n "$space_filler"$'\r'"$msg"$'\r' | tee $KERNEL_CONSOLES >/dev/null 2>&1
```
- If `GetKernelConsoleDevices` returns an empty string or malformed path list, `tee` silently fails (`2>&1 >/dev/null`). Status updates disappear from all consoles without warning.
- **Recommendation**: Validate `KERNEL_CONSOLES` is non-empty at boot; log to syslog if console output fails.

### 5. **Pause-Reason Comma Stripping May Lose Context**
```bash
reason=${1//[,$'\n']/ }
```
- Comma stripping (e.g., "node exceeded deadline (phase: draining, error: timeout)") becomes "node exceeded deadline (phase: draining  error: timeout)". Context is preserved but formatting is mangled.
- **Note**: This is acknowledged as necessary for CSV safety; acceptable trade-off, but document it.

## Low-Priority / Design Notes

### 6. **Heartbeat Only Checked in Draining/Staging Phases**
```bash
case "$phase" in
    draining) quiet=600 ;;
    staging)  quiet=900 ;;
    *)        return 0 ;;
esac
```
- Reboot and boot phases (orchestrator-driven, not write-driven) don't trigger on heartbeat silence—only deadline violations. This is intentional per the PR description but easy to miss during debugging.
- **Recommendation**: Add inline comment explaining why reboot/boot phases are excluded.

### 7. **MariaDB URL Update**
```
https://ftp.icm.edu.pl/... → https://dl.rockylinux.org/...
```
- Reasonable fallback; acceptable as a mirror change. Verify in testing that the package path structure is identical.

## Verification Checklist

- [ ] Verify `hex_tuning` call and VIP variable resolution on test control nodes
- [ ] Confirm `_power_roll_write()` heartbeat injection with real jq filters passed by callers
- [ ] Test watchdog behavior when `remote_run` fails (VM count query) on a compute node
- [ ] Validate bootstrap console device list is non-empty on boot
- [ ] Confirm pause/continue/abort guard symmetry with `ceph_enter_rolling` / `ceph_leave_rolling`
- [ ] Verify MariaDB package availability at new URL
- [ ] Lab validation reproduces the wedge scenarios and reconcile recovery successfully

## Summary
The watchdog design is sound with appropriate scope limits and defensive nullchecks. The three high-priority concerns are edge-case error paths that are unlikely to occur but should be hardened. Bootstrap refactoring is clean and fixes real issues (multi-device redirect, sshd timing).

Eandalf-Bigstack:
This PR introduces a solid, necessary backstop for stalled rolls. Reusing the VIP holder for the watchdog and avoiding an N×N polling matrix by having nodes self-stop their timers is a great design choice.

Building on the previous discussion, here is a detailed review of the diff, highlighting a new critical blocking issue alongside refinements for the existing concerns.

## Critical Issues (Blocking)

### 1. Unbounded SSH Hang in `_power_roll_kick`

In `_power_roll_kick()`, you added a remote call to calculate the timeout budget:

```bash
local _nvms=$(remote_run $host "virsh list --uuid --state-running 2>/dev/null | grep -c ." </dev/null 2>/dev/null | tr -dc 0-9)

```

**The Risk:** If `$host` is wedged at the network or OS level (e.g., blackholed routing, out of memory), `remote_run` (SSH) will hang until the system TCP timeout (often 2+ minutes). Because `_power_roll_kick` is run by the orchestrator, an SSH hang here stalls the orchestrator *before* the new deadline is properly committed, potentially preventing the watchdog from accurately enforcing the stall.
**The Fix:** Wrap the `remote_run` in a strict timeout. Since this is just a quick `virsh` query, 5-10 seconds is plenty:

```bash
local _nvms=$(timeout 10 remote_run $host "virsh list ..." | tr -dc 0-9)

```

### 2. Fragile `jq` Filter Injection

The previous reviewer was right to call out the `jq` injection in `_power_roll_write`:

```bash
local filter="${@: -1}"
set -- "${@:1:$(($#-1))}"
...
jq "$@" --argjson __hb "$(date +%s)" "($filter) | .heartbeat=\$__hb" $ROLLING_JOB > "$t"

```

**The Risk:** This blindly assumes the last argument is *always* the `jq` filter string. If any caller currently or in the future passes trailing arguments (e.g., `_power_roll_write '.phase = $p' --arg p "draining"`), `filter` becomes `"draining"` and the script fails, corrupting the job state.
**The Fix:** Since `_power_roll_write` already wraps the file operations in a tight `flock`, you don't need to surgically inject the heartbeat into the caller's execution. Execute them sequentially inside the lock:

```bash
jq "$@" $ROLLING_JOB > "$t" && jq --argjson __hb "$(date +%s)" '.heartbeat=$__hb' "$t" > "$t.2" && mv "$t.2" $ROLLING_JOB

```

---

## Medium Priority (Observability & Robustness)

### 3. Silent Watchdog Failure on VIP Lookup

In `power_roll_watchdog()`:

```bash
source hex_tuning $SETTINGS_TXT cubesys.control.vip
_vip=$T_cubesys_control_vip
[ -n "$_vip" ] || return 0

```

If `hex_tuning` fails (e.g., file permission issue, syntax error in the config), the watchdog silently exits. A silent failure in a watchdog defeats its purpose.
**The Fix:** Add a log emission before returning so operators know the watchdog is blinded:

```bash
if [ -z "$_vip" ]; then
    log_error "roll_watchdog: failed to resolve VIP from hex_tuning; watchdog disabled"
    return 0
fi

```

### 4. Masked VM Count Failures

The previous reviewer worried that `[ "$_nvms" -gt 0 ]` would crash bash if `$_nvms` was empty. Your guard `[ -n "$_nvms" ]` successfully prevents a crash, but it still fails silently, dropping the node to the 2700s default.
**The Fix:** If the node has dozens of VMs, falling back to 2700s guarantees a false-positive watchdog pause during the drain. Log a warning if `_nvms` fails to resolve so you have an audit trail if a massive node gets paused prematurely.

---

## Design / Code Polish

* **Bootstrap `tee` command:** `tee $KERNEL_CONSOLES >/dev/null 2>&1`. If `$KERNEL_CONSOLES` resolves to an empty string, this evaluates to `tee >/dev/null 2>&1`, which safely absorbs stdin without crashing. This is functionally safe, but explicitly defaulting it to `/dev/null` if empty (e.g., `tee ${KERNEL_CONSOLES:-/dev/null}`) makes the intent clearer.
* **Watchdog self-stopping:** `[ "$_state" != "running" ] && Quiet -n systemctl stop hex-roll-watchdog.timer` works perfectly to clean up N×N nodes without orchestrator intervention.
* **Pause-Reason Comma Stripping:** `${1//[,$'\n']/ }` effectively protects the CSV logging pipeline. Formatting degradation (missing commas in text) is a worthwhile trade-off for log stability here.

Eandalf-Bigstack (approved):

PR #1137, opened 2026-07-22
fix(fixpack): re-make triggers target re-build

leechienpang:
- pick up hex
- make -C core/fixpacks does not trigger re-build without file changes

Fixes #

> Note

#### Additional documentation

```docs

```

### DIFF
--- a/core/fixpacks/Makefile
+++ b/core/fixpacks/Makefile
@@ -44,12 +44,13 @@ fixpack: $(FIXPACK)
 
 secfixdraft:
 	$(call RUN_CMD_SILENT,$(SHELL) $(HEX_SCRIPTSDIR)/makesecfixdraft $@,"  GEN     $@")
+	$(Q)sed -i "s/SUPPORTED_FIRMWARES=.*/SUPPORTED_FIRMWARES=\"$(SUPPORTED_FIRMWARES)\"/" $@/fixpack.info
+	$(Q)sed -i "s/FIXPACK_ID=.*/FIXPACK_ID=\"$(FIXPACK_ID)\"/" $@/fixpack.info
+	$(Q)sed -i "s/FIXPACK_NAME=.*/FIXPACK_NAME=\"$(FIXPACK_ID)\"/" $@/fixpack.info
+	$(call RUN_CMD,$(SHELL) $(HEX_SCRIPTSDIR)/makehotfix -C -p 20M $@ $(FIXPACK_ID).fixpack,"  GEN     $@ $(FIXPACK_ID).fixpack")
+	$(Q)md5sum $(FIXPACK_ID).fixpack >$(FIXPACK_ID).fixpack.md5
 
 secfix: secfixdraft
-	$(Q)sed -i "s/SUPPORTED_FIRMWARES=.*/SUPPORTED_FIRMWARES=\"$(SUPPORTED_FIRMWARES)\"/" $</fixpack.info
-	$(Q)sed -i "s/FIXPACK_ID=.*/FIXPACK_ID=\"$(FIXPACK_ID)\"/" $</fixpack.info
-	$(Q)sed -i "s/FIXPACK_NAME=.*/FIXPACK_NAME=\"$(FIXPACK_ID)\"/" $</fixpack.info
-	$(call RUN_CMD,$(SHELL) $(HEX_SCRIPTSDIR)/makehotfix -C -p 20M $< $(FIXPACK_ID).fixpack,"  GEN     $@ $(FIXPACK_ID).fixpack")
-	$(Q)md5sum $(FIXPACK_ID).fixpack >$(FIXPACK_ID).fixpack.md5
+	$(Q)true
 
 include $(HEX_MAKEDIR)/hex_sdk.mk


Eandalf-Bigstack:
Summary
- Intent looks reasonable (move heavy work into secfixdraft so plain make -C core/fixpacks doesn't force rebuild). Good submodule bump.

Critical issues & suggestions
- Portability: sed -i and md5sum are GNU-specific and will fail on macOS. Use portable alternatives (perl -pi -e for in-place, or detect gnu-sed / gsed; use md5 on macOS) or document Linux-only expectations.
- Make semantics: secfix is now a no-op (true). If secfix is expected to be a file target (not .PHONY), this can break timestamps and incremental builds—confirm .PHONY declarations and that targets still represent produced artifacts.
- Duplication & ordering: moving sed/makehotfix into secfixdraft duplicates logic; consider generating fixpack.info correctly from the generator instead of post-editing.
- Race/timestamp: modifying $@/fixpack.info inside secfixdraft will update timestamps—verify this achieves the desired non-rebuild behavior and doesn't cause spurious downstream rebuilds.
- Submodule bump: ensure consumer CI updates submodules; include changelog if required.

Minor nit
- FIXPACK_NAME set to FIXPACK_ID — intentional? If name differs, use distinct variable.

Actionable fixes
- Replace sed -i with portable in-place edits or guard by OS; replace md5sum with a portable checksum command; verify .PHONY and target timestamps; prefer generator options over sed post-processing.

Eandalf-Bigstack:
Here is a review of the proposed PR changes.

The intent to prevent spurious rebuilds by adjusting the Makefile targets is valid, but the current approach introduces several Make anti-patterns and cross-platform compatibility issues that need to be addressed.

### 🛑 Critical Issues & Architecture Flaws

* **Make Target vs. Output Mismatch (Side-effects):**
The `secfixdraft` target is generating files (`$(FIXPACK_ID).fixpack` and its `.md5` checksum) that do not match the target name. Make relies on file timestamps matching the target name to determine if a rebuild is necessary. Because the target name `secfixdraft` doesn't match the artifact, Make cannot properly track the artifact's state. The actual artifacts should be the targets.
* **Directory as a Target:**
The `$@/fixpack.info` usage implies `secfixdraft` is a directory. Using directories as Make targets is notoriously flaky. A directory's modification time updates whenever a file inside it is added, removed, or modified. Modifying `fixpack.info` via `sed` inside the recipe updates the directory's `mtime`, which will likely trigger the exact spurious downstream rebuilds this PR is trying to prevent.
* **Broken Portability:**
As noted in the prior discussion, `sed -i` and `md5sum` are GNU-specific.
* On macOS/BSD, `sed -i` requires a backup extension (e.g., `sed -i ''`).
* macOS uses `md5`, not `md5sum`.
* *Recommendation:* Use a portable wrapper, `perl -pi -e`, or check the host OS to define these commands.



### ⚠️ Logic & Cleanliness

* **Post-processing vs. Generation:**
Running three sequential `sed` commands to overwrite values in `fixpack.info` immediately after generating it is an anti-pattern.
* *Recommendation:* Update the `makesecfixdraft` script to accept `SUPPORTED_FIRMWARES` and `FIXPACK_ID` as arguments or environment variables so the file is generated correctly the first time.


* **`.PHONY` Declarations:**
The `secfix` target is now effectively a no-op alias (`true`). If this is intended purely as a command alias and doesn't represent a file, it **must** be declared in `.PHONY: secfix`. Otherwise, if a file named `secfix` is ever created, the target will stop running entirely.
* **Variable Name Duplication:**
`FIXPACK_NAME=\"$(FIXPACK_ID)\"` overrides the name with the ID. Ensure this is the desired behavior, as it renders the `FIXPACK_NAME` concept redundant.

### 💡 Suggested Refactor Strategy

Instead of shifting the flawed logic up a level, consider refactoring the targets to reflect the actual artifacts produced:

```makefile
# Define the actual artifacts as targets
FIXPACK_FILE = $(FIXPACK_ID).fixpack

.PHONY: secfix secfixdraft

# Alias targets for developer convenience
secfix: $(FIXPACK_FILE)
secfixdraft: secfixdraft/fixpack.info

# Target for the draft directory contents
secfixdraft/fixpack.info:
	$(call RUN_CMD_SILENT,$(SHELL) $(HEX_SCRIPTSDIR)/makesecfixdraft $(@D),"  GEN     $(@D)")
	# (Ideally, pass the variables directly to makesecfixdraft instead of using sed here)

# Target for the final fixpack depends strictly on its inputs
$(FIXPACK_FILE): secfixdraft/fixpack.info
	$(call RUN_CMD,$(SHELL) $(HEX_SCRIPTSDIR)/makehotfix -C -p 20M $(<D) $@,"  GEN     $@")
	$(Q)$(MD5_CMD) $@ >$@.md5

```

*(Note: You will need to define a portable `$(MD5_CMD)` elsewhere in your Makefiles).*

Eandalf-Bigstack:
Although the Make anti-pattern issue does exist, we would let this PR go through to fix pipeline errors. We could handle these Make anti-patterns later.

Eandalf-Bigstack (approved):

PR #1138, opened 2026-07-22
fix(nginx): SPA cache contract for the cube-cos-ui vhost

### LINKED ISSUE #1134: [Bug] UI index.html served without Cache-Control: browsers pin a stale bundle after upgrades
## Description

The cube-cos-ui nginx vhost (`core/nginx/nginx.conf.in`, the `:8083` server behind haproxy) serves everything — including `index.html` — with **no `Cache-Control` header** (only `Last-Modified` + `ETag`). With no explicit policy, browsers apply *heuristic* caching (~10% of the file's age since Last-Modified, i.e. hours to days), so after any upgrade or UI hot-deploy a browser can keep serving the **old `index.html` → old hashed JS bundle** without ever revalidating — even in brand-new tabs. Users end up on a stale UI against a new API until the heuristic expires or they hard-reload.

Observed on the sky lab (2026-07-22): `curl -skD- https://10.32.10.140/` shows `last-modified` + `etag` but no `cache-control`; current entry bundle `index-CS-U1iMP.js` is referenced only by an `index.html` browsers are allowed to cache heuristically.

This is the standard SPA cache contract, currently inverted:
- `index.html` (mutable entry point) → must always revalidate: `Cache-Control: no-cache`
- `/assets/*` (content-hashed, immutable) → should be cached forever: `public, max-age=31536000, immutable` (today they get heuristic caching too, so they're *under*-cached)

## How to Reproduce

1. Load the UI, leave the tab (or just let the browser cache `index.html`).
2. Upgrade the cluster / hot-deploy a new `dist/` into `/usr/share/cube/ui/`.
3. Open the UI again within the heuristic-freshness window: the old bundle loads (no request for `index.html` hits nginx). New features/fixes are invisible; API/UI skew can surface as broken pages.

## Expected behavior

Every page load revalidates `index.html` (cheap 304 when unchanged), so a new deploy is picked up on the next navigation; hashed assets are cached immutably.

## Screenshots (Optional)

None — header capture in the description.

Found while debugging the /events keyword-search report (the search bug itself was server-side, fixed in cube-cos-api#611 — but this caching gap will bite any future UI change). A proposed nginx patch is attached in the comments; needs a cube-cos-ui domain owner to confirm the asset-path assumptions (`/assets/` hashed-only, anything else the SPA emits) before landing.

Any browser; observed with Chrome against sky (10.32.10.140).

- [x] **Handbook knowledge update** — landed in [bigstack-handbook#111](https://github.com/bigstack-oss/bigstack-handbook/pull/111) (SPA cache policy for the UI vhost).

traviswu-bigstack:
#### What type of PR is this?

/kind bug

The UI vhost serves everything with no `Cache-Control`, so browsers apply heuristic caching (~10% of the file's age since `Last-Modified`). On a long-lived production cluster that means **days to weeks** of a stale `index.html` → stale hashed bundle against a new API after every upgrade, in brand-new tabs, with no request ever reaching nginx — the classic inverted SPA cache contract.

Fix (simplified from the originally proposed per-location patch): a **server-level `Cache-Control: no-cache` default** makes every mutable path revalidate per load — `index.html`, the `try_files` SPA fallback for deep links, and the dist-root files the per-location version missed (`favicon.svg`, `mockServiceWorker.js`, `LICENSE`) — and a single `location /assets/` override gives the content-hashed bundles `public, max-age=31536000, immutable` (25 MB / 872 files load once per browser instead of per heuristic window; matters most over VPN/WAN/edge links). `add_header` inheritance does the work: locations without their own header inherit the server default. Side benefit: missing assets now 404 instead of falling back to `index.html`.

Validated live on all three sky nodes (config hot-applied + `nginx -t` + reload):
`/`, deep link `/events`, `favicon.svg` → `no-cache`; `assets/index-CS-U1iMP.js` → `immutable`; haproxy `OPTIONS` health check unaffected (200).

**Known one-release lag**: browsers that already cached `index.html` heuristically never ask nginx, so the upgrade *delivering* this fix still shows one final stale window — only upgrades after it are protected. Suggest a release-note line: hard-reload the UI once after upgrading to the release containing this.

Fixes #1134

- `location = /index.html` from the original proposal became unnecessary — the server-level default covers it (including the internal-redirect fallback path), with fewer blocks and wider coverage.
- `add_header` applies to success codes only by default; error pages and the health-check `return 200` behave as before.
- The template renders via hex_config's nginx module; a config commit or reboot re-renders `/etc/nginx/nginx.conf`.

#### Additional documentation

```docs
Issue #1134 DoD calls for a handbook known-issue entry (SPA cache policy);
pending as a follow-up handbook PR.
```

### DIFF
--- a/core/nginx/nginx.conf.in
+++ b/core/nginx/nginx.conf.in
@@ -291,6 +291,15 @@ http {
         access_log /var/log/cube-cos-ui/cube-cos-ui-nginx-access.log ui;
         error_log /var/log/cube-cos-ui/cube-cos-ui-nginx-error.log;
 
+        # SPA cache contract: everything mutable under its URL (index.html,
+        # favicon, dist-root files) revalidates every load; hashed bundles
+        # never change under their name and cache forever
+        add_header Cache-Control "no-cache";
+
+        location /assets/ {
+            add_header Cache-Control "public, max-age=31536000, immutable";
+        }
+
         location / {
             # health check endpoint for HAProxy
             if ($request_method = OPTIONS) {


Eandalf-Bigstack:
# Code Review: Fix(nginx): SPA cache contract for cube-cos-ui vhost

## Summary
This PR implements a standard SPA cache pattern by adding a server-level `Cache-Control: no-cache` default and a `/assets/` override for immutable hashed bundles. The approach is sound and addresses the documented problem.

---

## ✅ Strengths

1. **Correct caching semantics**: `no-cache` (revalidate on every request) is the right choice for mutable SPA files; `public, max-age=31536000, immutable` for hashed bundles is a best practice.

2. **Minimal, focused change**: Leveraging `add_header` inheritance rather than specifying cache headers in every location reduces maintenance surface.

3. **Good side effect**: Eliminating fallback to `index.html` for missing assets (404 instead) prevents silent failures and improves observability.

4. **Validated approach**: Already tested live on production nodes with verification of header behavior and health check compatibility.

---

## ⚠️ Observations & Questions

1. **Header inheritance verification**: Confirm that the location `/assets/` override properly replaces (not appends to) the server-level header in your nginx version. Standard behavior is replacement, but this should be verified in your test environment.

2. **Error pages**: If custom error pages (4xx/5xx from `error_page` directives) exist downstream in the config, they will NOT inherit the server-level `Cache-Control` header unless explicitly configured. This is likely intentional (error pages often shouldn't be cached), but worth documenting.

3. **Other mutable files coverage**: The PR notes this now covers `favicon.svg`, `mockServiceWorker.js`, `LICENSE` via the server default. Confirm these files don't have stale issues and that their hashed versions (if any exist) are correctly placed under `/assets/`.

4. **Release note**: The PR correctly calls out that browsers with already-cached `index.html` won't request a revalidation until their heuristic window expires. Ensure the suggested "hard-reload after upgrade" guidance is included in release notes.

---

## No Critical Issues Found

The implementation follows nginx best practices for SPA caching. The change is safe, backwards-compatible for browsers, and directly addresses the root cause of stale deployments.

Eandalf-Bigstack:
## Code Review: SPA Cache Contract for Nginx

While the caching strategy (`no-cache` for mutable files, immutable caching for hashed bundles) is the correct standard for SPAs, there is a critical Nginx configuration quirk that puts the primary goal of this PR at risk.

### 🔴 Critical Issue: The Nginx `add_header` Inheritance Trap

You placed `add_header Cache-Control "no-cache";` at the server level, relying on inheritance to apply it to `location /` (which serves `index.html`).

In Nginx, **`add_header` directives do not inherit if the child block contains *any* `add_header` directives of its own.**

If your `location /` block (or any other location blocks not fully visible in the diff) contains *any* security or CORS headers (e.g., `add_header X-Frame-Options SAMEORIGIN;` or `add_header Strict-Transport-Security ...`), the server-level `Cache-Control` header will be **silently ignored** for requests hitting that location. This would result in `index.html` remaining without a `Cache-Control` header, entirely defeating the purpose of this PR.

**Recommendation:**
Unless you can guarantee that `location /` will *never* contain an `add_header` directive in the future, it is safer to explicitly declare the `no-cache` header inside the `location /` block:

```nginx
        location / {
            add_header Cache-Control "no-cache";
            # ... existing directives (try_files, etc.)
            
            # health check endpoint for HAProxy
            if ($request_method = OPTIONS) {

```

### 🟡 Architectural Note: Missing Assets 404 Behavior

The PR notes: *"Side benefit: missing assets now 404 instead of falling back to `index.html`."*

This occurs because a request to `/assets/main.js` now perfectly matches `location /assets/`. Because this new location block lacks a `try_files` directive, Nginx defaults to standard file serving and returns a 404 if the file isn't on disk.

While this is the desired behavior for a bundler's static assets (preventing the browser from silently parsing HTML as JavaScript when an asset is missing), ensure that your deployment process doesn't rely on `try_files` routing for anything inside the `/assets/` directory. If any virtual routing needs to happen under `/assets/`, you will need to add a specific `try_files` directive inside the `location /assets/` block.

Eandalf-Bigstack (approved):

PR #1094, opened 2026-07-08
fix(health): repair OVN metadata by nb_cfg progress, not the Alive flag

traviswu-bigstack:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it

After a rolling restart, `neutron(N metadata not all up)` could stay NG for ~10-15 min, with the auto-repair *prolonging* it.

**Root cause.** OVN metadata "Alive" is an nb_cfg sequence: neutron-server bumps `NB_Global.nb_cfg`, ovn-northd copies it to SB, each agent writes the value it processed into its `Chassis_Private`. `Alive=False` only means the agent's `sb-cfg` **lags** `nb_cfg` — normal while it re-syncs after reconnecting. `health_neutron_check` flagged `ERR_CODE=3` on *any* `Alive=False` agent and the repair restarted it — but restarting a catching-up agent **drops its SB connection and resets the sync**, so the repair (every cycle) kept throwing away the agent's progress. Observed: 3 restarts over ~15 min, each interrupting re-sync; healed only once a restart was left alone.

**Fix — classify by nb_cfg progress:**
- `caught_up` — `sb-cfg >= nb_cfg`
- `catching_up` — `sb-cfg` advancing since last check (or first-seen)
- `stuck` — `sb-cfg` frozen

Then split by caller, because the two repair paths have different contracts:

- **`auto_repair`** (periodic, backgrounded, must not block): flag + restart **only stuck** agents; leave catching-up ones entirely — they self-heal within a check cycle.
- **`repair()` = `health_neutron_repair`** (explicit: operator `check_repair`, `set_ready`): its contract is to return a *converged* service, so it now **waits (bounded 120 s)** for catching-up agents before escalating. Previously any `Alive=False` triggered a full `hex_config bootstrap neutron` (~75 s) and **deleted the agent record** — resetting the very sync that was about to complete. The wait is self-limiting (only while a read is advancing) and capped by an absolute wall-clock deadline.

**Scale.** Metadata agents run on **every compute** (target: 100+ nodes), so the OVN reads are **bulk**: one `nb_cfg` read plus a single `Chassis`/`Chassis_Private` snapshot (joined on chassis name) classify *every* agent at once — a fixed **~3 ovn calls per pass regardless of node count**, not 3 per host. A timed-out/unreadable read → `stuck` (fail-safe), so a hung OVN never holds the wait; the per-pass cost is bounded even at 100+ nodes, in both `check()` and `repair()`.

[...]

PR #1140, opened 2026-07-22
fix(cluster): import builtin glance images at set_ready, not at boot

### LINKED ISSUE #1139: [Bug] Built-in Cirros/Alpine glance images never import on multi-node installs (silent run-once, wedged retries)
## Description

On multi-node e2e installs, the built-in Cirros/Alpine glance images are never imported, so anything that boots a VM fails. Accept #691 (the first e2e run of the refactored `cube_cluster_start` split): `3cc/test_23_cc3_server` failed with **"No Image found for Cirros"** and `2ec_1m/test_32_ec1_diagnostics-dns` failed inside `hex_sdk diagnostics_network` (its `_diagSrvCreate Cirros` probe VM). The identical diagnostics test passed on single-node `1ec` (50s), and sky (real hardware) has the image — the failure is multi-node + loaded-QEMU timing.

The import is a run-once-and-silent path with three compounding defects:

1. `cube_cluster_start_cluster()` calls `os_image_import` with the failure swallowed — a first attempt that fails (glance/cinder not fully servable when the master's cluster-wide half fires) leaves no evidence and nothing re-drives it explicitly.
2. The old "delete `Cirros queued` records" cleanup was removed in the cluster-start refactor — a killed first attempt leaves a queued/saving record that wedges the image name for retries.
3. `os_image_import`'s `/run/os_image_import_*` dedup marker is removed only on the success and quota-error paths — a killed import leaks it, and every later attempt on any node aborts with "Image is being imported" (also silent under the caller).

Since set_ready re-invokes `cube_cluster_start` (and the gate re-runs the cluster-wide half on the master), the retry OPPORTUNITY exists — defects 2 and 3 are what stop the retry from succeeding, and defect 1 is why nobody can see any of it.

## How to Reproduce

1. Run the accept e2e on a loaded host (or any multi-node install where the master's first `cluster_start` fires while glance/cinder are still settling).
2. After set_ready + cluster-check green: `openstack image list` → no `Cirros`.
3. `openstack server create --image Cirros …` → "No Image found for Cirros"; `hex_sdk diagnostics_network public` → rc 1.

## Expected behavior

The built-in images import reliably: a failed first attempt is loudly logged, leaves no wedged queued record or stale marker, and the set_ready re-run of the cluster-wide half completes the import.

## Screenshots (Optional)

None — accept #691 console (`test_23_cc3_server`, `test_32_ec1_diagnostics-dns` transcripts).

Follow-up worth considering separately: `cube_cluster_start_cluster_if_master()` waits up to 30 min for peers' CLUSTER_SYNC and then **runs the cluster-wide half anyway** — on a first-time multi-node install where peers aren't installed yet (e2e installs nodes serially), the half runs against a half-formed cluster. The import retry makes this benign for images, but the run-anyway semantics deserve a look.

Third #691 failure for the record: `3c_3p_5s/test_33_c1_network` was an unrelated transient (neutron API connection drop during router ops — the known oslo_db/eventlet worker-crash window).

N/A — QEMU e2e rig (all multi-node topologies).

- [x] **Handbook knowledge update** — landed in [bigstack-handbook#160](https://github.com/bigstack-oss/bigstack-handbook/pull/160) (known-issue: builtin glance images missing on multi-node installs — run-once/silent import + marker-leak pattern, workaround, and the #1140 fix).

traviswu-bigstack:
#### What type of PR is this?

/kind bug

**Moves the built-in glance image import (Cirros/Alpine) to `cluster set_ready` — the one point where the cluster is guaranteed complete — and removes it from the boot path entirely.**

Accept #691 — the first e2e run of the cluster-start refactor — lost the built-in **Cirros** image on both multi-node topologies: `3cc/test_23_cc3_server` ("No Image found for Cirros") and `2ec_1m/test_32_ec1_diagnostics-dns` (its `_diagSrvCreate Cirros` probe VM). Single-node `1ec` passed the identical diagnostics test, and sky (real hardware, near-simultaneous FTS) has the image.

**Failure mechanism** (multi-node serial install): the master's boot-time cluster half runs the import against a still-forming cluster. glance-api is up locally, so it *accepts the image record* (`queued`) — then the upload **blocks forever on the quorumless ceph backend** (ceph clients hang on no-quorum by design). The import marker stays held by the wedged process, so the set_ready retry aborts on "Image is being imported" — technically correctly, for an import that will never finish. All of it silent under `Quiet`.

**Why it never happened before the refactor**: the import lived inline in `cube_cluster_start()`, which ran on **every node's boot** and again on **every node at set_ready** — ≥4 attempts per install, each against a healthier cluster, with the queued-record cleanup healing debris in between. Robust by accidental brute force. The refactor collapsed that to a single doomed boot attempt plus one sabotage-able retry.

**The fix** stops patching around the doomed attempt and removes it:

1. New `cube_import_builtin_images()`, a numbered set_ready step (`[4/6] Importing builtin images`); the block is deleted from `cube_cluster_start_cluster()`, so boot never attempts an import. The function is deliberately lean — its only caller is set_ready on a complete cluster, so the earlier defensive layers (marker clearing, logger fallbacks) are gone; the one heal it keeps is deleting queued/saving records (restored old-code behavior: an interrupted set_ready leaves a `queued` record that would make a re-run create a duplicate name). After importing, the function **verifies** `Cirros` is `active` and returns an explicit "built-in image Cirros could not be imported" error otherwise; the set_ready step prints a `Warning: … re-run 'cluster set_ready' to retry.` on failure instead of proceeding silently.
2. set_ready's step list is renumbered with the two disabled no-op blocks removed (`[2/6] Checking SDN services` and the commented-out `[6/6]` neutron repair, both #1094 leftovers).
3. Defense in depth: `os_image_import` itself now sweeps import markers older than 2 h before its duplicate-import check (previously only the success/quota paths removed the marker).

**Attribution**: the regression trigger is the boot-time-initiative cluster-start split; the marker-hold behavior in `os_image_import` and ceph's block-on-no-quorum are pre-existing — the refactor removed the accidental retries that had been hiding them. Same latent-bug-exposed-by-refactor shape as the influxdb/ceph `CONFIG_REQUIRES` race (#1132). For the record, the third #691 failure (`3c_3p_5s/test_33_c1_network`) is unrelated: a transient neutron API connection drop (known pre-existing oslo_db/eventlet worker-crash window).

Fixes #1139

> `cli_cluster.o` compiles; `bash -n` clean. All topologies run set_ready (e2e: `test_13_cc1_setready`, `test_29_*_setready*`), so no install path loses the import. Upgrades keep their already-imported images (glance persists). The `if_master` gate's wait-30-min-then-run-anyway semantics remain flagged in #1139 as a follow-up — with the import gone from boot, its blast radius is much smaller. Real verification = the next accept run: `test_23_cc3_server` and `test_32_ec1_diagnostics-dns` are the regression tests.

#### Additional documentation

```docs
Handbook: candidate known-issue on the wedged-import pattern (glance upload
blocks on quorumless ceph; marker holds; retry self-blocks) — see #1139 DoD.
```

### DIFF
--- a/core/sdk_sh/modules/sdk_os.sh
+++ b/core/sdk_sh/modules/sdk_os.sh
@@ -775,8 +775,11 @@ os_image_import()
     local _distro=${7:-linux}
     local properties=${8:---property hw_disk_bus=scsi --property hw_scsi_model=virtio-scsi --property hw_machine_type=q35 --property hw_video_model=vga}
 
-    # prevent duplicated importing
+    # prevent duplicated importing; a marker older than 2h is a leak from a
+    # killed import (only the success/quota paths remove it) -- sweep the leak
+    # so a retry of this same import is not blocked forever
     local mf_importing="/run/${FUNCNAME[0]}_${file}_${name}_${pool}"
+    cmd find $mf_importing -mmin +120 -delete >/dev/null 2>&1
     if  cmd -v ls $mf_importing | grep -q "|0|" ; then
         Error "Image is being imported. If this is not the case, remove $mf_importing on all nodes"
     else
--- a/core/main/proj_functions
+++ b/core/main/proj_functions
@@ -323,19 +323,6 @@ cube_cluster_start_cluster()
     # ready yet, leaving single-node flooding STONITH errors + unbounded pe-error.
     Quiet -n pcs property set stonith-enabled=false pe-error-series-max="1000" pe-warn-series-max="1000" cluster-recheck-interval="5min"
 
-    if $OPENSTACK image list -f value | grep -q "Cirros active" ; then
-        cmd rm -f /etc/glance/cirros-0.4.0-x86_64-disk.qcow2
-        cmd rm -f /etc/glance/alpine-tools.qcow2
-    else
-        [ ! -e /etc/glance/cirros-0.4.0-x86_64-disk.qcow2 ] || $HEX_SDK os_image_import /etc/glance cirros-0.4.0-x86_64-disk.qcow2 Cirros
-        [ ! -e /etc/glance/alpine-tools.qcow2 ] || $HEX_SDK os_image_import /etc/glance alpine-tools.qcow2 Alpine
-    fi
-    local cnt=0
-    for i in $($OPENSTACK image list -f value | grep "Cirros active" | cut -d " " -f1) ; do
-        ((cnt++))
-        [ $cnt -le 1 ] || $OPENSTACK image delete $i
-    done
-
     cmd "$HEX_SDK migrate_fixpack"
     nohup bash -c "$HEX_SDK git_init" >/dev/null 2>&1 &
 
@@ -374,6 +361,29 @@ cube_cluster_start_cluster_if_master()
         sleep 20
     done
     cube_cluster_start_cluster
+}
+
+# import the builtin glance images (Cirros, Alpine); set_ready-only -- never at
+# boot, where a still-forming ceph wedges the upload (#1139). Idempotent.
+cube_import_builtin_images()
+{
+    if $OPENSTACK image list -f value | grep -q "Cirros active" ; then
+        cmd rm -f /etc/glance/cirros-0.4.0-x86_64-disk.qcow2
+        cmd rm -f /etc/glance/alpine-tools.qcow2
+        return 0
+    fi
+    # heal an interrupted earlier attempt: a queued/saving record wedges the name
+    for i in $($OPENSTACK image list -f value | grep -E "(Cirros|Alpine) (queued|saving)" | cut -d" " -f1) ; do
+        $OPENSTACK image delete $i
+    done
+    [ ! -e /etc/glance/cirros-0.4.0-x86_64-disk.qcow2 ] || $HEX_SDK os_image_import /etc/glance cirros-0.4.0-x86_64-disk.qcow2 Cirros
+    [ ! -e /etc/glance/alpine-tools.qcow2 ] || $HEX_SDK os_image_import /etc/glance alpine-tools.qcow2 Alpine
+
+    # verify the import actually landed
+    if ! $OPENSTACK image list -f value | grep -q "Cirros active" ; then
+        echo "Error: built-in image Cirros could not be imported" >&2
+        return 1
+    fi
 }
 
 cube_cluster_stop()
--- a/core/modules/cli_cluster.cpp
+++ b/core/modules/cli_cluster.cpp
@@ -857,28 +857,26 @@ ClusterReadyMain(int argc, const char** argv)
     CliPrintf("[1/6] Updating storage replication rule");
     HexUtilSystemF(0, 0, HEX_SDK " host_local_run " HEX_SDK " ceph_pool_replicate_init");
 
-    CliPrintf("[2/6] Checking SDN services");
-    // Disabled to verify it's still needed (#1094): health_neutron_repair now waits for OVN
-    // metadata nb_cfg catch-up, and [6/6] cluster_check_repair_async repairs neutron anyway.
-    // HexUtilSystemF(0, 0,  HEX_SDK " host_local_run " HEX_SDK " health_neutron_repair");
-
-    CliPrintf("[3/6] Configuring modules");
+    CliPrintf("[2/6] Configuring modules");
     if (cfgFlatExt)
         HexSpawn(0, HEX_CFG, "-p", "trigger", "cluster_ready", cidr.c_str(), gw.c_str(), iplist.c_str(), ZEROCHAR_PTR);
     else
         HexSpawn(0, HEX_CFG, "-p", "trigger", "cluster_ready", ZEROCHAR_PTR);
 
-    CliPrintf("[4/6] Starting cluster");
+    CliPrintf("[3/6] Starting cluster");
     HexUtilSystemF(0, 0, HEX_SDK " host_local_run hex_sdk cmd " HEX_SDK " cube_cluster_start");
+
+    // the cluster is guaranteed complete here -- never import at boot, where a
+    // still-forming ceph wedges the upload (#1139)
+    CliPrintf("[4/6] Importing builtin images");
+    if (HexUtilSystemF(0, 0, HEX_SDK " cube_import_builtin_images") != 0)
+        CliPrintf("      Warning: built-in image import failed; re-run 'cluster set_ready' to retry.");
 
     CliPrintf("[5/6] Strengthening password");
     HexUtilSystemF(0, 0, HEX_SDK " host_local_run hex_sdk cmd " HEX_CFG " cube_password_init");
 
     CliPrintf("[6/6] Cluster check and repair started in the background.");
     CliPrintf("      The box is ready to use now; the result will pop up here when it completes.");
-    // Disabled to verify it's still needed (#1094): cluster_check_repair_async below already
-    // repairs neutron via check_repair, which now waits for OVN metadata nb_cfg catch-up.
-    // HexSpawn(0, HEX_SDK, "host_local_run", "hex_sdk", "-m force",  "health_neutron_repair", ZEROCHAR_PTR);
     HexSpawn(0, HEX_SDK, "cmd", "rm -f /tmp/health_*_error.count", ZEROCHAR_PTR);
     HexSystemF(0, HEX_SDK " cluster_check_repair_async");
 


Eandalf-Bigstack:
# Code Review: fix(cluster): import builtin glance images at set_ready

## Summary
The PR moves built-in Glance image imports from cluster boot (where a still-forming Ceph backend blocks indefinitely) to the guaranteed-complete `set_ready` phase. This addresses regression #1139 where multi-node topologies lost the Cirros image. The approach is sound, but there is **one critical bug** in the implementation.

---

## Critical Issues

### 1. **Shell Quote Escaping Bug in `os_image_import`** ❌
**File:** `core/sdk_sh/modules/sdk_os.sh` (line added)

```bash
cmd find /run -maxdepth 1 -name "'${FUNCNAME[0]}_*'" -mmin +120 -delete >/dev/null 2>&1
```

**Problem:** The inner single quotes prevent variable expansion. The `find` command will search for a literal filename `'${FUNCNAME[0]}_*'` instead of expanding to `'os_image_import_*'`. This defeats the entire purpose of cleaning up stale marker files.

**Impact:** Stale import markers (>2h old) from interrupted imports won't be cleaned up, causing `set_ready` retries to fail with "Image is being imported" indefinitely.

**Fix:** Remove the inner single quotes:
```bash
cmd find /run -maxdepth 1 -name "${FUNCNAME[0]}_*" -mmin +120 -delete >/dev/null 2>&1
```

---

## Medium Issues

### 2. **Incomplete Import Verification** ⚠️
**File:** `core/main/proj_functions` - `cube_import_builtin_images()`

```bash
if ! $OPENSTACK image list -f value | grep -q "Cirros active" ; then
    echo "Error: built-in image Cirros could not be imported" >&2
    return 1
fi
```

**Problem:** The function only verifies Cirros is active but silently succeeds even if Alpine fails. The old code didn't delete duplicate Cirros either; the new version skips that check entirely.

**Recommendation:** Either verify both images or document why Alpine is optional. Consider whether duplicate cleanup is still needed for safety.

---

## Low Issues / Observations

### 3. **Loose Coupling Between Success and Failure Handling**
**File:** `core/modules/cli_cluster.cpp`

```cpp
if (HexUtilSystemF(0, 0, HEX_SDK " cube_import_builtin_images") != 0)
    CliPrintf("      Warning: built-in image import failed; re-run 'cluster set_ready' to retry.");
```

The C++ code allows set_ready to continue on import failure and just prints a warning. This matches the PR's intent (avoid silent failure), but verify that downstream code doesn't assume Cirros is present.

---

## Positive Aspects ✅
- Root cause analysis is thorough and well-documented
- Moving import to set_ready (complete cluster) is the correct fix
- Queued/saving record cleanup prevents blocking on duplicate names  
- Return code propagation allows graceful degradation
- Idempotency is preserved

---

## Recommendation
**Fix the critical quote-escaping bug before merge.** The other issues are design choices that appear intentional; confirm with author if verification should include Alpine or if the simpler Cirros-only check is sufficient.

Eandalf-Bigstack:
The architectural shift to move the builtin image import to the `set_ready` phase is fundamentally sound. Relying on accidental retries at boot was fragile, and placing this operation where the Ceph backend is guaranteed to have quorum resolves the blocking issue cleanly.

However, there are two significant logic flaws in the implementation that will break idempotency and stale marker cleanup.

## Critical Issues

### 1. `find` Command Quote Bug (Stale Marker Leak)

**File:** `core/sdk_sh/modules/sdk_os.sh`

The previous reviewer (Eandalf-Bigstack) correctly identified that the `find` command is broken, but the exact mechanism is slightly different:

```bash
cmd find /run -maxdepth 1 -name "'${FUNCNAME[0]}_*'" -mmin +120 -delete >/dev/null 2>&1

```

Because the single quotes are inside the double quotes, bash *does* expand `${FUNCNAME[0]}`. However, it then passes the literal single quotes to `find`. As a result, `find` searches for a file literally named `'os_image_import_*'` (including the quote characters).

**Impact:** Stale import markers will not be deleted, re-introducing the "Image is being imported" wedge if a previous run is interrupted.

**Fix:** Remove the inner single quotes. Bash will not glob expand the asterisk as long as it is within double quotes, so `find` will receive the correct pattern.

```bash
cmd find /run -maxdepth 1 -name "${FUNCNAME[0]}_*" -mmin +120 -delete >/dev/null 2>&1

```

### 2. Alpine Image Idempotency and Validation

**File:** `core/main/proj_functions`

The new `cube_import_builtin_images` function introduces a regression in how the Alpine image is handled by binding its fate entirely to the Cirros image check.

**Early Exit Flaw:**

```bash
if $OPENSTACK image list -f value | grep -q "Cirros active" ; then
    cmd rm -f /etc/glance/cirros-0.4.0-x86_64-disk.qcow2
    cmd rm -f /etc/glance/alpine-tools.qcow2
    return 0
fi

```

If a previous `set_ready` run successfully imported Cirros but failed or was interrupted before importing Alpine, subsequent runs will see `Cirros active`, exit immediately, and **never** import Alpine.

**Missing Verification:**
Similarly, the verification block at the end of the function only checks if Cirros is active. If the `os_image_import` for Alpine fails, the function still returns `0` and silently proceeds.

**Fix:** Ensure both images are validated for both the early exit and the final success check.

```bash
# Early exit requires both to be active
if $OPENSTACK image list -f value | grep -q "Cirros active" && \
   $OPENSTACK image list -f value | grep -q "Alpine active" ; then
    cmd rm -f /etc/glance/cirros-0.4.0-x86_64-disk.qcow2
    cmd rm -f /etc/glance/alpine-tools.qcow2
    return 0
fi

# ... import logic ...

# Verify both landed
if ! $OPENSTACK image list -f value | grep -q "Cirros active" || \
   ! $OPENSTACK image list -f value | grep -q "Alpine active" ; then
    echo "Error: built-in images (Cirros or Alpine) could not be imported" >&2
    return 1
fi

```

## Minor Notes

* **C++ Warning Logic (`cli_cluster.cpp`):** The C++ wrapper prints a warning but allows `set_ready` to complete successfully if the import fails. Given the PR description says `set_ready` should explicitly instruct the user to retry, ensure that downstream cluster operations aren't relying on a hard dependency on these images being present immediately after a successful `set_ready` exit code. If they are required, this should be a hard failure (`return -1` or similar).

core/sdk_sh/modules/sdk_os.sh
```diff
@@ -775,8 +775,11 @@ os_image_import()
     local _distro=${7:-linux}
     local properties=${8:---property hw_disk_bus=scsi --property hw_scsi_model=virtio-scsi --property hw_machine_type=q35 --property hw_video_model=vga}
 
-    # prevent duplicated importing
+    # prevent duplicated importing; a marker older than 2h is a leak from a
+    # killed import (only the success/quota paths remove it) -- sweep leaks so
+    # the set_ready/cluster_start retry is not blocked forever
     local mf_importing="/run/${FUNCNAME[0]}_${file}_${name}_${pool}"
+    cmd find /run -maxdepth 1 -name "'${FUNCNAME[0]}_*'" -mmin +120 -delete >/dev/null 2>&1
```
Eandalf-Bigstack: ### 1. **Shell Quote Escaping Bug in `os_image_import`** ❌
**File:** `core/sdk_sh/modules/sdk_os.sh` (line added)

```bash
cmd find /run -maxdepth 1 -name "'${FUNCNAME[0]}_*'" -mmin +120 -delete >/dev/null 2>&1
```

**Problem:** The inner single quotes prevent variable expansion. The `find` command will search for a literal filename `'${FUNCNAME[0]}_*'` instead of expanding to `'os_image_import_*'`. This defeats the entire purpose of cleaning up stale marker files.

**Impact:** Stale import markers (>2h old) from interrupted imports won't be cleaned up, causing `set_ready` retries to fail with "Image is being imported" indefinitely.

**Fix:** Remove the inner single quotes:
```bash
cmd find /run -maxdepth 1 -name "${FUNCNAME[0]}_*" -mmin +120 -delete >/dev/null 2>&1
```

traviswu-bigstack: Good catch on that line being too clever — resolved in the latest push, though the mechanism was subtly different from described. The variable actually **does** expand: the outer quotes are double, so `"'${FUNCNAME[0]}_*'"` becomes `'os_image_import_*'` — the inner single quotes are literal characters that ride through `cmd()`'s ssh string-join and get stripped by the *remote* shell, delivering the unglobbed pattern to `find`. (Verified: `arg = 'os_image_import_*'` locally → remote parses to `os_image_import_*`.) The suggested unquoted form would work too but leaves the glob exposed to the remote shell's cwd expansion.

That said, a line needing this much explanation is a bug of a different kind. Replaced with a glob-free sweep of exactly the marker this import would block on:

```bash
cmd find $mf_importing -mmin +120 -delete >/dev/null 2>&1
```

No pattern, no quoting subtleties, same self-heal. Thanks for the review!

Eandalf-Bigstack (approved):

PR #894, opened 2026-06-03
860 list gpu resources infra implementation

steven-chiu-bigstack:
#### What type of PR is this?

Feat

#### What this PR does / why we need it

- Add `gpu_device_list` function
- Add `gpu_vgpu_profile_list` function

#### Which issue(s) this PR fixes

Close #860 

#### Special notes for your reviewer

- Try it out on 150.1 (with dedicated shell script)
  - `sh ~/list-gpu-resources-860.sh gpu_device_list`
  - `sh ~/list-gpu-resources-860.sh gpu_vgpu_profile_list -gpuId GPU-28eca022-0c66-97c1-a527-17e900626924`
- The creation, update, and migration of the config file will be handled in #893
- Config file path: `/etc/cube/cos/gpu/config.json`
  - Data structure (an array of GPU):
    ```ts
    {
      id: string
      type: 'unset' | 'pgpu' | 'sriovVgpu' | 'migBackedVgpu'
      profiles: {
        id: number
        name: string
        count: number
        alias: string
      }[] | null
    }[]
    ```
  - `profiles` will be `null` for pgpu.

#### Additional documentation

- Mock `/etc/cube/cos/gpu/config.json` on 150.1:
  ```json
  [
    {
      "id": "GPU-28eca022-0c66-97c1-a527-17e900626924",
      "type": "sriovVpu",
      "profiles": [
        {
          "id": 100,
          "name": "NVIDIA H100 1-10C",
          "count": 6,
          "alias": "nvidia_h100_1_10c_100"
        },
        {
          "id": 101,
          "name": "NVIDIA H100 1-10CME",
          "count": 0,
          "alias": "nvidia_h100_1_10cme_101"
        },
        {
          "id": 102,
          "name": "NVIDIA H100 2-20C",
          "count": 9,
          "alias": "nvidia_h100_2_20c_102"
        }
      ]
    }
  ]
  ```
- `gpu_device_list`
  - Usage: `hex_sdk gpu_device_list`
  - Output (beautified, with mock config):
    ```json
    [
      {
        "id":"GPU-28eca022-0c66-97c1-a527-17e900626924",
        "name":"NVIDIA RTX A2000",
        "type":"sriovVpu",
        "pciAddress":"00000000:86:00.0",
        "status":"idle",
        "allocation":{
          "current":0,
          "total":15
        }
      }
    ]
    ```
- `gpu_vgpu_profile_list`
  - Usage: `hex_sdk gpu_vgpu_profile_list -gpuId <gpuId>`
[...]

Issue #905, opened 2026-06-10
[User Story] Set Resource Type to MIG-Backed vGPU Infra Implementation


steven-chiu-bigstack:
## Description

As a CubeCOS operator, I want to select a GPU and set it to be MIG-backed for VMs so that we could pass a whole GPU card for a VM to utilize.

## Acceptance Criteria

> A detailed list of conditions that must be met for the story to be accepted. These statements must focus on functional behavior and constraints rather than specific code-level implementation.

- [x] GPU config file (`/etc/cube/cos/gpu/config.json`) is updated when the command is run successfully.
  - Please see #894 for config file schema.
- [x] Log error when the target GPU is in-use by any VM.
- [x] Re-apply GPU config (with `nvidia-smi` or any necessary tools) after node restarts.
  - Apply configs/update policies in hex_config instead of hex_sdk.

## Story Points: 3

>| Points | Effort / Resources | Description |
> | -- | -- | -- |
>| 1 | Low | Low complexity; can be resolved easily.
>| 3 | Medium | Moderate workload; requires some focus, but risks are well-managed.
>| 5 | High | High complexity; difficult to implement, requiring significant time and effort.
>| 8 | Very High | Extreme complexity or high uncertainty. If possible, this should be divided into multiple user stories.

## QA Verification

> How QA verifies this slice once it ships — written as **checkable claims**: the exact command
> plus the expected outcome.
>
> **Scope of this section.** This ticket is a `Feature` — a QA-deliverable slice of the larger story
> [#820](https://github.com/bigstack-oss/cubecos/issues/820) — and it carries the story's points, so
> it is the scheduled and QA-deliverable unit. Its slice is the **infra layer only**:
> `hex_config gpu_resource_set … migBackedVgpu` and everything it writes. The sibling slices are
> #896 (API) and #878 (UI), both already closed; the cross-slice path that none of the three covers
> alone is on the parent story #820. Everything below is this Feature's complete acceptance test —
> read the handoff rules at the end before you file anything.
>
> Reference: [`kb/cubecos/architecture/gpu-resource-set-mig-backed-vgpu.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/architecture/gpu-resource-set-mig-backed-vgpu.md)
> (bigstack-handbook#227), which also corrects what this work falsified in the SR-IOV and listing topics.
[...]

PR #1141, opened 2026-07-22
feat(gpu): SR-IOV vGPU resource-set infra (#904)

### LINKED ISSUE #904: [User Story] Set Resource Type to SR-IOV vGPU Infra Implementation
## Description

As a CubeCOS operator, I want to select a GPU and set it to be SR-IOV for VMs so that we could pass a whole GPU card for a VM to utilize.

## Acceptance Criteria

- [x] GPU config file (`/etc/cube/cos/gpu/config.json`) is updated when the command is run successfully.
  - Please see #894 for config file schema.
  - Verified on hardware. Note the carve only became functional once the `HexUtilPOpen` exit-code sentinel was quoted in #1175 (merged 2026-07-27); before that fix every attempt failed with an empty error message.
- [x] Log error when the target GPU is in-use by any VM.
  - `ResourceSetMain()` (`core/modules/config_gpu.cpp`) shells out to `hex_sdk gpu_resource_set_check`, which rejects a GPU UUID absent from `gpu_device_list` or one whose computed `status` is `inUse`. Validation runs *before* teardown, so a rejected request cannot destroy a working carve.
- [ ] Re-apply GPU config (with `nvidia-smi` or any necessary tools) after node restarts. — **deferred, not verified**
  - Apply configs/update policies in hex_config instead of hex_sdk.
  - Code-complete in `Commit()` (re-applies persisted intent unless `SriovVgpuApplied()` already sees the layout), but never exercised on hardware: the development node has no BMC or console, so a bad bootstrap on reboot would be unrecoverable. This was an agreed deferral, not an oversight; recorded as unproven in bigstack-handbook#152 so nobody downstream assumes reboot survival.

## Story Points: 3

> | -- | -- | -- |

## QA Verification

> How QA verifies this slice once it ships — written as **checkable claims**: the exact command
> plus the expected outcome.
>
> **Scope of this section.** This ticket is a `Feature` — a QA-deliverable slice of the larger story
> [#819](https://github.com/bigstack-oss/cubecos/issues/819) — and it carries the story's points, so
> it is the scheduled and QA-deliverable unit. Its slice is the **infra layer only**:
> `hex_config gpu_resource_set … sriovVgpu` and everything it writes. The sibling slices are #895
> (API) and #877 (UI); the cross-slice path that none of the three covers alone is on the parent
> story #819. Everything below is this Feature's complete acceptance test — read the handoff rules
> at the end before you file anything.
>
> Reference: [`kb/cubecos/architecture/gpu-resource-set-sriov-vgpu.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/architecture/gpu-resource-set-sriov-vgpu.md)
> (bigstack-handbook#152, amended by #176).
> Shipped by [#1141](https://github.com/bigstack-oss/cubecos/pull/1141) (the feature),
> [#1160](https://github.com/bigstack-oss/cubecos/pull/1160) (best-effort per-card apply in
> `Commit()`) and [#1175](https://github.com/bigstack-oss/cubecos/pull/1175) (**the carve does not
> work at all without this one** — see 1.1).
>
> **Tick a row only if you actually exercised it.** The four resource types are easy to confuse
> and they fail differently: a check that passes on a native card proves nothing about an SR-IOV
> one, and nothing at all about MIG-backed. Where a row does not name a type, it applies to the
> one named in that section's heading.

### The three stores you are checking, and which one is authoritative

| Store | Holds | Durable? |
|---|---|---|
| sysfs (`/sys/bus/pci/devices/<addr>/…`) | VFs enabled, per-VF `current_vgpu_type` | No — lost on reboot |
| `/etc/cube/cos/gpu/config.json` | declared intent: which card is which type, with which profiles | **Yes — the truth file** |
| `/etc/nova/nova.d/gpu.conf` | Nova PCI aliases + passthrough whitelist | Derived — regenerated wholesale from the truth file |

Flow is one-way: validate → mutate hardware → persist intent → regenerate the Nova drop-in →
restart Nova. Nothing reads back from Nova or sysfs to decide what a card *should* be.

**A CubeCOS SR-IOV vGPU is PCI passthrough of a virtual function, not Nova's mdev `resources:VGPU`
mechanism.** A flavor consumes it with `pci_passthrough:alias:<alias>:1`. If you test with a
`resources:VGPU=1` flavor the VM will not schedule and that is not a defect.

### Preconditions

- **An NVIDIA GPU node whose cards support SR-IOV vGPU** (e.g. RTX PRO 6000 Blackwell). No
  substitute exists; a node without one cannot exercise any check here.
- **A node you can power-cycle out of band (BMC or console).** Section 6 is the one acceptance
  criterion that was never verified on hardware, and it is the highest-value thing in this pass.
  Dev's node had no BMC, which is why it was deferred rather than tested.
- `/etc/cube/cos/gpu/config.json` exists (`Commit()` creates it as `[]` on first run).
- Record the identifiers once; **the two are not interchangeable** and mixing them fails silently:

```bash
hex_sdk gpu_device_list | jq -r '.[] | "\(.id)  \(.pciAddress)  \(.type)  \(.status)"'
GPU=GPU-…                       # the nvidia-smi UUID — what gpu_resource_set and profile_list take
PCI=00000001:C8:00.0            # nvidia-smi form, 8-char uppercase domain
SYSFS=0001:c8:00.0              # sysfs form, 4-char lowercase — strip 4 chars AND lowercase
hex_sdk gpu_vgpu_profile_list "$GPU" | jq '.sriov[] | {id, name, vramMiB, alias, vmCountLimit}'
```

### Setup — the node this section needs

This section needs a node that can **carve vGPU**, which is a stricter requirement than "has an
NVIDIA GPU". Confirm it before running anything below:

```bash
hex_sdk -f json gpu_device_list | jq -r '.[] | "\(.pciAddress) \(.resourceType) \(.supportTypes|join(","))"'
hex_config gpu_resource_set 2>&1 | head -2
```

Two things must be true:

1. **At least one card lists `sriovVgpu` or `migBackedVgpu` in `supportTypes`.** A card showing
   only `pgpu` is either genuinely passthrough-only hardware, or it is currently bound to
   `vfio-pci` and therefore invisible to `nvidia-smi` — release it with
   `hex_sdk gpu_unbind_vfio_pci <pciAddress>` and look again before concluding the hardware cannot
   do it. A build predating [cubecos#1282](https://github.com/bigstack-oss/cubecos/pull/1282)
   also reports this wrongly.
2. **`hex_config gpu_resource_set` prints a usage line rather than "not found".** A node whose
   `hex_config` has no GPU verbs predates #904 and cannot run any of this.

**On a GPU node that fails either check, every vGPU row below fails for the wrong reason — that is
not a defect, it is the wrong machine.** A node with a GPU but only native / pass-through support,
and a node with no GPU at all, both produce failures that look like product defects and are not.
If no node in front of you passes both checks, stop and ask for one rather than filing what you
see.

### 1 · The build carries the fixes

|---|---|---|
| 1.1 | `strings /usr/sbin/hex_config \| grep -c 'echo "\|__HEX_RC_'` | non-zero. The exit-code sentinel passed to `sh -c` **must be quoted**; unquoted, the shell reads its `\|` characters as pipes, aborts on a syntax error, `sriov-manage` never runs, and the missing marker is parsed as `rc=-1` — so every carve failed with an empty error message and the unbindLock retry branch was unreachable. The grep looks for the opening double quote immediately before the marker, which only the fixed build has. If it returns `0`, nothing below will pass and the cause will not look like this. |
| 1.2 | `strings /usr/sbin/hex_config \| grep -c 'Cannot obtain unbindLock'` | non-zero — the 5×/1 s retry policy is present in the C++ apply path, not only in the shell teardown path |
| 1.3 | `hex_config gpu_resource_set` (no arguments) | usage text, non-zero exit |

### 2 · Validation rejects bad input **without touching the hardware**

Run this block first. Validation deliberately runs before teardown, so a malformed request must
not be able to destroy a working carve. Note the VF count before and after — it must not move.

```bash
cat /sys/bus/pci/devices/$SYSFS/sriov_numvfs      # record this
```

|---|---|---|
| 2.1 | `hex_config gpu_resource_set "$GPU" badtype` | non-zero; `invalid type 'badtype'. Must be one of: pgpu, sriovVgpu, migBackedVgpu`. **There is no `unset` value** — you cannot return a card to unconfigured through this command (a known product gap, not a defect to file here). |
| 2.2 | `hex_config gpu_resource_set "$GPU" pgpu '[{"id":1,"count":1}]'` | non-zero; `profiles is not allowed when new_type is pgpu` |
| 2.3 | `hex_config gpu_resource_set "$GPU" sriovVgpu` | non-zero; `profiles is required` |
| 2.4 | `hex_config gpu_resource_set "$GPU" sriovVgpu 'oops'` | non-zero; profiles must be a non-empty JSON array of `{id,count}` |
| 2.5 | `hex_config gpu_resource_set "$GPU" sriovVgpu '[{"id":9999,"count":1}]'` | non-zero — id 9999 is not in `gpu_vgpu_profile_list`'s `sriov` array |
| 2.6 | `hex_config gpu_resource_set "$GPU" sriovVgpu '[{"id":<valid>,"count":<sriov_totalvfs + 1>}]'` | non-zero; the summed count must fit the PF's `sriov_totalvfs` (one vGPU instance = one VF) |
| 2.7 | `hex_config gpu_resource_set GPU-does-not-exist sriovVgpu '[{"id":<valid>,"count":1}]'` | non-zero; `GPU … not found` |
| 2.8 | after 2.1–2.7: `cat /sys/bus/pci/devices/$SYSFS/sriov_numvfs` and `jq . /etc/cube/cos/gpu/config.json` | **unchanged from the values you recorded.** Any movement here is a serious defect — it means validation ran after teardown. |

### 3 · The carve itself

```bash
hex_config gpu_resource_set "$GPU" sriovVgpu '[{"id":<profileId>,"count":2}]' ; echo "rc=$?"
```

|---|---|---|
| 3.1 | the command above | `rc=0` |
| 3.2 | `cat /sys/bus/pci/devices/$SYSFS/sriov_numvfs` | **every VF the PF supports** (e.g. `48`), *not* `2`. `sriov-manage -e` enables all of them; only the requested ones get a type. **Never read `sriov_numvfs` as "how many vGPUs are carved"** — `config.json`'s counts are the answer. |
| 3.3 | `cat /sys/bus/pci/devices/$SYSFS/virtfn{0,1}/nvidia/current_vgpu_type` | the requested profile id, twice |
| 3.4 | `cat /sys/bus/pci/devices/$SYSFS/virtfn2/nvidia/current_vgpu_type` | `0` — VFs beyond the request are enabled but uncarved |
| 3.5 | `jq . /etc/cube/cos/gpu/config.json` | one entry for `$GPU`: `type: "sriovVgpu"`, the card `name`, `pciAddress` in **nvidia-smi form** (8-char uppercase), and `profiles: [{id, name, count, alias}]` where `name` is the full vGPU type name and `alias` is that name lowercased with non-alphanumerics squeezed to `_` and `_<id>` appended (`"NVIDIA RTX Pro 6000 Blackwell DC-3Q"` + `1519` → `nvidia_rtx_pro_6000_blackwell_dc_3q_1519`) |
| 3.6 | `cat /etc/nova/nova.d/gpu.conf` | a `[pci]` section with **one `alias = {…}` per profile** and **one `passthrough_whitelist = {…}` per assigned VF** (so 2 here), each carrying the VF's own vendor/product id read from sysfs |
| 3.7 | `ls -l /etc/nova/nova.d/gpu.conf` | `0644` — Nova runs non-root and the file carries no secrets, so this permission is deliberate. `0600` would make Nova blind to it. |
| 3.8 | `systemctl is-active openstack-nova-compute` and `systemctl show -p NRestarts --value openstack-nova-compute` | `active`, counter not climbing — `hex_config restart_nova` ran at the end of the write path |
| 3.9 | `hex_sdk gpu_device_list \| jq '.[] \| select(.id=="'"$GPU"'") \| {type, status, allocation}'` | `type: "sriovVgpu"`, `allocation: {current: 0, total: 2}` |
| 3.10 | `journalctl -u hex_config --since "-5min" \| grep -i sriov` (or the hex log) | no error lines, and **no empty-message failures** — an empty error message is the 1.1 signature |

### 4 · The in-use guard — never hardware-verified, please close it

The acceptance criterion "log error when the target GPU is in-use by any VM" was satisfied by
reading the code (`gpu_resource_set_check` rejects a GPU whose computed `status` is `inUse`), not
by watching it reject anything. This is the cheapest real gap in this ticket.

1. Boot a VM onto the carved vGPU (see #819 §flavor, or `pci_passthrough:alias:<alias>:1`).
2. `hex_sdk gpu_device_list | jq -r '.[] | select(.id=="'"$GPU"'") | .status'` → `inUse`
3. `hex_config gpu_resource_set "$GPU" sriovVgpu '[{"id":<otherProfile>,"count":1}]' ; echo "rc=$?"`

Expected: **non-zero**, with `GPU … is in-use` on stderr, and — critically — `sriov_numvfs`,
`config.json` and the running VM all **unchanged**. A request that is rejected must cost nothing.

Then repeat step 3 for `pgpu` as the target type: the guard is type-agnostic and must reject that
too.

### 5 · Switching away — stale Nova entries must disappear

```bash
openstack server delete <the vm> --wait     # release in-use first
hex_config gpu_resource_set "$GPU" pgpu ; echo "rc=$?"
```

|---|---|---|
| 5.1 | the command above | `rc=0` |
| 5.2 | `cat /sys/bus/pci/devices/$SYSFS/sriov_numvfs` | `0`. VF teardown goes through `sriov-manage -d`; a raw `echo 0 > sriov_numvfs` is **always** rejected (`write error: No such file or directory`) because it skips the driver's unbind handshake — do not use it to "clean up", and do not file its failure as a bug. |
| 5.3 | `cat /etc/nova/nova.d/gpu.conf` | the card's alias and whitelist entries are **gone**. The drop-in is regenerated wholesale from the truth file, so there is no cleanup code — stale entries simply stop being emitted. |
| 5.4 | `lspci -nnk -s $SYSFS \| grep "in use"` | `vfio-pci` — the card is now a passthrough device |
| 5.5 | `systemctl is-active openstack-nova-compute` | `active`, and the PF is **not** left unbound from every driver |
| 5.6 | now carve it back: `hex_config gpu_resource_set "$GPU" sriovVgpu '[{"id":<profileId>,"count":1}]'` | `rc=0`. Going `pgpu → sriovVgpu` requires releasing the card from vfio-pci first (a `pgpu` is invisible to `nvidia-smi`, so profile validation would otherwise reject every profile as unsupported); every failure path re-binds it, so a rejected request leaves the card exactly as it was. |

### 6 · Reboot re-apply — the deferred acceptance criterion

**This is unproven on hardware.** The code path exists in `Commit()` (re-apply persisted intent
unless `SriovVgpuApplied()` already sees the layout), but dev's GPU node had no BMC or console, so
a bad bootstrap on reboot would have been unrecoverable. It was an agreed deferral, recorded as
unproven in the handbook so nobody downstream assumes reboot survival. **Please run it on a node
you can power-cycle out of band.**

```bash
openstack server delete <any vm on the card> --wait     # nothing may be in-use across the reboot
jq . /etc/cube/cos/gpu/config.json                      # record the intent
md5sum /etc/nova/nova.d/gpu.conf                        # record the derived file
reboot
```

After it comes back:

|---|---|---|
| 6.1 | `cat /sys/bus/pci/devices/$SYSFS/sriov_numvfs` | the full VF count again — sysfs is not durable, so this proves `Commit()` re-applied |
| 6.2 | `cat /sys/bus/pci/devices/$SYSFS/virtfn0/nvidia/current_vgpu_type` | the persisted profile id |
| 6.3 | `jq . /etc/cube/cos/gpu/config.json` | byte-identical to what you recorded — the truth file is durable and must not be rewritten by the re-apply |
| 6.4 | `md5sum /etc/nova/nova.d/gpu.conf` | same hash. The drop-in is regenerated at boot but **Nova is deliberately not restarted** — the content is a pure function of `config.json`, so regeneration is self-healing only. |
| 6.5 | boot a VM onto the alias again | schedules successfully. If it fails **and** the log shows Nova starting before the GPU module's `Commit()`, that is an ordering defect worth reporting precisely — it is the specific risk this deferral left open. |
| 6.6 | `journalctl -b \| grep -iE 'gpu\|sriov' \| head -40` | the re-apply is visible, and a per-card failure is logged-and-continued rather than fatal. GPUs are optional hardware; one card's hiccup must not fail the whole node's policy apply. (A `pgpu` re-bind failure **is** fatal, by design.) |

### 7 · Known behaviour — do not file these as defects

| Observation | Why it is expected |
|---|---|
| `sriov_numvfs` is 48 after asking for 2 | `sriov-manage -e` enables every VF the PF supports; only the requested ones are carved and whitelisted |
| `hex_config gpu_resource_set "$GPU" migBackedVgpu '[…]'` falls through to "unhandled type" | MIG-backed is accepted by the argument parser but not implemented — [#905](https://github.com/bigstack-oss/cubecos/issues/905) |
| A **mixed-capacity** carve (two different VRAM sizes on one card) is refused, or fails during apply and tears the card back down | **depends on the build.** [#1316](https://github.com/bigstack-oss/cubecos/issues/1316) delivered heterogeneous carving (merged 2026-08-25) together with the per-capacity Nova aliases that make the VFs distinguishable. On a build containing it, this is a supported path with its own acceptance test on #1316. On an older build the fail-fast teardown is correct. Two same-capacity profiles on one card were always allowed. |
| There is no way to set a card back to "unset" | no such command value exists; a known product gap |
| A card whose UUID is absent from `config.json` shows up as `unset` with no error | silent by design in the read path — check `config.json` before filing it |
| A VM on the vGPU shows "No devices found" inside the guest | the carve is only half of a working GPU VM: the guest needs a matching NVIDIA GRID driver, and on Blackwell that means the **open kernel module** build. Not an infra defect. |

### Cleanup

Restore every card you re-carved to the resource type it had when you started —

```bash
hex_sdk -f json gpu_device_list | jq -r '.[] | "\(.pciAddress) \(.resourceType)"'
```

— and delete the VMs you booted. A card left carved blocks the next person's test and, on a shared
GPU node, is indistinguishable from a real allocation.

### Already verified on the vGPU test node — and what that did not cover

The carve, the validation order, the truth file, the Nova drop-in and the switch-away path were all
exercised on a Blackwell node against `sriovVgpu`. What that did **not** cover, and what the rows
above are therefore load-bearing for:

- **§6, reboot re-apply** — never run on hardware at all. The development node had no BMC, so a bad
  bootstrap would have been unrecoverable. This is the single highest-value row in this pass.
- **§4, the in-use guard** — reasoned from the code and never hardware-verified with a real VM
  holding the VF.
- Anything driven **through the UI or the API** rather than `hex_config` directly — those are #877
  / #895, and the seams between them are on #819.

- [x] **QA Status declared at Done** — this is a `Feature`, which carries a `QA Status`, and the dev
      has declared **`Ready to QA`**. (A `Done` ticket left empty is undecided and gets flagged.)
- **A defect you find is a new, unpointed `Bug` filed under this Feature** — a closed parent is fine,
  and the rework is covered by the 3 points this Feature already earned. **Never reopen this ticket**;
  its velocity is frozen.
- **§4 and §6 are the two things most likely to produce a Bug**, and both are cases where dev
  verification stopped short rather than passed: §4 (the in-use guard) was satisfied by reading the
  code, and §6 (reboot re-apply) was an agreed deferral. A failure in either is a real finding, not a
  test-environment problem.
- **Genuinely new scope is a new pointed ticket.** The two known gaps in §7 already have homes — MIG
  is [#905](https://github.com/bigstack-oss/cubecos/issues/905), and "no way to set a card back to
  unset" has none yet, so that one is worth a new ticket rather than a Bug against this Feature.

- [x] **Handbook knowledge update** — landed as bigstack-handbook#152 (merged 2026-07-28),
      `kb/cubecos/architecture/gpu-resource-set-sriov-vgpu.md`: the write path's validation
      order, the `sriov-manage` unbindLock handshake, the Nova PCI drop-in regenerated from
      `config.json`, and which failure modes are hardware-confirmed.
      (This section was added retroactively — the issue predates the DoD block in
      `.github/ISSUE_TEMPLATE/user-story.md`. An earlier DoD sweep mis-mapped this issue to
      handbook#132, which documents the *read* path only; #152 is the correct entry.)

SekiXu:
#### What type of PR is this?

/kind feature

Implements SR-IOV vGPU support for `hex_config gpu_resource_set`, so an
operator can carve a GPU into SR-IOV vGPU-backed VFs and hand them to VMs,
instead of only pGPU passthrough.

- Validates the requested vGPU profiles (format, existence, VF capacity)
  before touching hardware.
- Applies the partition on real hardware: enables VFs via `sriov-manage -e`
  and assigns each VF's vGPU type, with fail-fast teardown if any
  assignment fails partway through.
- Resolves full vGPU type names and derives Nova PCI passthrough aliases
  per the #894 schema/formula.
- Persists `/etc/cube/cos/gpu/config.json` and regenerates the Nova PCI
  drop-in (`/etc/nova/nova.d/gpu.conf`), then restarts `nova-compute`.
- Re-applies the SR-IOV partition on node boot via `Commit()`, and drops
  the stale Nova drop-in entries when a card is switched away from
  `sriovVgpu`.
- Hardens `sriov-manage` calls (both enable and the fail-fast teardown
  path) with a bounded retry against a real, hardware-confirmed transient
  failure mode: another process (`cube-cos-api`) holding an NVML session
  open blocks the driver's unbind lock. Retries up to 5x/1s only on that
  specific `Cannot obtain unbindLock` failure; anything else fails fast.
- Fixes a pre-existing bug in `gpu_unset_current_type` (sdk_gpu.sh, not
  introduced by this PR) that made switching a GPU *away* from
  `sriovVgpu` fail end-to-end: it didn't lowercase the PCI address before
  building the sysfs path, and it zeroed `sriov_numvfs` via a raw sysfs
  write without going through `sriov-manage -d`'s unbind handshake, which
  the driver rejects. Bundled here because it directly blocks practical
  use of this feature (configure once, cannot reconfigure/undo).

Fixes #904

**AC verification status** (all 3 ACs, verified on real hardware — cn13,
NYCU GPU node — 2026-07-20/21, plus unit harness 32/32 green):

- ✅ **AC1** (`config.json` updated on success): confirmed end-to-end —
  `hex_config gpu_resource_set <gpu> sriovVgpu '[{"id":1519,"count":2}]'`
  → `rc=0`, sysfs (`sriov_numvfs`, per-VF `current_vgpu_type`),
  `config.json`, `/etc/nova/nova.d/gpu.conf`, and
  `openstack-nova-compute` all correct.
- ✅ **AC2** (log error when target GPU is in-use by a VM): delegated to
  the pre-existing, type-agnostic `gpu_resource_set_check` (sdk_gpu.sh),
  which every `gpu_resource_set` call already runs through and which
  already has this check — not modified by this PR. Verified by code
  read, not by hardware-triggering an actual in-use rejection this round
  (no VM was attached to the test GPU during this pass; the full
  flavor+VM-attach path was exercised in an earlier manual SOP run).
- ⏳ **AC3** (re-apply config after reboot): code path exists
  (`Commit()`'s `SriovVgpuApplied` detection, this PR), but an actual
  node reboot was **not** tested — cn13 has no BMC/console, so a bad
  bootstrap on reboot is unrecoverable. This was an explicit, agreed
  deferral, not an oversight; needs a separate go/no-go before ever
  attempting on this node.

**Known residual gaps** (disclosed, not blocking, worth reviewer
awareness):

- `TeardownSriov`'s fail-fast rollback path (used when a VF assignment
  fails partway through `ApplySriovVgpu`) was rewritten to the same
  `sriov-manage -d` + retry pattern as the `gpu_unset_current_type` fix
  above, but that specific failure-injection trigger was never hit in
  any test session — the fix is by-construction (identical helper,
  identical call as the now-verified path), not independently
  hardware-exercised via its own trigger.
- The `sriov-manage` unbind-lock contention this PR retries around has a
  structural root cause upstream: `cube-cos-api` never releases its NVML
  session for the life of the process, so on any real deployment it can
  hold the unbind lock indefinitely. That's out of scope for this repo —
  a fix proposal has been drafted for the `cube-cos-api` owner
  separately. This PR's retry is a defensive layer on the `hex_config`
  side regardless of whether/when that gets addressed.

#### Additional documentation

```docs
Testing performed (cn13, NYCU GPU node, GPU 3 =
GPU-0a5ba9ad-a575-72c7-d51d-3e726232cfd1 @ 0001:c8:00.0):

1. Unit harness (Mac, no GPU needed): `bash harness/run.sh` under
   `.claude/scratch/feat-904-sriov-vgpu-infra/` — 32/32 passed (profile
   validation, sysfs address normalization, VF assignment planning,
   alias derivation, gpu.conf content generation).

2. Negative-path, real hardware (2026-07-20): invalid type, malformed
   JSON, unknown profile id, and over-capacity count all rejected with
   rc=1 and no hardware touched (sriov_numvfs stayed 0).

3. Apply path, real hardware (2026-07-20):
   `hex_config gpu_resource_set GPU-0a5ba9ad-a575-72c7-d51d-3e726232cfd1 \
     sriovVgpu '[{"id":1519,"count":2}]'`
   -> rc=0. Verified after: sriov_numvfs=48, virtfn0/virtfn1
   current_vgpu_type=1519, config.json has the full entry (id/name/type/
   pciAddress/profiles per #894 schema), /etc/nova/nova.d/gpu.conf
   regenerated (1 alias + 2 passthrough_whitelist entries), and
   openstack-nova-compute stayed active. This is AC1's hardware evidence.

4. Flavor + VM attach, real hardware (2026-07-17, precursor run before
   this code existed): hand-crafted the same gpu.conf shape this PR's
   code now generates, created a flavor with the matching
   pci_passthrough:alias, booted a VM, and confirmed via
   `virsh dumpxml` a <hostdev> pointing at the VF's PCI address —
   i.e. the config shape this PR emits is proven to make Nova actually
   schedule and attach a VF to a guest. Note: this run predates the
   code and didn't go through hex_config; the 2026-07-20 run above
   proves hex_config emits that same shape, but the two were not
   re-combined in a single pass (no VM was created against
   hex_config's own generated config in this round).

5. gpu_unset_current_type fix, real hardware (2026-07-21): after fixing
   the case-sensitivity + missing-per-VF-clear bug, re-ran unset on an
   already-configured sriovVgpu GPU -> correctly invoked
   `sriov-manage -d` and reported "already has VFs disabled" (idempotent
   success) instead of the old cascading failure that used to leave the
   PF with no driver bound at all.

6. sriov-manage retry logic, real hardware (2026-07-21): reproduced the
   cube-cos-api unbind-lock contention with a synthetic NVML-hold
   program. A 12s hold correctly exhausts all 5 retry attempts (1s
   apart, matching the real "Cannot obtain unbindLock" kernel message)
   and fails cleanly with GPU state untouched (numvfs=0, nothing
   half-applied). A 6s hold releases mid-retry and the apply completes
   end-to-end (numvfs=48, correct vgpu type, config.json correct).

Not tested / explicitly deferred:

- AC3 (reboot re-apply): cn13 has no BMC/console, so this needs a
  separate go/no-go before ever attempting it on this node.
- TeardownSriov's fail-fast rollback path: fixed to the same
  sriov-manage -d + retry pattern as #5 above, but the actual trigger
  (a VF assignment failing partway through) was never hit in testing,
  so it's unverified by direct hardware trigger.
```


Eandalf-Bigstack:
# Code Review: SR-IOV vGPU Resource-Set Infrastructure (PR #904)

## Summary
Solid implementation of SR-IOV vGPU partitioning with comprehensive validation, retry logic for transient failures, and persistent configuration. The PR correctly prioritizes safety (validate before modifying) and includes a fix for a pre-existing bug in `gpu_unset_current_type`.

---

## Critical Issues

None identified that would block merge.

---

## High-Priority Concerns

### 1. **SysFS Path Assumptions**
The code assumes `/sys/bus/pci/devices/{pciAddr}/virtfn{N}/nvidia/current_vgpu_type` exists for all NVIDIA vGPU devices. This directory structure is NVIDIA-driver-version-dependent. Add a defensive check in `ApplySriovVgpu` before attempting writes:
```cpp
// Verify the sysfs path exists before writing
if (access(typePath.c_str(), F_OK) != 0) {
    HexLogError("vGPU sysfs path not found; check NVIDIA driver version");
    TeardownSriov(sysfsAddr);
    return false;
}
```

### 2. **Incomplete Error Context in jq Command**
Line ~717 (`ResourceSetMain`, sriovVgpu path): When the jq pipeline fails, error logs only show generic "failed to build updated GPU config". The jq stderr is silently dropped. Redirect jq errors for debugging:
```bash
jq ... 2>&1 | tee /tmp/jq_error.log
```
Or check jq's exit status and log its stderr separately.

### 3. **Race Condition: config.json Read-Modify-Write**
`WriteNovaGpuConf()` reads `config.json`, iterates it to build VF lists, then writes `nova.conf`. Between read and write, another process could modify `config.json`, causing the generated `nova.conf` to be stale. While unlikely in practice, consider reading config once at function entry and passing it as parameter instead of re-reading mid-operation.

---

## Medium-Priority Issues

### 4. **Fragile Exit Code Parsing**
`RunSriovManageWithRetry()` uses `rfind("HEX_RC:")` to extract the exit code. If `sriov-manage` outputs "HEX_RC:" in its normal output, this breaks. More robust approach:
```cpp
const std::string cmd = "... ; echo |__HEX_RC_$?__|";  // Unique delimiter
const size_t start = output.rfind("|__HEX_RC_");
```

### 5. **Missing Vendor/Product ID Validation**
`BuildNovaGpuConfContent()` uses `vfs[0].vendorId/productId` for all aliases, assuming all VFs share the same IDs. Add assertion:
```cpp
for (const SriovVf& vf : vfs) {
    if (vf.vendorId != vfs[0].vendorId || vf.productId != vfs[0].productId) {
        HexLogError("Inconsistent VF vendor/product IDs for GPU %s", gpuId.c_str());
        return "";
    }
}
```

### 6. **Unreadable sriov_totalvfs Falls Through Silently**
In `ValidateVgpuProfiles()`, if `sriov_totalvfs` is unreadable, the capacity check is skipped without logging. This masks configuration issues:
```cpp
if ((totalVfsStream >> totalVfs)) {
    // ... capacity check ...
} else {
    HexLogWarning("Could not read sriov_totalvfs for GPU %s; skipping capacity check", gpuId);
}
```

---

## Low-Priority Observations

- **Defensive null/empty checks**: Code generally handles edge cases well (empty profiles array, missing sysfs files).
- **Hardcoded retry count**: 5 retries × 1s delay is reasonable; document why in a comment if needed.
- **Permission on nova.conf**: `chmod 644` is correct (non-root nova must read it).
- **Commit() idempotency**: `SriovVgpuApplied()` check prevents re-applying already-configured VFs on runtime commits. Good.
- **TODO(#905)**: MIG-backed vGPU capacity validation deferred; acceptable as separate issue.

---

## Verification Checklist

✅ Validation runs before any hardware modifications  
✅ Retry logic targets only the specific transient failure (`Cannot obtain unbindLock`)  
✅ Pre-existing `gpu_unset_current_type` bug is fixed (lowercase + sriov-manage)  
✅ config.json used as single source of truth  
✅ Nova config regenerated from config.json (auto-cleans stale entries)  
✅ Reboot re-apply path checks `SriovVgpuApplied()` to avoid disturbing running VMs  
✅ Test coverage: 32/32 unit harness passed; AC1 hardware-verified end-to-end  

**AC3 deferral is documented and acceptable** (reboot test blocked by hardware constraints).

---

**Recommendation:** Approve with the above issues addressed, particularly #1 (sysfs path check) and #2 (error context).

Eandalf-Bigstack:
This is a well-structured pull request. You’ve successfully prioritized safety by validating the profiles before mutating hardware state, and the idempotent retry logic for `sriov-manage` is a smart way to handle the known `unbindLock` contention.

I’ve reviewed the diff against the codebase and evaluated the prior discussion points. Eandalf-Bigstack correctly identified several important edge cases. Below is the synthesized review with actionable recommendations.

## High-Priority Issues

### 1. Silent Bypass of Capacity Validation (Logic Flaw)

In `ValidateVgpuProfiles`, if the `sriov_totalvfs` sysfs file cannot be read, the stream extraction `totalVfsStream >> totalVfs` evaluates to false. This causes the code to silently skip the `requestedTotal > totalVfs` check entirely, allowing an invalid configuration to proceed.
**Fix:** Explicitly handle the read failure so that a missing or unreadable file fails the validation.

```cpp
// ValidateVgpuProfiles
long totalVfs = 0;
if (!(totalVfsStream >> totalVfs)) {
    HexLogError("gpu_resource_set: could not read sriov_totalvfs for GPU %s; failing capacity check", gpuId);
    return false;
}
if (requestedTotal > totalVfs) {
    HexLogError("gpu_resource_set: requested %ld vGPU(s) exceeds the %ld VF(s) available", requestedTotal, totalVfs);
    return false;
}

```

### 2. Fragile Exit Code Parsing (Robustness)

In `RunSriovManageWithRetry`, extracting the exit code via `rfind("HEX_RC:")` assumes that `sriov-manage` will never organically output the string `HEX_RC:`. If it does, the offset calculation `atoi(output.c_str() + marker + 7)` will grab garbage data or crash.
**Fix:** Use a highly unique delimiter to guarantee clean extraction, as suggested in the prior review.

```cpp
const std::string cmd = HexUtilStringFormat("%s %s %s 2>&1; echo |__HEX_RC_$?__|", NVIDIA_SRIOV, op, sysfsAddr.c_str());
// ... 
const size_t marker = output.rfind("|__HEX_RC_");
const int rc = (marker != std::string::npos) ? atoi(output.c_str() + marker + 10) : -1;

```

---

## Medium-Priority Issues

### 3. Masked `jq` Errors (Operability)

When the `jq` pipeline fails during the `config.json` rewrite in `ResourceSetMain`, the application returns a generic `failed to build updated GPU config` message. The actual `jq` syntax/execution error on `stderr` is either dumped uncaptured to the parent shell or lost entirely, making debugging difficult in production.
**Fix:** Capture `stderr` in the `jq` command so it can be logged, or redirect it to a known debug location:

```bash
jq -c ... 2>> /var/log/cube/jq_gpu_config_errors.log > %s

```

### 4. Assumption of Homogeneous Vendor/Product IDs

In `BuildNovaGpuConfContent()`, the loop assigns `vfs[0].vendorId` and `vfs[0].productId` to all aliases without verifying that the subsequent VFs actually match. While this is *usually* true for VFs carved from the same physical GPU, asserting it prevents obscure Nova scheduling bugs if a driver glitch returns inconsistent IDs.
**Fix:** Add a quick validation loop when you populate `vfsByGpuId`:

```cpp
for (const SriovVf& vf : vfs) {
    if (vf.vendorId != vfs[0].vendorId || vf.productId != vfs[0].productId) {
        HexLogError("Inconsistent VF vendor/product IDs on GPU %s", gpuId.c_str());
        return false;
    }
}

```

### 5. `config.json` Read-Modify-Write Race

`ResourceSetMain` writes to `config.json`, and then immediately calls `WriteNovaGpuConf()`, which opens and reads `config.json` *again* from disk. If another process modifies the configuration in the split second between the write and the re-read, the resulting Nova configuration will be inconsistent with the state you just applied.
**Fix:** Consider parsing the JSON object *once* at the beginning of your flow and passing the in-memory `json11::Json` object directly to `WriteNovaGpuConf(const json11::Json& gpuConfig)` rather than relying on sequential file reads.

---

## Low-Priority / Nitpicks

* **Sysfs Path Pre-Check:** Eandalf suggested verifying the sysfs path exists before calling `WriteSysfs`. Your implementation inside `WriteSysfs` (checking `!stream.is_open()`) is completely sufficient and avoids a Time-of-Check to Time-of-Use (TOCTOU) race condition. I recommend keeping your approach, as `TeardownSriov` is correctly triggered if it fails.
* **Bash string escaping:** In `sdk_gpu.sh`, `echo "$output" | grep -q "Cannot obtain unbindLock"` is mostly safe, but using `printf "%s\n" "$output" | grep -q ...` avoids edge cases where `output` might contain backslashes or start with `-e`.

core/modules/config_gpu.cpp
```diff
@@ -140,12 +650,95 @@ ResourceSetMain(int argc, char* argv[])
...
+        // alias/whitelist entries from the Nova drop-in (regenerated from
+        // the just-updated truth file). Nodes that never had a drop-in are
+        // left untouched.
+        if (access(NOVA_GPU_CONF, F_OK) == 0) {
+            if (!WriteNovaGpuConf()) {
+                HexLogError("gpu_resource_set: failed to regenerate %s for GPU %s", NOVA_GPU_CONF, gpuId);
+                return EXIT_FAILURE;
+            }
+
+            if (HexUtilSystemF(0, 0, HEX_CFG " restart_nova") != 0) {
```
Eandalf-Bigstack, line 669: It is a bit weird to call hex_config itself inside hex_config's binary. We could call the underlying function of `hex_config restart_nova` instead of executing an external process via bash then back to hex_config.

core/modules/config_gpu.cpp
```diff
@@ -140,12 +650,95 @@ ResourceSetMain(int argc, char* argv[])
...
+            HexLogError("gpu_resource_set: failed to persist GPU %s config", gpuId);
+            return EXIT_FAILURE;
+        }
+
+        if (!WriteNovaGpuConf()) {
+            HexLogError("gpu_resource_set: failed to regenerate %s for GPU %s", NOVA_GPU_CONF, gpuId);
+            return EXIT_FAILURE;
+        }
+
+        if (HexUtilSystemF(0, 0, HEX_CFG " restart_nova") != 0) {
```
Eandalf-Bigstack, line 739: It is a bit weird to call hex_config itself inside hex_config's binary. We could call the underlying function of `hex_config restart_nova` instead of executing an external process via bash then back to hex_config.

core/modules/config_gpu.cpp
SekiXu: Dug into this before making the change, and it turns out the obvious fix (expose `NovaService` via a header, call it in-process) would actually regress a path we've hardware-verified works.

`NovaService`'s behavior branches entirely on `s_eCubeRole` (a file-local static in `config_nova.cpp`), which only gets set inside `NotifyCube()` - and `NotifyCube` only runs as part of `NotifyModules()`, the last step of `SafeParseSettings`. `SafeParseSettings` only runs for commands registered `CONFIG_COMMAND_WITH_SETTINGS` (see `hex/src/hex_config/config_main.cpp`). `gpu_resource_set` is a plain `CONFIG_COMMAND`, so calling `NovaService()` in-process from here would run against `s_eCubeRole`'s zero-initialized default, not this node's actual role - `IsControl()`/`IsCompute()` would branch on garbage, and the new PCI config might never actually get a nova restart (or restart the wrong service set).

Shelling out to `hex_config restart_nova` looks redundant but is actually doing real work here: it forks a fresh process through the `WITH_SETTINGS` path specifically so nova's own role state gets parsed correctly before `RestartMain` runs. That's why `restart_nova` itself needs `WITH_SETTINGS` in the first place.

I checked whether there's an existing lighter-weight convention for this (a single-module parse entry point, or the `/etc/hex_config/post.d/<command>.post.sh` post-script mechanism used by `apply`/`bootstrap`) - the post-script hook exists, but it wouldn't actually remove the recursive `hex_config` call, just relocate an equivalent one into a `.post.sh` file instead of inline C++, since it'd still need to reach `restart_nova` through the same `WITH_SETTINGS` path for the same reason. There's no existing "parse just one module" shortcut anywhere in `core/modules/` to fall back on instead.

Given the current call is hardware-verified (AC1: `openstack-nova-compute` active and actually serving the new PCI config after each apply), I'd rather not change this inside this PR. Open to a follow-up ticket if there's appetite for building that lighter-weight single-module-parse entry point, but that's a bigger, separate piece of work than this feature - let me know if you'd rather track it that way.

core/modules/config_gpu.cpp
SekiXu: Dug into this before making the change, and it turns out the obvious fix (expose `NovaService` via a header, call it in-process) would actually regress a path we've hardware-verified works.

`NovaService`'s behavior branches entirely on `s_eCubeRole` (a file-local static in `config_nova.cpp`), which only gets set inside `NotifyCube()` - and `NotifyCube` only runs as part of `NotifyModules()`, the last step of `SafeParseSettings`. `SafeParseSettings` only runs for commands registered `CONFIG_COMMAND_WITH_SETTINGS` (see `hex/src/hex_config/config_main.cpp`). `gpu_resource_set` is a plain `CONFIG_COMMAND`, so calling `NovaService()` in-process from here would run against `s_eCubeRole`'s zero-initialized default, not this node's actual role - `IsControl()`/`IsCompute()` would branch on garbage, and the new PCI config might never actually get a nova restart (or restart the wrong service set).

Shelling out to `hex_config restart_nova` looks redundant but is actually doing real work here: it forks a fresh process through the `WITH_SETTINGS` path specifically so nova's own role state gets parsed correctly before `RestartMain` runs. That's why `restart_nova` itself needs `WITH_SETTINGS` in the first place.

I checked whether there's an existing lighter-weight convention for this (a single-module parse entry point, or the `/etc/hex_config/post.d/<command>.post.sh` post-script mechanism used by `apply`/`bootstrap`) - the post-script hook exists, but it wouldn't actually remove the recursive `hex_config` call, just relocate an equivalent one into a `.post.sh` file instead of inline C++, since it'd still need to reach `restart_nova` through the same `WITH_SETTINGS` path for the same reason. There's no existing "parse just one module" shortcut anywhere in `core/modules/` to fall back on instead.

Given the current call is hardware-verified (AC1: `openstack-nova-compute` active and actually serving the new PCI config after each apply), I'd rather not change this inside this PR. Open to a follow-up ticket if there's appetite for building that lighter-weight single-module-parse entry point, but that's a bigger, separate piece of work than this feature - let me know if you'd rather track it that way.

SekiXu:
Follow-up after going through both review passes point by point against the actual source, git history, and the hardware-test notes from this ticket's testing. Splitting into what changed, what's deferred, and what I'm not changing (with the reasoning).

### Fixed

- **Capacity check fail-open** (`ValidateVgpuProfiles`): the old `if ((totalVfsStream >> totalVfs) && requestedTotal > totalVfs)` silently skipped the capacity check (no rejection, no log) whenever `sriov_totalvfs` couldn't be read, instead of treating an unreadable required precondition as a hard failure. Fixed in `79734b07` - now fails closed with a log line. Confirmed real via code read, not just taking the review's word for it.
- **Fragile exit-code delimiter** (`RunSriovManageWithRetry`): swapped the `"HEX_RC:"` sentinel for the suggested `"|__HEX_RC_$?__|"` form. Real-world collision risk was negligible either way (sriov-manage is a fixed vendor script, not variable/user-influenced output), but it's a free hardening. Fixed in `44011f4f`.

### Deferred to a separate ticket (real, but not new, and not confirmed reachable in practice)

- **`config.json` read-modify-write has no locking**: confirmed the mechanism is real - `gpu_resource_set` is registered `CONFIG_COMMAND` (not `CONFIG_COMMAND_WITH_SETTINGS`), so it never goes through `SafeParseSettings`/`COMMIT_LOCKFILE`, meaning nothing serializes two concurrent invocations (e.g. configuring two different GPUs back to back) against the same `config.json` read-modify-write. That said: this exact shape (jq-merge + `mv`, no lock, same `CONFIG_COMMAND` registration) already exists in the pre-existing `pgpu` branch of this same function, predating this PR's first commit - this PR extends the pattern into the new `sriovVgpu` branch rather than introducing it. Real-world reachability (does anything actually fire concurrent `gpu_resource_set` calls for different GPUs?) isn't confirmed either way. Given it's a pre-existing gap rather than something new or uniquely blocking this feature, I'd rather track it as its own ticket than expand this PR's scope - happy to file it if that sounds right.

### Not changing, with the reasoning

- **VF vendor/product ID assumed homogeneous** (`BuildNovaGpuConfContent` using `vfs[0]`'s ids for every alias on a card): this isn't an assumption that might not hold, it's a hardware invariant - confirmed on cn13 (2026-07-17 manual SOP): `cat /sys/bus/pci/devices/<VF>/vendor /sys/bus/pci/devices/<VF>/device` reads back identical to the PF's own vendor/device id. VF PCI vendor/product id is a fixed hardware property of the physical function, not something that varies per vGPU instance/profile, so an added assertion here would be dead code that can't trigger given how this hardware actually behaves.
- **jq stderr "silently dropped"**: checked `HexUtilRunSystem_` (`hex/src/hex_sdk_library/process/process_util.cpp`) - it's debug-level-gated: stderr pipes to `logger -t <program>` (syslog) whenever `HexLogDebugLevel >= debugLevel`, and `HexLogDebugLevel` defaults to 0. Every `HexUtilSystemF` call in this codebase (including both jq calls here) passes `debugLevel=0`, so by default `0 >= 0` holds and stderr already reaches syslog - it isn't dropped. This is the standard convention across every `config_*.cpp` module, not something specific to this file.
- **`echo "$output" | grep` vs `printf "%s\n" "$output" | grep`** in `sdk_gpu.sh`: this exact `echo "$var" | ...` shape is the established idiom throughout the shell modules (same file has ~10 more instances; `sdk_cinder.sh`, `sdk_ceph.sh` do the same) - singling out this one call site would be inconsistent rather than an improvement. Real risk is also low here specifically: `$output` is `sriov-manage`'s own captured text (a fixed vendor script), not user-influenceable input, so the leading-flag/backslash edge cases this guards against aren't realistically reachable.

core/modules/config_gpu.cpp
Eandalf-Bigstack: Got it. Yep, that is the legacy architectural design issue. We could not call functions as if calling pure functions. Those functions are heavily coupled together. Leaving the current implementation of going out to a separate process and re-enter hex_config to have setting values parsed was and is the easiest way to get the function under `restart_nova` to work.

Eandalf-Bigstack (approved):

PR #1157, opened 2026-07-23
fix(rolling): survive a wedged cephfs client at staging time

### LINKED ISSUE #1156: [Bug] Rolling-upgrade staging fails when a node's cephfs client is wedged (MDS eviction after large pkg write)
## Description

During the 3.1.0 → 3.1.10 rolling-upgrade rehearsal on the sky lab, sky141's cephfs kernel client went stale (every operation returned EACCES, `ls /mnt/cephfs` Permission denied for root) within a minute of the 11GB upgrade package copy onto `/var/update` (cephfs), while ceph itself was HEALTH_OK and the other nodes' mounts were fine. The roll's staging step then failed on that node (`stage_failed`) and the job paused. The signature matches MDS client eviction/blocklisting under heavy write + capability churn; the kernel client's default behavior after eviction is to fail the mount permanently until remounted.

## How to Reproduce

1. On a 3-node HA cluster, copy an ~11GB `.pkg` to `/var/update/` (cephfs) on a converged node.
2. Start `rolling_update` immediately after the copy returns.
3. Timing-dependent (hit 1 of 2 rehearsal iterations): the writing node's cephfs client is evicted; staging fails on it with the pkg unreadable; the roll pauses.

## Expected behavior

Staging either survives a wedged cephfs client (self-heal) or refuses to start with a precise reason before any node is touched — never fails mid-stage. Client eviction should not permanently wedge `/mnt/cephfs`.

- The all-or-nothing staging gate held correctly: zero nodes rebooted, zero workload impact.
- Manual heal used in the rehearsal: `umount -l /mnt/cephfs && hex_sdk ceph_mount_cephfs`, then abort + restart the roll (staging is idempotent).
- Fix (PR linked): mount cephfs with `recover_session=clean` so an evicted client reconnects cleanly, plus a staging preflight in `_power_roll_stage_all` (per-node mount+pkg check, one self-heal attempt, pause pre-stage).
- Full rehearsal context: bigstack-handbook `kb/cubecos/known-issues/rolling-upgrade-rehearsal-findings.md` (landed via bigstack-handbook#118).

- [x] **Handbook knowledge update** — landed in `kb/cubecos/known-issues/rolling-upgrade-rehearsal-findings.md` (finding 2) via bigstack-oss/bigstack-handbook#118.

traviswu-bigstack:
#### What type of PR is this?

/kind bug

Hardening for the rolling upgrade, from the 3.1.0→3.1.10 sky rehearsal. Two independent fixes:

**1. Survive a wedged cephfs client at staging (commit 1).**
- `ceph_mount_cephfs` mounts with `recover_session=clean` so an MDS-evicted (blocklisted) kernel client reconnects instead of failing every op with EACCES until remount.
- `_power_roll_stage_all` preflights every node (cephfs mounted + pkg readable) before any `hex_install` runs, self-heals once via `ceph_mount_cephfs`, and pauses pre-stage with a precise reason instead of failing mid-stage.

**2. `continue` must not re-pause on a stale deadline or re-roll a finished node (commit 2).**
- continue set `state=running` and re-armed the watchdog BEFORE `_power_roll_kick` wrote a fresh deadline → a watchdog tick in that window read the old (blown) deadline and re-paused immediately. Now continue zeroes the deadline before `state=running` (watchdog skips the check at 0) and disarms any stale timer before re-arming.
- continue always reset the inflight node to pending and re-kicked it — even when it had actually finished (booted the target) before the pause, rebooting an already-upgraded node in a loop. New `_power_roll_node_on_target` checks the inflight node’s ACTIVE firmware against the roll target and records it **done** instead of re-rolling.

Fixes #1156 (cephfs staging wedge). Also fixes the watchdog stale-deadline / continue-re-kick edges found in the same rehearsal (finding 13; no separate issue — folded here at the reporter’s request).

- Both were hit live recovering sky141 after a lab PXE-detour stall; the manual equivalents (heal + mark-done + resume) unblocked the roll.
- `_power_roll_node_on_target` only returns true on a reachable, positive firmware match — an unreachable node still falls through to re-kick (safe default).

#### Additional documentation

```docs
bigstack-handbook kb/cubecos/known-issues/rolling-upgrade-rehearsal-findings.md (findings 2, 13)
```

https://claude.ai/code/session_0154HK1F8x8BhXYfTNBXc3AR

### DIFF
--- a/core/sdk_sh/modules/sdk_ceph.sh
+++ b/core/sdk_sh/modules/sdk_ceph.sh
@@ -2807,7 +2807,11 @@ ceph_mount_cephfs()
     done
     for i in {1..6} ; do
         mountpoint -q $CEPHFS_STORE_DIR && return 0
-        timeout $SRVSTO mount -t ceph :/ $CEPHFS_STORE_DIR -o name=admin,secretfile=/etc/ceph/admin.key >/dev/null 2>&1
+        # recover_session=clean: an MDS-evicted (blocklisted) kernel client
+        # reconnects with a clean session instead of failing every op with
+        # EACCES until someone remounts -- eviction happens under heavy write
+        # + cap churn (e.g. staging an 11G pkg right before a roll).
+        timeout $SRVSTO mount -t ceph :/ $CEPHFS_STORE_DIR -o name=admin,secretfile=/etc/ceph/admin.key,recover_session=clean >/dev/null 2>&1
         mountpoint -q $CEPHFS_STORE_DIR && return 0
         sleep 10
     done
--- a/core/sdk_sh/modules/sdk_ceph.sh
+++ b/core/sdk_sh/modules/sdk_ceph.sh
@@ -2807,10 +2807,8 @@ ceph_mount_cephfs()
     done
     for i in {1..6} ; do
         mountpoint -q $CEPHFS_STORE_DIR && return 0
-        # recover_session=clean: an MDS-evicted (blocklisted) kernel client
-        # reconnects with a clean session instead of failing every op with
-        # EACCES until someone remounts -- eviction happens under heavy write
-        # + cap churn (e.g. staging an 11G pkg right before a roll).
+        # recover_session=clean: an MDS-evicted client reconnects cleanly instead
+        # of EACCES-ing every op until remount (see PR for the eviction scenario).
         timeout $SRVSTO mount -t ceph :/ $CEPHFS_STORE_DIR -o name=admin,secretfile=/etc/ceph/admin.key,recover_session=clean >/dev/null 2>&1
         mountpoint -q $CEPHFS_STORE_DIR && return 0
         sleep 10
--- a/core/sdk_sh/modules/sdk_power.sh
+++ b/core/sdk_sh/modules/sdk_power.sh
@@ -480,6 +480,21 @@ _power_roll_stage_all()
 {
     local pkg=$1 h rc=0
     local -A _pids=()
+    # Preflight: every node reads the pkg from shared cephfs, and a wedged
+    # cephfs client (MDS-evicted after the large pkg write, mount returns
+    # EACCES) fails its stage mid-flight. Verify mount+pkg per node first,
+    # self-heal once via ceph_mount_cephfs, and pause BEFORE staging if a
+    # node still can't see the package.
+    for h in $(jq -r '.nodes[].hostname' $ROLLING_JOB 2>/dev/null) ; do
+        if ! remote_run "$h" "mountpoint -q /mnt/cephfs && test -r $pkg" </dev/null >/dev/null 2>&1 ; then
+            log_warning "stage preflight: cephfs/pkg not readable on $h, running ceph_mount_cephfs"
+            remote_run "$h" "$HEX_SDK ceph_mount_cephfs" </dev/null >/dev/null 2>&1
+            if ! remote_run "$h" "mountpoint -q /mnt/cephfs && test -r $pkg" </dev/null >/dev/null 2>&1 ; then
+                _power_roll_pause "cephfs not mounted or $(basename $pkg) not readable on $h; nothing was staged"
+                return 1
+            fi
+        fi
+    done
     # Stage every node at once. Each writes its own inactive slot from a
     # read-only package on shared cephfs, so there is nothing to serialize --
     # doing them in turn just multiplies the wait before the roll can start.
--- a/core/sdk_sh/modules/sdk_power.sh
+++ b/core/sdk_sh/modules/sdk_power.sh
@@ -480,11 +480,9 @@ _power_roll_stage_all()
 {
     local pkg=$1 h rc=0
     local -A _pids=()
-    # Preflight: every node reads the pkg from shared cephfs, and a wedged
-    # cephfs client (MDS-evicted after the large pkg write, mount returns
-    # EACCES) fails its stage mid-flight. Verify mount+pkg per node first,
-    # self-heal once via ceph_mount_cephfs, and pause BEFORE staging if a
-    # node still can't see the package.
+    # Preflight: verify each node can read the pkg from cephfs, self-heal once via
+    # ceph_mount_cephfs, and pause before staging if it still can't -- a wedged
+    # cephfs client would otherwise fail its stage mid-flight.
     for h in $(jq -r '.nodes[].hostname' $ROLLING_JOB 2>/dev/null) ; do
         if ! remote_run "$h" "mountpoint -q /mnt/cephfs && test -r $pkg" </dev/null >/dev/null 2>&1 ; then
             log_warning "stage preflight: cephfs/pkg not readable on $h, running ceph_mount_cephfs"
@@ -850,6 +848,17 @@ power_roll_status_json()
         }' $ROLLING_JOB
 }
 
+# 0 if node $1's ACTIVE firmware already matches the roll target (commit in
+# .version) -- it upgraded before/during a pause. Unreachable/mismatch => 1.
+_power_roll_node_on_target()
+{
+    local node=$1 target
+    target=$(jq -r '.version // ""' $ROLLING_JOB 2>/dev/null | grep -oE '[0-9a-f]{7,}' | tail -1)
+    [ -n "$target" ] || return 1
+    is_sshable "$node" || return 1
+    remote_run "$node" "hex_cli -c firmware list 2>/dev/null | grep -i active" </dev/null 2>/dev/null | grep -q "$target"
+}
+
 power_roll_decision()
 {
     [ -e "$ROLLING_JOB" ] || { echo "no rolling job in progress" >&2 ; return 1 ; }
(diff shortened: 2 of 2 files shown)


Eandalf-Bigstack:
Summary
- Good fixes: recover_session=clean + preflight pause, and safer continue/watchdog logic. Both address real rehearsal failures.

Critical issues & recommendations
- Shell quoting: quote variables in mount/remote commands (e.g., "$CEPHFS_STORE_DIR", "$pkg") to avoid word-splitting/globbing on local and remote shells.
- remote_run injection/expansion: wrap remote commands or use printf %q to prevent unintended expansion on the remote shell.
- Version matching fragility: _power_roll_node_on_target uses grep -oE '[0-9a-f]{7,}' — may match unrelated hex sequences. Parse .version reliably and compare canonical active SHA.
- Watchdog race: zeroing deadline reduces risk but isn’t atomic; consider an atomic state+deadline update or a lock to eliminate micro-races.
- Tests/logging: add integration tests for cephfs preflight and continue/watchdog edge-cases and improve debug logs.

Minor
- Verify SRVSTO, CEPh vars exist and bash associative arrays are supported on CI.

Eandalf-Bigstack:
Here is a code review of the proposed PR, focusing on critical issues, logic flaws, and architectural best practices.

### 1. Shell Quoting and Remote Code Execution (Critical)

When passing locally expanded variables to a remote shell command, unquoted variables can lead to word-splitting, globbing, or unintended command injection on the remote host if the path contains spaces or special characters.

* **In `sdk_power.sh`:**
```bash
if ! remote_run "$h" "mountpoint -q /mnt/cephfs && test -r $pkg" </dev/null >/dev/null 2>&1 ; then

```


If `$pkg` contains spaces (e.g., `/mnt/cephfs/my packages/v1.tar`), this translates remotely to `test -r /mnt/cephfs/my packages/v1.tar`, which errors out. You must escape/quote it.
*Fix:* Use `printf %q` to safely escape the variable, or properly nest quotes:
`remote_run "$h" "mountpoint -q /mnt/cephfs && test -r \"$pkg\""`
* **In `sdk_ceph.sh`:**
```bash
timeout $SRVSTO mount -t ceph :/ $CEPHFS_STORE_DIR -o name=admin...

```


`$CEPHFS_STORE_DIR` lacks quotes. This will break if the mount path contains spaces.
*Fix:* Quote the variable: `timeout "$SRVSTO" mount -t ceph :/ "$CEPHFS_STORE_DIR" -o ...`

### 2. Fragile Version Matching Logic (High)

The regex used to extract the target version SHA is overly permissive and fragile.

* **In `_power_roll_node_on_target`:**
```bash
target=$(jq -r '.version // ""' $ROLLING_JOB 2>/dev/null | grep -oE '[0-9a-f]{7,}' | tail -1)

```


Matching any hex string of 7 or more characters (`[0-9a-f]{7,}`) means if the version string contains other sequences that happen to be valid hex (e.g., `v3.1.10-rc2-0-gabcdef1`, or even just letters A-F in a branch name like `feature-decade`), it might accidentally latch onto the wrong substring.
*Fix:* Rely on a structured parsing approach rather than arbitrary regex matching. If `.version` is a JSON object with a specific SHA key, extract it with `jq`. If it's a known string format, parse it directly (e.g., using `cut` or a more rigidly anchored regex).

### 3. Non-Atomic State Updates (Medium)

While zeroing the deadline before setting the state to `running` shrinks the race condition window, it doesn't entirely eliminate it.

* **In `power_roll_decision`:**
```bash
_power_roll_set_raw deadline 0
_power_roll_set_str state running

```


Because these are likely two distinct operations (e.g., two sequential writes/updates to `$ROLLING_JOB`), there is still a micro-race where the file could be read by another process in an intermediate state.
*Fix:* If your underlying JSON/store update utility supports it, update both the `deadline` and `state` fields in a single, atomic operation (e.g., a single `jq` write passing both key-value pairs).

### 4. General Best Practices (Minor)

* **Variable Expansion in Logging:** In the pause reason `$(basename $pkg)`, you should quote the variable `$(basename "$pkg")` to prevent errors if the file path has special characters.
* **Error Handling on `jq`:** In `for h in $(jq -r '.nodes[].hostname' $ROLLING_JOB 2>/dev/null)`, if `$ROLLING_JOB` is unreadable or malformed, `jq` fails silently and the loop does not execute. Ensure there is a prior check guaranteeing the integrity of `$ROLLING_JOB` before executing this function, or handle the `jq` failure explicitly.

Eandalf-Bigstack:
## Code Review

_Reviewed the diff against the surrounding `sdk_power.sh` / `sdk_ceph.sh` code._

### Overview

Two independent rolling-upgrade fixes:

1. **cephfs staging wedge** — `ceph_mount_cephfs` now mounts with `recover_session=clean`; `_power_roll_stage_all` gains an all-or-nothing preflight that verifies every node can read the pkg off cephfs (self-healing once, else pausing cleanly pre-stage).
2. **`continue` edge cases** — zeroes a stale `deadline` before `state=running` (so a watchdog tick can't re-pause on the blown deadline), disarms the stale watchdog timer before re-arming, and adds `_power_roll_node_on_target` so a node that finished booting during the pause is marked **done** instead of being re-drained and rebooted.

Small, surgical, and the commit messages/comments explain *why*, not just *what*. 👍

### What's solid

- **The `deadline=0` ordering is race-correct.** `deadline` is set to `0` *before* `state=running`, and jq writes are atomic (temp + `mv`), so any watchdog tick that reads `state=running` is guaranteed to also read `deadline=0` and skip (`[ "$deadline" -gt 0 ] && ...`). The heartbeat branch also can't fire because `inflight` is cleared (`phase=""` → `*) return 0`). Closes the described window cleanly.
- **`disarm` before `arm`** correctly kills a stale timer that could otherwise fire on old state.
- **Preflight is genuinely all-or-nothing** and pauses with a precise, actionable reason before anything is staged.
- **Safe defaults**: `_power_roll_node_on_target` returns 1 (→ re-kick) on unreachable/mismatch.

### Issues & risks

#### 1. (High) `continue`'s "mark done" path skips node finalization

When `_power_roll_node_on_target` is true, `continue` writes `status=done` + `finished` directly and clears `inflight` — then `power_roll_advance` runs but skips its finalize block (`booted == inflight` is false, since `inflight` is now `""`).

But the only way to *reach* this branch is a node that booted onto the target **while the roll was paused**. `power_roll_advance` early-returns when `state != running`, so that node's own boot-path advance was a no-op — finalization never happened. The new done-path therefore permanently skips, for that node:

- **`power_node_resume`** — evacuation disabled `nova-compute` (`sdk_power.sh:1344`); finalize re-enables it (`:1386`). A compute node marked done this way stays **`--disable`d in nova** and won't host VMs, even though it's upgraded and healthy.
- **VM restore** from `CLUSTER_ACTIVE_RUNNING` — VMs suspended/stopped for the reboot are never restored.
- **Marker cleanup** — `rm -f $ROLLING_RECOVER_MARKER` and `.drained.$booted`. The recover marker is touched for **every** kicked node regardless of role (`:236`) and is local to the node, so `continue` on the operator box never clears it. Per the code's own comment, a leftover marker "would make every later boot look like [a roll]" — a latent bug that outlives this roll.
- `power_vm_reconcile_error_state` for that node.

**Suggestion:** route this node through the existing finalize logic — e.g. keep it `inflight` and call `power_roll_advance "$stuck"` so the established path restores VMs, resumes the service, reconciles, and clears markers. At minimum the done-branch should replicate the resume + marker-clear steps.

#### 2. (Medium) Self-heal can't recover an *already-mounted* wedged client

The preflight comment says it "self-heal[s] once via `ceph_mount_cephfs`", but `ceph_mount_cephfs` short-circuits at its top: `mountpoint -q $CEPHFS_STORE_DIR && return 0` (`sdk_ceph.sh:2800`). An MDS-evicted/blocklisted kernel client is *still mounted* — the mount stays in `/proc/mounts` and only ops return EACCES. So for the exact wedge this PR targets, `remote_run "$h" ceph_mount_cephfs` returns 0 without remounting, the recheck still fails, and preflight goes straight to pause.

`recover_session=clean` only affects mounts created *after* this lands, and it makes the kernel auto-reconnect — which is the actual fix. So the self-heal branch is effectively only useful for the "not mounted at all" case (e.g. the PPU cold-boot race). Still valuable, but consider either an unmount+remount in the heal step, or softening the comment to match reality.

#### 3. (Low) Firmware-target extraction is order/format-fragile

`target=$(jq -r '.version' ... | grep -oE '[0-9a-f]{7,}' | tail -1)` runs over a `.pkg` path. Decimal date stamps (e.g. `20260724`) are valid `[0-9a-f]{7,}` matches, so `tail -1` picks the wrong token if the commit isn't last in the filename — a false match marks a not-yet-upgraded node `done` and skips a needed reboot. Worth anchoring to the known pkg naming convention rather than "last hex-ish run".

#### 4. (Low) Style / conventions

- Preflight hardcodes `/mnt/cephfs`; the module already has `CEPHFS_STORE_DIR=/mnt/cephfs` (`sdk_01-var-static.sh:38`). Using the var keeps it consistent with `ceph_mount_cephfs`.
- `test -r $pkg` / `basename $pkg` are unquoted inside the `remote_run` strings — fine for `/var/update/*.pkg` but inconsistent with quoting elsewhere.

#### 5. (Low) Test coverage

The module has a unit-test pattern (`core/sdk_sh/tests/test_power_roll_*.sh`), but nothing covers the new `_power_roll_node_on_target`, the `continue` deadline-zeroing, or the stage preflight — exactly the kind of pure-logic branches (#1, #3) that pattern tests cheaply.

### Recommendation

The watchdog/deadline changes are correct and ready. I'd hold on the `continue` "mark done" path (finding #1) — leaving an upgraded compute node `nova`-disabled with unrestored VMs and a stale recover marker is a worse operational state than the re-reboot it replaces. #2–#3 are worth addressing or acknowledging before merge.

_🤖 Review generated with [Claude Code](https://claude.com/claude-code)_

Eandalf-Bigstack (approved):

PR #1162, opened 2026-07-24
feat(manila): expose one generic share backend per configured Cinder volume type

steven-chiu-bigstack:
#### What type of PR is this?

/kind feature

Lets Manila expose one additional generic share backend per Cinder volume type, so tenants can place file shares on different storage tiers the same way block volumes already can.

Manila mirrors Cinder's own existing multi-backend array (`cinder.storage.backend.%d.name`) directly -- the same way `config_glance.cpp` and `config_nova.cpp` already observe it for their own Cinder integration -- rather than introducing a separate Manila-specific config knob. Whatever additional Cinder backends an admin configures automatically become Manila share backends too; there's no independent Manila-side list to keep in sync.

Stays entirely within the existing **generic** driver (Nova service-VM + Cinder volume, Neutron-isolated) -- never a native protocol driver. This supersedes the native-driver approach (CephFS-native / NetApp toggles) built in the closed PR #1114, which was rejected on an architecture decision: native drivers let tenants reach the storage backend's own network directly, bypassing the isolation the generic driver already provides.

Landed in 4 commits, each independently reviewed:
- `3e191bd`, `c0ec03c` -- (superseded by `4a0dc9d`, see below) originally added a dedicated `manila.share.backend.%d.volume_type` tuning and generalized the previously-hardcoded single `[generic]` manila.conf backend section into a reusable per-backend writer; assembles `enabled_share_backends` as a loop over configured entries (mirrors Cinder's `SetStorageBackend()`).
- `6bd55a4` -- idempotent share-type create/prune (`os_manila_backend_types_init` in `sdk_os.sh`), wired to run on every `cubectl config commit` where the backend list changed. Deleting a config entry removes the `manila.conf` backend section immediately (blocking new share placement on it) but the Manila share-type object itself refuses deletion while any share still references it (mirrors Cinder's `cinder_is_volume_type_in_use` pattern) -- existing shares are never orphaned.
- `4a0dc9d` -- **revision from offline review**: removed the dedicated `manila.share.backend.%d.volume_type` tuning entirely; Manila now observes Cinder's own `CINDER_STORAGE_BACKEND` array on the existing "cinder" observer channel (group 4, already used for `CINDER_VOLUME_TYPE_DEFAULT`), matching the established Glance/Nova pattern. No other logic changed -- `SetShareBackends()`, `GetConfiguredShareBackendNames()`, and `os_manila_backend_types_init()` already treated the array as an opaque list of Cinder volume-type names.

The default backend's rendered output (section name `generic`, `share_backend_name=GENERIC`) remains byte-for-byte unchanged when no additional Cinder backends are configured.

Design spec and implementation plan (not committed to this branch, kept local per request) walked through the acceptance criteria, removal semantics, and the byte-for-byte-when-empty requirement in detail before implementation.

Fixes bigstack-oss/cubecmp#1053

- No automated test harness exists for hex config C++ modules in this repo (confirmed during design research -- only `config_etcd` has one), so verification is manual against a live dev cluster. **That manual verification (build via the jailed `make full`, then configure/create/mount/remove/prune against a real cluster) has not been run yet** -- this is opened as a draft pending that pass.
- Each commit went through an isolated task-scoped review (spec compliance + code quality) plus a final whole-branch review; all came back clean with no Critical/Important findings. A few Minor, explicitly-triaged edge cases were left as-is: no de-duplication of a repeated volume-type name in Cinder's config, case-sensitive reserved-name check (`generic`/`tenant_share_type`), and a `CommitCheck()` bootstrap-time gap that's an exact structural mirror of Cinder's own already-shipped behavior for its analogous flag.
- CMP-side UI/API/billing (share-type selection at create time, per-type pricing) is out of scope, tracked in sibling stories under epic bigstack-oss/cubecmp#907.

#### Additional documentation

```docs

```

### DIFF
--- a/core/modules/config_manila.cpp
+++ b/core/modules/config_manila.cpp
@@ -287,7 +287,6 @@ SetDefaults(Configs& config)
 {
     config["DEFAULT"]["api_paste_config"] = "/etc/manila/api-paste.ini";
     config["DEFAULT"]["default_share_type"] = "tenant_share_type";
-    config["DEFAULT"]["enabled_share_backends"] = "generic";
     config["DEFAULT"]["enabled_share_protocols"] = "NFS,CIFS";
     config["DEFAULT"]["rootwrap_config"] = "/etc/manila/rootwrap.conf";
     config["DEFAULT"]["share_name_template"] = "share-%s";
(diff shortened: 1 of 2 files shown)


core/modules/config_manila.cpp
```diff
@@ -52,6 +52,7 @@ CONFIG_TUNING_STR(MANILA_DBPASS, "manila.db.password", TUNING_UNPUB, "Set manila
 // public tunings
 CONFIG_TUNING_BOOL(MANILA_DEBUG, "manila.debug.enabled", TUNING_PUB, "Set to true to enable manila verbose log.", false);
 CONFIG_TUNING_STR(MANILA_VOLUME_TYPE, "manila.volume.type", TUNING_PUB, "Set manila backend volume type.", BUILTIN_VOLUME_TYPE, ValidateRegex, DFT_REGEX_STR);
+CONFIG_TUNING_STR(MANILA_SHARE_BACKEND, "manila.share.backend.%d.volume_type", TUNING_PUB, "Set an additional Cinder volume type Manila should serve as a share backend.", "", ValidateRegex, DFT_REGEX_STR);
```
traviswu-bigstack: How does CMP interact with the tuning? If mark it as a public tuning. there should be a change in cube-cos-api to cover which role it can land on (like control-only, control+compute, etc)

steven-chiu-bigstack:
Hi all.
I think this PR is ready for review (🤔?), thanks!

steven-chiu-bigstack:
I'll wait for personal CI pipeline to finish and verify that the implementation meets the AC. 🙏

traviswu-bigstack (approved):

PR #1169, opened 2026-07-25
fix(keycloak): gate the commit on the SAML metadata download

### LINKED ISSUE #1168: FTS: keycloak SAML metadata download is best-effort and never retried — a one-time failure permanently breaks httpd/keystone SSO, rancher and the ceph dashboard
## Description

Intermittently after FTS, httpd fails to start and takes keystone (and the UI login) down with:
```
httpd: AH00526: Syntax error on line 7 of /etc/httpd/conf.d/v3_mellon_keycloak_master.conf:
httpd: apr_stat: Error opening "/etc/keycloak/saml-metadata.xml" [2] "No such file or directory"
```
The same missing file also fails rancher's keycloak SAML auth (`rancher/main.tf:43 file(...)`) and the ceph dashboard SSO setup (`sdk_ceph.sh ceph_dashboard_idp_config`).

**Root cause — the metadata download is a best-effort step inside the deploying run only, with no retry path afterwards.**

`/etc/keycloak/saml-metadata.xml` is produced exactly once: the node that deploys keycloak (`isKeycloakUpdated=true`) downloads it at the end of its keycloak `Commit()`. Distribution to peers happens later via `syncSamlMetadata()` on the master's `cluster_start` trigger. Every consumer (`keystone_idp`, `rancher`, `ceph_dashboard_idp` — all `CONFIG_REQUIRES(keycloak)`) configures from that file. The pipeline has no second chance at any layer:

| Layer | Behavior when the file is missing |
|---|---|
| keycloak `Commit()` | download failure is log-and-continue — the module still **returns success**, so downstream consumers configure against a missing file |
| later commits / reboots | `checkKeycloak()` sees pods up → `isKeycloakUpdated=false` → the whole block including the download is **skipped** (observed: `keycloak commit(x) took 1.2 secs`, ×10 retries, file still missing) |
| `cluster_start` sync | `syncSamlMetadata()` **silently no-ops** when the master has no file (return value ignored) |
| health/repair | `check_keycloak` only checks pods/replicas — the file is not health-checked, so nothing repairs it |

So any one-time failure in the narrow window between keycloak deploy and the download becomes **permanent, cluster-wide** SSO breakage with no workaround (a reboot re-runs the same skips). The observed trigger on the 1cc was hex_config being killed 3 s after the keycloak terraform apply — the crash tracked separately as the timezone OOB bug — but any interruption or exhausted retry in that window has the same effect.

## How to Reproduce

1. FTS a node with `time.timezone=UTC` (hits the companion hex_config crash mid-keycloak-commit), or kill hex_config between the keycloak helm deploy and the metadata download.
2. `ls /etc/keycloak/saml-metadata.xml` → missing; keycloak pods are up.
3. Every subsequent commit logs `keycloak commit(x) took ~1.2 secs` and never regenerates the file.
4. `systemctl restart httpd` → failed, `AH00526`; rancher keycloak auth and ceph dashboard SSO also fail. Reboot → same result.

## Expected behavior

The keycloak module must not report success without the SAML metadata on disk. Since every downstream SSO consumer configures from this file, the commit should **wait until the download succeeds** (retry forever) and tell the operator on the console when it is stuck, instead of proceeding into a permanently broken SSO state. An operator can deliberately release the gate (accepting degraded SSO) by touching `/run/cube_keycloak_saml_gate_release`; the console message names the command.

## Additional Context

- The download's internal retries used unbounded curl (no `--connect-timeout`/`--max-time`); against a blackholed or hung endpoint a single call could block for hours, which also has to be fixed for any stuck-notification to fire on schedule.
- Peers are covered by the `cluster_start` scp-sync by design; this issue is about the producing run.
- rancher and the ceph dashboard consume the file **by value** at configure time, so guaranteeing the file before keycloak returns (DAG ordering) protects them with no per-component repair logic.

- [x] **Handbook knowledge update** — landed in [bigstack-handbook#170](https://github.com/bigstack-oss/bigstack-handbook/pull/170) (known-issue: missing keycloak SAML metadata permanently breaks httpd/keystone SSO — symptom chain, no-retry root cause, workaround, gate fix).

traviswu-bigstack:
#### What type of PR is this?

/kind bug

`/etc/keycloak/saml-metadata.xml` is consumed at configure time by `keystone_idp` (the mellon httpd conf), `rancher` and the ceph dashboard — all `CONFIG_REQUIRES(keycloak)`. Today the download is a best-effort step inside the deploying run: if that run is interrupted (e.g. the hex_config timezone crash, #1172) or the retries run out, `Commit()` still returns success, every consumer configures against a missing file (httpd `AH00526` → keystone + UI login down; rancher auth and ceph dashboard SSO fail), and no later commit ever retries (`checkKeycloak()` short-circuits to a ~1.2 s skip). The breakage is permanent.

This makes the download a **gate**: the deploying run does not return until the metadata is on disk.

- **Retry forever**: one download attempt per ~10 s round, unbounded, until it succeeds. By DAG ordering no SSO consumer can ever configure without the file.
- **Console notification every 600 s while stuck** (plus a journal Warning), so a hung FTS tells the operator what it is waiting on and where to look.
- **Time-bounded attempts**: new optional `connectTimeoutSecs`/`maxTimeSecs` on `HttpRequest` (default 0 = existing unbounded behavior for all other callers) let the keycloak requests use `--connect-timeout 10 --max-time 60`. Without this, one curl against a blackholed/hung endpoint could block for hours and starve the notification schedule; the previous inner 120-retry loop is folded into the outer gate (same unbounded retries, one bounded attempt per round).
- **Unreachable endpoint falls into the same gate** instead of bailing out with success: the cube-groups terraform is deferred until after the gate (a successful download implies the endpoint answers).
- **Operator release valve**: touching `/run/cube_keycloak_saml_gate_release` releases the gate deliberately (one-shot — the marker is consumed; `/run` is tmpfs so it cannot outlive a boot). The console message names the exact command, so the escape hatch is discoverable exactly when needed. Releasing means accepting degraded SSO until keycloak is repaired.
- **`CONFIG_REQUIRES(keystone_idp, keystone)`** (second commit): `SetupIdp` drives keystone's API (identity provider/mapping/group/role/protocol creation) but only declared keycloak — the scheduler could legally run it while keystone was still committing. Declare the real dependency instead of relying on timing.

Fixes #1168

- Gate semantics: the module **waits**, it does not fail — other independent modules keep committing on the other dataflow workers; the commit as a whole completes once the metadata exists. There is no state where keycloak "succeeds" without the file.
- Verified on a live 1cc:
  - fresh purge + bootstrap: rc=0 in 87 s, file present (happy path unchanged);
  - stuck simulation (file unproducible): gated 12+ min without proceeding, console notification at **603 s** after loop entry, released within **10 s** of the blocker clearing, commit completed (`keycloak commit(o) took 712.9 secs`);
  - executed curl journal-verified to carry `--connect-timeout 10 --max-time 60`;
  - release valve: gate engaged (blocked download) → `touch /run/cube_keycloak_saml_gate_release` → gate released within **5 s**, commit completed rc=0, marker consumed, journal Warning logged.
- Normal FTS keycloak commit is ~80–90 s end-to-end, so the 600 s threshold cannot false-alarm on a healthy install.
- Residual (by design): a file deleted *after* a successful deploy is re-distributed by the `cluster_start` sync on the next boot; this PR only guarantees the producing run. The companion crash fix (#1172) removes the observed cause of that run dying.
- Needs a clean 1cc + 3cc FTS with the built image as final validation (lab node was reclaimed mid-verification).

#### Additional documentation

```docs
Known-issue: missing /etc/keycloak/saml-metadata.xml after FTS permanently broke
httpd/keystone SSO, rancher auth and ceph dashboard SSO with no retry path.
The keycloak commit now gates on the metadata download (retry forever,
console notification every 600s while stuck).
```

### DIFF
--- a/core/cube_sdk_library/src/http.cpp
+++ b/core/cube_sdk_library/src/http.cpp
@@ -130,16 +130,26 @@ DoHttp(const HttpRequest& req)
         bodyString = "-d '" + req.body + "'";
     }
 
+    // form the time-bound string; empty keeps curl's default (unbounded)
+    std::stringstream timeoutString;
+    if (req.connectTimeoutSecs > 0) {
+        timeoutString << "--connect-timeout " << req.connectTimeoutSecs << " ";
+    }
+    if (req.maxTimeSecs > 0) {
+        timeoutString << "--max-time " << req.maxTimeSecs << " ";
+    }
+
     // send the request via curl
     bool isCurlSuccessful = false;
     if (req.url.scheme == "http") {
         isCurlSuccessful = HexUtilSystemF(
                                0,
                                0,
-                               "curl --silent --show-error "
+                               "curl --silent --show-error %s"
                                "--output %s --write-out \"%%{http_code}\" "
                                "--request %s \"%s\" "
                                "%s %s >%s 2>%s",
+                               timeoutString.str().c_str(),
                                outputFile.fileName.c_str(),
                                req.method.c_str(),
                                req.url.String().c_str(),
@@ -152,10 +162,11 @@ DoHttp(const HttpRequest& req)
         isCurlSuccessful = HexUtilSystemF(
                                0,
                                0,
-                               "curl --insecure --silent --show-error "
+                               "curl --insecure --silent --show-error %s"
                                "--output %s --write-out \"%%{http_code}\" "
                                "--request %s \"%s\" "
                                "%s %s >%s 2>%s",
+                               timeoutString.str().c_str(),
                                outputFile.fileName.c_str(),
                                req.method.c_str(),
                                req.url.String().c_str(),
--- a/core/cube_sdk_library/src/http.hpp
+++ b/core/cube_sdk_library/src/http.hpp
@@ -14,6 +14,11 @@ struct HttpRequest {
     Url url;
     std::map<std::string, std::string> header;
     std::string body;
+    /**
+     * curl time bounds in seconds; 0 keeps curl's default (unbounded).
+     */
+    int connectTimeoutSecs = 0;
+    int maxTimeSecs = 0;
 };
 
 struct HttpResponse {
--- a/core/modules/config_keycloak.cpp
+++ b/core/modules/config_keycloak.cpp
@@ -12,6 +12,10 @@ static const char KEYCLOAK_SAML_METADATA_FILE[] = "/etc/keycloak/saml-metadata.x
 static const std::string KEYCLOAK_ADMIN_PASSWORD_K8S_SECRET = "admin-password";
 static const std::string KEYCLOAK_ADMIN_PASSWORD_TERRAFORM_VARIABLE_FILE
     = "/etc/cube/cos/terraform/values/keycloak-admin-password.tfvars";
+static const int SAML_METADATA_STUCK_NOTIFY_SECS = 600;
+// operator escape hatch: touching this releases the saml metadata gate below
+// (one-shot; consumed when honored). /run is tmpfs so it cannot outlive a boot.
+static const char SAML_METADATA_GATE_RELEASE[] = "/run/cube_keycloak_saml_gate_release";
 
 // external global variables
 CONFIG_GLOBAL_STR_REF(MGMT_ADDR);
@@ -314,6 +318,8 @@ checkKeycloakEndpoint(const std::string& endpointIp)
     for (int i = 0; i < 15; i++) {
         const HttpRequest req = {
             .url = url,
+            .connectTimeoutSecs = 10,
+            .maxTimeSecs = 60,
         };
         const HttpResponse res = GetHttp(req);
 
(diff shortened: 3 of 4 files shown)


traviswu-bigstack:
@Eandalf-Bigstack Following up on our discussion this morning — I'd still prefer to keep the waiting gate. The SAML metadata is a hard requirement for every SSO consumer downstream (keystone_idp/mellon httpd conf, rancher, ceph dashboard): proceeding without it doesn't fail fast, it configures the whole chain wrong, and that state never self-heals (later commits skip the download via `checkKeycloak`). Waiting is the only behavior that can't produce a misconfigured cluster.

To keep the wait operable rather than a silent hang, the gate notifies the console every 600s, and an operator can deliberately release it with `touch /run/cube_keycloak_saml_gate_release` (one-shot, boot-scoped) if keycloak is genuinely broken and degraded SSO is acceptable.

On top of that, this PR now also declares the missing module dependency — `CONFIG_REQUIRES(keystone_idp, keystone)` that introduced DAG commit order really breaks it — since `SetupIdp` drives keystone's API and the ordering shouldn't rely on timing luck. The PR is ready for review now

Eandalf-Bigstack (approved):

PR #1173, opened 2026-07-26
fix(horizon): escape the timezone without assuming two components

### LINKED ISSUE #1172: FTS: hex_config SIGSEGV in config_horizon on slashless timezones (e.g. UTC) — policy apply aborts, role stays undef
## Description

Intermittently during FTS, `hex_config` crashes with **SIGSEGV** while committing modules, which aborts the whole policy apply (`Failed to apply policy changes` / `Error applying snapshot for policy component`). The pending policies (`cubesys`, `network`, `time`, …) are dumped to `/etc/failed_policies/`, the node's role stays `undef`, and **FTS never completes**. It is rare (~1 in 50 installs) and has no workaround — a reboot re-runs the same commit and can crash again.

**Root cause — out-of-bounds `std::vector` access in `config_horizon.cpp`.** To escape the `/` in the timezone for the Django settings file, `Commit()` splits the timezone on `/` and rejoins `tzv[0] + "\/" + tzv[1]`, assuming exactly two components:

```cpp
// core/modules/config_horizon.cpp:179-180
std::vector<std::string> tzv = hex_string_util::split(s_timezone.newValue(), '/');
std::string tz = tzv[0] + "\\/" + tzv[1];
```

When the timezone has **no `/`** — e.g. `UTC` — `split` returns a **1-element** vector, so `tzv[1]` reads **past the end of the vector**. That yields an uninitialized `std::string` with a garbage data pointer and a bogus length; the subsequent `std::string` concatenation `memmove`s from that bad pointer and segfaults. Whether it actually faults depends on heap layout, which is why it is intermittent (and the 16-way parallel commit scheduler, churning the heap concurrently, makes it surface more readily).

Confirmed from a core dump (unstripped binary, matching build-id):
```
Thread (crashing):
  __memmove_avx_unaligned_erms_rtm ()
  std::char_traits<char>::copy (__n=149, __s2=0x747200312600 <error: Cannot access memory>)
  std::string::_M_append / std::operator+
  Commit () at core/modules/config_horizon.cpp:180
  CommitModulesDataflow(...)::worker-lambda   (core parallel commit scheduler)

Frame locals:
  s_timezone.m_new = "UTC"
  tzv = std::vector of length 1 = {"UTC"}      <-- tzv[1] is out of bounds
```

The same code also silently mishandles 3-component zones (e.g. `America/Argentina/Buenos_Aires`) — it drops the third part.

**Exposure:** any cluster whose timezone has no `/` — `UTC` (the common one), `GMT`, `Singapore`, `Hongkong`, `Egypt`, `Turkey`, etc. Two-component zones like `America/New_York` are safe.

## How to Reproduce

1. Configure a node with a slashless timezone, e.g. `time.timezone=UTC`.
2. Run FTS (or `hex_config apply_snapshot <snapshot>` / a full commit) repeatedly.
3. Intermittently `hex_config` segfaults during the commit; `journalctl` shows `hex_crashd … CRASH … program=hex_config … signal=11`, followed by `Failed to apply policy changes`.
4. `cluster` role stays `undef`; `/etc/failed_policies/{cubesys,network,time}` populated; FTS does not complete.

## Expected behavior

`config_horizon` escapes the timezone's `/` for the Django config without ever indexing past the split result, for any timezone (0, 1, or N `/` components). FTS completes on clusters using `UTC` (or any slashless zone).

## Additional Context

- Fix: escape every `/` in place instead of assuming two components:
  ```cpp
  std::string tz = s_timezone.newValue();
  for (size_t pos = 0; (pos = tz.find('/', pos)) != std::string::npos; pos += 2)
      tz.replace(pos, 1, "\\/");
  ```
- This is distinct from the keycloak SAML-metadata FTS bug (#1168) — both surface as "FTS breaks services," but this one is the dominant, timezone-dependent crash.
- Verified on a live 1cc (`timezone=UTC`): the unpatched binary cored within 1–2 `apply_snapshot` runs (3 crashes captured, incl. a core dump symbolicated to this exact frame); the patched binary ran **15/15** consecutive `apply_snapshot` runs with **zero** crashes, and `role` recovered from `undef` → `control-converged`.

- [x] **Handbook knowledge update** — record the crash signature (hex_config SIGSEGV during FTS → `config_horizon:180` OOB on slashless timezone) and the `failed_policies`/`role=undef` symptom chain in the CubeCOS known-issues, so future intermittent-FTS reports map straight to it.

traviswu-bigstack:
#### What type of PR is this?

/kind bug

`config_horizon`'s `Commit()` escapes the `/` in the timezone for the Django settings file by splitting on `/` and rejoining `tzv[0] + "\/" + tzv[1]`, assuming the timezone always has **exactly two** components:

```cpp
std::vector<std::string> tzv = hex_string_util::split(s_timezone.newValue(), '/');
std::string tz = tzv[0] + "\\/" + tzv[1];
```

For a **slashless** timezone such as `UTC`, `split` returns a 1-element vector, so `tzv[1]` reads **out of bounds** — an uninitialized `std::string` whose garbage data pointer makes the following concatenation `memmove` from bad memory and **SIGSEGV**. Because it depends on heap layout it is intermittent (~1/50), and the 16-worker parallel commit scheduler churning the heap makes it surface. When it hits during FTS, hex_config dies mid-commit → `Failed to apply policy changes` → `cubesys/network/time` land in `failed_policies` → role stays `undef` → **FTS never completes** (confirmed by core dump; crashing frame is `config_horizon.cpp:180` with `s_timezone="UTC"`, `tzv={"UTC"}`).

This escapes **every** `/` in place, with no indexing past the split — correct for 0, 1, or N components (also fixes the latent bug where 3-part zones like `America/Argentina/Buenos_Aires` silently dropped their third part):

```cpp
std::string tz = s_timezone.newValue();
for (size_t pos = 0; (pos = tz.find('/', pos)) != std::string::npos; pos += 2)
    tz.replace(pos, 1, "\\/");
```

Fixes #1172

- Single-line-of-logic change in `core/modules/config_horizon.cpp`; the produced `tz` is identical to before for two-component zones (`America/New_York` → `America\/New_York`).
- Root-caused from a core dump on a live 1cc (`timezone=UTC`): unpatched binary cored within 1–2 `apply_snapshot` runs (core symbolicated to `config_horizon.cpp:180`, `s_timezone="UTC"`, `tzv={"UTC"}`); patched binary ran **15/15** consecutive runs with **zero** crashes and `role` recovered `undef` → `control-converged`.
- Distinct from #1168 (keycloak SAML metadata). Both present as "FTS breaks services"; this is the dominant, timezone-dependent crash.

#### Additional documentation

```docs
Known-issue: intermittent hex_config SIGSEGV during FTS on clusters using a
slashless timezone (UTC/GMT/Singapore/…). Root cause: out-of-bounds tzv[1] in
config_horizon when escaping the timezone '/'. Symptom chain: crash →
"Failed to apply policy changes" → failed_policies/{cubesys,network,time} →
role=undef → FTS never completes.
```

### DIFF
--- a/core/modules/config_horizon.cpp
+++ b/core/modules/config_horizon.cpp
@@ -176,8 +176,10 @@ Commit(bool modified, int dryLevel)
     if (s_bConfigChanged) {
         std::string sharedId = G(SHARED_ID);
         std::string cachesrvs = "'" + sharedId + ":11211'";
-        std::vector<std::string> tzv = hex_string_util::split(s_timezone.newValue(), '/');
-        std::string tz = tzv[0] + "\\/" + tzv[1];
+        // escape every '/' as '\/' for the django conf; safe for slashless zones (e.g. UTC)
+        std::string tz = s_timezone.newValue();
+        for (size_t pos = 0; (pos = tz.find('/', pos)) != std::string::npos; pos += 2)
+            tz.replace(pos, 1, "\\/");
         WriteDjangoConf(sharedId.c_str(), cachesrvs.c_str(), dbPass.c_str(), tz.c_str());
     }
 


Eandalf-Bigstack:
# Code Review: fix(horizon): escape the timezone without assuming two components

## Summary
✅ **Approved** — This fix correctly addresses a critical out-of-bounds bug that causes intermittent SIGSEGV during FTS operations on systems using slashless timezones.

---

## Analysis

### The Bug (Original Code)
```cpp
std::vector<std::string> tzv = hex_string_util::split(s_timezone.newValue(), '/');
std::string tz = tzv[0] + "\\/" + tzv[1];  // UB when tzw has <2 elements
```
Accessing `tzv[1]` for single-element timezones (UTC, GMT, Singapore, etc.) reads uninitialized memory → garbage pointer → SIGSEGV on `memmove` during concatenation.

### The Fix
```cpp
std::string tz = s_timezone.newValue();
for (size_t pos = 0; (pos = tz.find('/', pos)) != std::string::npos; pos += 2)
    tz.replace(pos, 1, "\\/");
```

**Strengths:**
- ✅ Handles all timezone formats: 0 components (UTC), 1 component (slashless), 2+ components (America/New_York, America/Argentina/Buenos_Aires)
- ✅ `pos += 2` correctly skips over the newly-inserted `\` to avoid re-matching the same `/`
- ✅ No vector indexing — eliminates the bounds check entirely
- ✅ Fixes latent bug: 3-part zones no longer silently drop components
- ✅ Good explanatory comment
- ✅ Output identical to original for 2-part zones (no behavioral regression)

**Correctness Verification:**
- `"UTC"` → `"UTC"` (no slashes, loop skipped) ✓
- `"America/New_York"` → `"America\/New_York"` ✓  
- `"America/Argentina/Buenos_Aires"` → `"America\/Argentina\/Buenos_Aires"` ✓

### Evidence Quality
The PR provides strong evidence:
- Core dump symbolicated to exact crash site (`config_horizon.cpp:180`, `s_timezone="UTC"`)
- Patched binary: 15/15 consecutive runs zero crashes
- Fixes the `role=undef` → FTS hang symptom chain

---

## Minor Notes
- The loop assumes `std::string::replace()` invalidates future searches correctly — standard library behavior is sound here
- No unit tests referenced, but this is a single-line logic change in a critical path that appears under-tested upstream (hence the intermittent crash)

---

## Recommendation
**Merge.** This is a minimal, targeted fix for a production-critical crash with clear root-cause analysis and validation.

Eandalf-Bigstack:
### Code Review

**Status:** ✅ Approved for the immediate fix, but with a **critical security/architectural warning** regarding downstream handling.

---

### 1. Code Analysis (The Fix)

The PR successfully identifies and fixes the root cause of the SIGSEGV.

* **The Bug:** The previous implementation assumed exactly two components when using `hex_string_util::split()`. For slashless timezones like `UTC`, `tzv[1]` accessed uninitialized memory, leading to Undefined Behavior and intermittent crashes.
* **The Fix:** The new implementation safely handles 0, 1, and N-component timezones.
* **Logic Verification:** The `for` loop logic is robust. Replacing `/` with `\/` increases the string length by 1, pushing the newly inserted `/` to index `pos + 1`. The `pos += 2` increment perfectly skips the newly inserted characters, preventing any infinite loops.
* **Positive Side Effect:** As noted, this correctly preserves 3-part IANA timezones (like `America/Argentina/Buenos_Aires`) which were being silently truncated by the old `tzv[0] + "\\/" + tzv[1]` logic.

### 2. Architectural / Security Concern ⚠️

While this C++ logic perfectly implements the escaping, **why are we escaping slashes for a Django config?**

Standard Python/Django `.py` files do not require slashes to be escaped (e.g., `TIME_ZONE = 'America/New_York'` is perfectly valid Python). The requirement to escape `/` as `\/` strongly suggests that `WriteDjangoConf()` under the hood is shelling out to a utility like `sed` using `/` as a delimiter (e.g., `sed 's/TIME_ZONE=.*/TIME_ZONE="$TZ"/g'`).

If `WriteDjangoConf` is indeed building shell commands, you have a much larger architectural issue:

* **Shell/Command Injection:** If `s_timezone.newValue()` is user-controllable and not strictly validated upstream, an attacker could supply shell metacharacters (e.g., `UTC'; rm -rf /; '`) to achieve Remote Code Execution.
* **Sed Injection:** Even without breaking the shell quote, an attacker could inject `sed` metacharacters like `&` to corrupt the config file.

### Recommendations

1. **Merge this PR:** It cleanly fixes a production-crashing bug and restores FTS stability.
2. **Follow-up Action:** Please audit `WriteDjangoConf` and how `s_timezone.newValue()` is validated upstream.
* Ensure the timezone input is strictly validated against an allowlist of valid IANA timezones.
* Long-term, refactor `WriteDjangoConf` to write/modify the file using native C++ file I/O rather than shelling out to system utilities, eliminating entire classes of injection vulnerabilities.

Eandalf-Bigstack (approved):

PR #1175, opened 2026-07-27
fix(gpu): make sriovVgpu carve and pgpu passthrough actually work

SekiXu:
#### What type of PR is this?

/kind bug

Three bugs that, together, made GPU resource-type management unusable on a
node with real hardware. They are one causal chain, which is why they ship
together — each commit is self-contained and can be reviewed on its own.

---

**Commit 1 — `sriovVgpu` carve failed 100% of the time.**

`RunSriovManageWithRetry` (core/modules/config_gpu.cpp) captured
`sriov-manage`'s exit code with an **unquoted** marker:

```c
"%s %s %s 2>&1; echo |__HEX_RC_$?__|"
```

The shell parses those `|` as pipes, so `echo |__HEX_RC_$?__|` is a syntax
error (a trailing pipe with no following command). `/bin/sh -c` aborts the
**entire** command line on that syntax error, so `sriov-manage` **never
runs**. The marker never appears, `marker == npos` → `rc = -1` → treated as
failure, and the call logs an empty message:

```
Error: gpu_resource_set: sriov-manage -e 0001:c8:00.0 failed:
```

Regressed in `44011f4f` ("use a more unique exit-code delimiter"), which
wrapped the delimiter in `|...|` without quoting it. Quoting it makes `|` a
literal `echo` argument; `$?` still expands inside double quotes, and the
existing `rfind` / `atoi(... + 10)` parsing is unchanged (`"|__HEX_RC_"` is
exactly 10 characters).

---

**Commit 2 — `pgpu` left the card bound to no driver, and could never be undone.**

vfio-pci carries no id-table entry matching an NVIDIA display device, so
`echo <addr> > /sys/bus/pci/drivers/vfio-pci/bind` is refused by the kernel.
The unbind that runs immediately before it *does* succeed, so the card ended
up bound to **no driver at all** — invisible to nvidia-smi, unusable by any
VM, recoverable only by running `echo <addr> > /sys/bus/pci/drivers_probe` by
hand. Both sysfs writes discarded stderr, so the caller saw a bare exit 1
with no explanation. Confirmed on cn13: the device's `driver_override` read
`(null)` and `/sys/bus/pci/drivers/vfio-pci/` held **zero** device ids.

The fix sets `driver_override` before unbinding and binds via
`drivers_probe`. `driver_override` tells the PCI core which driver the device
must use, bypassing the id-table match that was rejecting the bind. If the
bind still fails, the override is cleared and the device re-probed so it
returns to its previous driver instead of being stranded, and every failure
path explains itself on stderr. Re-binding a device already on vfio-pci
returns early, so `Commit()` re-apply does not cycle a live passthrough card.

`gpu_unset_current_type` also gained the pgpu branch it never had, calling
the new `gpu_unbind_vfio_pci`. `driver_override` is sticky: without clearing
it the device re-binds to vfio-pci on every probe and can never return to the
nvidia driver. Its PCI address is read from config.json, not nvidia-smi,
because a vfio-bound card is invisible to nvidia-smi.

`ResourceSetMain` releases a pgpu from vfio-pci **before** running
`gpu_resource_set_check` and `ValidateVgpuProfiles`, and re-binds it if
either fails. Both read the card's supported vGPU types through nvidia-smi,
which cannot see a vfio-bound device — so on a pgpu they rejected every
profile (`profile id 1518 is not a valid sriovVgpu profile`) before the
teardown step could hand the card back. The `gpu_unset_current_type` failure
path re-binds too, so no early exit leaves the hardware disagreeing with what
config.json records.

---

**Commit 3 — `gpu_device_list` silently reported an empty device list.**

The loop declared `local profile_count_limit="null"` while the sriovVgpu
branch assigned — and the jq call read — `sriov_vgpu_profile_count_limit`. On
any other type that second name was never set, so `--argjson
sriovVgpuProfileCountLimit ""` failed with `jq: invalid JSON text passed to
--argjson`. Since the loop rebuilds `$output` by piping it through jq, that
one failure collapses `$output` to an empty string and every device
accumulated so far is lost — yet the function still exited 0 and printed
nothing, which callers cannot distinguish from "this node has no GPUs"
(`gpu_resource_set_check` reports the GPU as not found; cube-cos-api drops
every card on the node). Being non-local also let one card's value leak into
the next.

Triggering it needs a card recorded as pgpu that nvidia-smi can still see —
exactly the state a failed pgpu bind used to leave behind, and the state
commit 2 now passes through deliberately when releasing a pgpu for
validation. Commit 2 without this one would trade a broken bind for a broken
listing.

The listing now also refuses to print anything that is not a JSON array,
rather than passing an empty one off as success.

Fixes # 

None

#### Additional documentation

```docs
Hardware verification (cn13, NVIDIA RTX PRO 6000 Blackwell, 2026-07-25/27).

Commit 1:
- Reproduced: the exact emitted string
  `/usr/lib/nvidia/sriov-manage -e 0001:c8:00.0 2>&1; echo |__HEX_RC_$?__|`
  run in a shell → "syntax error: unexpected end of file"; numvfs stays 0.
  The pre-fix binary failed the carve 100% of the time.
- With the fix: carve → exit 0, numvfs=48, VF 0001:c8:00.2
  current_vgpu_type=1518, and a Nova VM booted ACTIVE against the carved VF
  (nvidia-smi vgpu -q: Active vGPUs 1).
- The unbindLock retry this helper exists for had never executed since
  44011f4f (rc always -1, so the "Cannot obtain unbindLock" branch was
  unreachable). Re-verified by holding the device open in a background loop
  while carving: "unbindLock busy, retrying (1/5)...(4/5)" fires, i.e. rc
  capture and contention detection both work again.

Commit 2 — diagnosis was done step by step rather than through hex_config,
which swallowed the detail:
- `hex_sdk gpu_unset_current_type <uuid>` alone → exit 0, numvfs 0, PF still
  on nvidia. Rules out the teardown step.
- Pre-state: driver_override = (null); vfio-pci has 0 device ids.
- `hex_sdk gpu_bind_vfio_pci <addr>` alone → exit 1, PF left with no driver.
  Reproduces the reported symptom exactly.
- After the fix, a full cycle, every step exit 0:
      sriovVgpu -> pgpu -> sriovVgpu -> pgpu -> sriovVgpu
  Each step checked against three sources that must agree: config.json's
  recorded type, sysfs sriov_numvfs, and the driver symlink.
    into pgpu:   driver = vfio-pci, driver_override = vfio-pci, numvfs = 0
    out of pgpu: driver = nvidia, driver_override = (null), numvfs = 48,
                 VF 0001:c8:00.2 current_vgpu_type = 1518
  Before this change the very first step exited 1 and left the PF driverless.
- Failure handling exercised on purpose by holding the device open during a
  teardown: the call failed after 5 unbindLock retries and the card was left
  fully consistent — config.json unchanged, numvfs 48, driver nvidia,
  driver_override empty.

Commit 3:
- With GPU-0a5ba9ad recorded as pgpu and bound to nvidia, gpu_device_list
  returned nothing (exit 0) before; lists all four cards after.

End to end with all three, plus the cube-cos-api side: /gpuCards reports
degraded false, profile 1518 count 1 with its alias, attachedInstances[0].id
== the VM UUID, per-instance 128 MiB == host FB Used, device 4240 ==
Reserved 2288 + Used 1952.

NOT covered, stated so this is not over-read:
- Boot-time Commit() re-apply of a pgpu card. After a reboot driver_override
  is gone and nvidia has claimed the card, so Commit() takes the full
  set-override/unbind/probe path rather than the early return. A pgpu card
  that fails to re-bind at boot would not be available to its VM. Needs a
  node that can actually be rebooted.
- cube-cos-api's /gpuCards was never called while a card was in pgpu state,
  so the API-side pgpu reporting path is unverified end to end.
- TeardownSriov's `sriov-manage -d` is only reachable from the apply-rollback
  at config_gpu.cpp:158; that specific path was not contrived.
```

### DIFF
--- a/core/sdk_sh/modules/sdk_gpu.sh
+++ b/core/sdk_sh/modules/sdk_gpu.sh
@@ -768,6 +768,28 @@ gpu_unset_current_type()
             echo "Error: failed to disable MIG mode for GPU $gpu_id" >&2
             exit 1
         fi
+    elif [ "$current_type" = "pgpu" ]; then
+        # Hand the card back from vfio-pci to the nvidia driver. Without this
+        # the card stays bound to vfio-pci (and keeps its sticky
+        # driver_override), so switching it to sriovVgpu/migBackedVgpu next
+        # fails: nvidia-smi cannot see a vfio-bound device.
+        #
+        # The address comes from config.json, not nvidia-smi: a pgpu is bound
+        # to vfio-pci and therefore invisible to nvidia-smi, so the
+        # --query-gpu lookup used for sriovVgpu above would return nothing here.
+        local pgpu_pci_address
+        pgpu_pci_address=$(jq -r --arg id "$gpu_id" \
+            'map(select(.id == $id)) | if length > 0 then (.[0].pciAddress // "") else "" end' \
+            "$GPU_CONFIG_FILE_PATH" 2>/dev/null)
+
+        if [ -z "$pgpu_pci_address" ]; then
+            echo "Error: GPU $gpu_id has type pgpu but no recorded pciAddress; cannot release it from vfio-pci" >&2
+            exit 1
+        fi
+
+        if ! gpu_unbind_vfio_pci "$pgpu_pci_address"; then
+            exit 1
+        fi
     fi
 }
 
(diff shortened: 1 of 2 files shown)


Eandalf-Bigstack:
# Review — critical issues & suggestions

- Good: quoting the exit-marker fixes the sriov-manage hang; shell parsing bug properly addressed.  
- Verify pciAddress is initialized before use in ResourceSetMain (pciAddress.c_str() is used early). Potential build/run crash if not set. (High)

- Releasing/re-binding vfio devices can race with other users/probes. Ensure unbindLock/serialization covers the new early unbind/rebind paths and add retries/timeouts. (High)

- gpu_bind_vfio_pci/gpu_unbind_vfio_pci: sysfs writes may fail silently on permission/selinux; current error paths are helpful but consider stronger retry/logging and check return of redirections on all shells. (Medium)

- gpu_device_list: jq safety improved. Also verify GPU_CONFIG_FILE_PATH exists/readable before jq to avoid confusing errors. (Low)

- Tests: add unit/integration tests exercising: sriov carve success, pgpu bind/unbind failure paths, and gpu_device_list malformed-jq handling. (Medium)

- Docs/rollback: mention manual recovery steps in logs when device left driverless and ensure messages direct operators to drivers_probe command. (Low)

Overall changes look correct functionally, but please confirm pciAddress initialization and add concurrency/robustness tests.

Eandalf-Bigstack:
## Code Review

### 1. Critical Logic Flaw: Uninitialized Variable Use

**File:** `core/modules/config_gpu.cpp` (Lines 629-631)

In `ResourceSetMain`, the variable `pciAddress` is referenced in the newly added early teardown path:

```cpp
if (releasedFromVfio &&
    HexUtilSystemF(0, 0, HEX_SDK " gpu_unbind_vfio_pci %s", pciAddress.c_str()) != 0) {

```

The diff shows that `FindGpuDevice(gpuId)` is called earlier (Line 607), but it does not show `pciAddress` being extracted from the `device` JSON object before this command executes. If `pciAddress` was originally declared further down in the function (which is likely, given the pre-condition check removal), this will cause a compilation error. If it is declared but uninitialized, it will pass an empty string to the shell script and fail silently or target the wrong device.

**Recommendation:** Extract and validate `pciAddress` immediately after confirming the `device` object is valid:

```cpp
const std::string pciAddress = device["pciAddress"].is_string() ? device["pciAddress"].string_value() : "";
if (pciAddress.empty()) {
    HexLogError("gpu_resource_set: GPU %s missing pciAddress", gpuId);
    return EXIT_FAILURE;
}

```

### 2. Concurrency Risk: Missing Sysfs Synchronization

**File:** `core/sdk_sh/modules/sdk_gpu.sh`

The new `gpu_bind_vfio_pci` and `gpu_unbind_vfio_pci` functions manipulate `/sys/bus/pci/devices/.../driver_override`, `unbind`, and `drivers_probe`. The PR description mentions an `unbindLock` for `sriov-manage`, but these new pure-sysfs shell functions do not appear to implement any locking.

If two processes attempt to bind or unbind GPUs simultaneously, they could race on `drivers_probe` or conflict on kernel-level PCI locks, potentially leaving the PCI bus in an inconsistent state or stranding devices.

**Recommendation:** Wrap the critical sysfs mutations inside a file lock (e.g., `flock`), mirroring the `unbindLock` mechanism used elsewhere in the GPU management stack.

### 3. Error Handling and Defensive Checks

**File:** `core/sdk_sh/modules/sdk_gpu.sh`

* **Missing Config File Check:** In `gpu_unset_current_type`, `jq` is used to parse `$GPU_CONFIG_FILE_PATH`. While `2>/dev/null` and the `-z` check will safely catch the failure, explicitly verifying `[ -r "$GPU_CONFIG_FILE_PATH" ]` before executing `jq` would allow for a much more descriptive error message ("Config file unreadable" vs. "GPU has no recorded pciAddress").
* **Restoration Logic:** In `gpu_bind_vfio_pci`, if binding to `vfio-pci` fails, the script attempts to restore the `previous_driver`. Ensure that `previous_driver` is strictly validated. If a driver name contains spaces or special characters (unlikely but possible), the unquoted sysfs writes could fail. Always quote variables in paths: `/sys/bus/pci/drivers/${previous_driver}/unbind`.

### 4. Positive Architectural Notes

* **Shell Syntax Fix:** Wrapping the marker in quotes (`"|__HEX_RC_$?__|"`) correctly resolves the pipeline parsing bug in `/bin/sh`.
* **Driver Override Implementation:** Using `driver_override` to bypass the `vfio-pci` ID table mismatch is the correct, kernel-native approach for dynamic passthrough. The fallback restoration logic is a strong defensive addition.
* **JQ Safety:** Initializing `sriov_vgpu_profile_count_limit="null"` properly scopes the variable and protects the JSON pipeline from collapsing due to unassigned bash variables.

Eandalf-Bigstack (approved):
We could ignore most of the AI-assisted code review from GitHub Copilot and Gemini Pro above. However, there is a legitimate concern, the race condition on device driver operations, including binding and unbinding.

2. Concurrency Risk: Missing Sysfs Synchronization
core/sdk_sh/modules/sdk_gpu.sh

The new gpu_bind_vfio_pci and gpu_unbind_vfio_pci functions manipulate /sys/bus/pci/devices/.../driver_override, unbind, and drivers_probe. The PR description mentions an unbindLock for sriov-manage, but these new pure-sysfs shell functions do not appear to implement any locking.

If two processes attempt to bind or unbind GPUs simultaneously, they could race on drivers_probe or conflict on kernel-level PCI locks, potentially leaving the PCI bus in an inconsistent state or stranding devices.

Recommendation: Wrap the critical sysfs mutations inside a file lock (e.g., flock), mirroring the unbindLock mechanism used elsewhere in the GPU management stack.

PR #1181, opened 2026-07-28
Add an eBPF toolchain to the build jail and bump Go to 1.25.12

arasHi87:
#### What type of PR is this?

/kind feature

Adds an eBPF compilation toolchain to the build jail, and bumps Go.

The network-telemetry agent (`bigstack-oss/lachesis`) is being onboarded as a CubeCOS component, and it is the first component in the tree that compiles eBPF C. A grep of cubecos returns zero hits for `clang`, `llvm`, `libbpf` or `bpftool` — the capability is simply absent, and `gcc` cannot substitute because it has no BPF target. Without this the component cannot be built by the pipeline at all.

Two changes, both in `jail/jail.ubi9.dockerfile`:

- **New `HEX_BPF` package group** — `clang llvm libbpf-devel kernel-headers bpftool` — added to both the `weak_dep_0` and `weak_dep_1` install lines, following the existing `HEX_BE` / `HEX_SDK` / `HEX_EXTRA` / `HEX_TEST` grouping. `libbpf-devel` comes from CRB, which tier1 already enables.
- **`GOLANG_VER` 1.24.2 → 1.25.12.** lachesis declares `go 1.25.3`, so 1.24.2 cannot build it. Pinned to the latest 1.25 patch rather than the bare minimum so the jail is not a release behind on toolchain fixes.

`kernel-headers` is declared explicitly even though `glibc-devel` already requires it transitively (`kernel-headers >= 3.2`): BPF builds reference `/usr/include/{linux,asm,asm-generic}` by path, and a build input that is only present by transitive luck is a latent breakage.

`bpftool` is diagnostic only. The BPF sources use UAPI types and take no CO-RE relocation, so nothing in the build needs BTF — worth stating so nobody later adds a `bpftool btf dump` step assuming it is required.

Fixes bigstack-oss/lachesis#279

> **The Go bump is the regression risk here, not the new packages.** It recompiles every Go component in the tree. Verified by building `core/cubectl` (pure Go) inside the freshly built jail under 1.25.12 — clean, 2m55s.

Verified end to end on a CentOS Stream 9 host by running the real target, not a replica:

- `make centos9-jail` completes and the container comes **up** — `Up`, PID 1 is `systemd`
- inside the running jail: `clang 22.1.3`, `go1.25.12`, `bpftool v7.5.0`, `libbpf-devel-1.5.0-3.el9`, `/usr/include/bpf/bpf_helpers.h`, `/usr/include/asm`
- the lachesis RPM builds **inside the jail, non-privileged** → `lachesis-v0.1.0-1.el9.x86_64.rpm`
- `core/cubectl` still builds (the regression check above)

Two dockerfile warnings appear in the build output and are **pre-existing, not from this change**: `InvalidDefaultArgInFrom` (line 46, `FROM tier1_${DIST}` has no default) and `LegacyKeyValueFormat` (line 146, `ENV BASH_ENV /root/.bash_env`). The `ENV` added here uses the modern `key=value` form. Similarly, an rpm transaction-lock failure importing Chrome's GPG key and a pip dependency-resolver warning both occur in untouched layers and do not abort the build.

One environment note for anyone reproducing this on a VM rather than bare metal, since it cost me two false starts and presents as a hang rather than an error: on an OpenStack guest the interface MTU is 1442 (geneve overlay) while `docker0` defaults to 1500, so `dnf` stalls on whichever metalink mirrors do not handle PMTUD — one 25 KB package took 378 s to resolve and then hung. `/etc/docker/daemon.json` with `{"mtu": 1442}` fixes it. The host also needs `docker-buildx-plugin`, since the target invokes `DOCKER_BUILDKIT=1 docker build --progress=plain`.

Not addressed here, deliberately: `core/lachesis/` (the component itself) and `config_lachesis.cpp` are separate tasks — lachesis#280 and #281. This PR only makes the jail *capable* of building them.

#### Additional documentation

```docs

```

### DIFF
--- a/jail/jail.ubi9.dockerfile
+++ b/jail/jail.ubi9.dockerfile
@@ -5,7 +5,11 @@ ARG WEAK_DEP=0
 FROM quay.io/centos/centos:stream9 AS tier1
 ENV LANG=C.UTF-8
 ENV HEX_VER=hex2.0
-ENV GOLANG_VER=1.24.2
+# 1.25.x is required by components whose go.mod declares `go 1.25.3` (lachesis);
+# pinned to the latest 1.25 patch rather than the bare minimum so the jail is not
+# a release behind on toolchain fixes. Bumping this recompiles every Go component
+# in the tree -- see the regression note in the PR.
+ENV GOLANG_VER=1.25.12
 ENV HEX_ARCH=x86_64
 ENV DEVOPS_ENV=__JAIL__
 ENV TZ=Asia/Taipei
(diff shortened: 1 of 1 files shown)


arasHi87:
Closed as collateral when the head branch was renamed from \`arashi.li/jail-bpf-toolchain\` to \`arashi.li/onboard-lachesis\` to widen the scope — GitHub's branch-rename does not retarget a PR's head, and the PR could not be reopened once the old ref was gone.

Continues as #1186, same commit, no content change.

PR #1127, opened 2026-07-21
fix(config): hex CONFIG_PROVIDES_GLOBALS companion + converged storage config fixes

traviswu-bigstack:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it

The cubecos companion to **bigstack-oss/hex#135** (`CONFIG_PROVIDES_GLOBALS`, construct-on-first-use tuning specs, rootfs copy-tracking), plus the storage-config fixes validated with it. Supersedes #1121 (its ceph commit is included here unchanged).

- **ceph perf-tuned defaults for converged nodes** (ex-#1121) — the default-on `ceph.perf.tuned` block was copied from Red Hat's 2019 all-flash dedicated-OSD benchmark; on converged nodes it starved tenant IO. `osd max backfills` 5→1, drop the uncapped 8G static cache, pg log entries 10→500 (no full backfill on a brief OSD blip), `csum type` none→crc32c. Full analysis in #1120.
- **hex bump to feat/config-provides-globals** — scoped `hex_config commit <settings> <module>` no longer runs with unset globals (the class of bug that corrupted configs and keystone endpoints cluster-wide when a module-scoped commit ran live).
- **cube_scan declares itself provider of the config globals** — with lazy specs, the provider must be explicit for scoped commits to populate them.
- **nova reads cinder's backend tuning directly** — now safe because specs construct on first use; removes the copy-paste indirection in `config_nova/neutron/k3s`.
- **`volume_use_multipath` only when an FC/iSCSI backend exists** — unconditional multipath made every RBD-only deployment pay the multipathd path (and broke when it wasn't running).
- **multipath commit robustness** — wait for multipathd to be up; a map refresh failure logs instead of failing the whole commit.

#### Which issue(s) this PR fixes

Fixes #1120

#### Special notes for your reviewer

> Merge **bigstack-oss/hex#135 first**; the hex bump here points at its head (9633566). Content comes unchanged from the lab-validated `combined/local-work` branch.

#### Additional documentation

```docs
Handbook: multipath_link / storage-backend detection notes; hex_config
scoped-commit globals RCA.
```

PR #1194, opened 2026-07-30
feat(ironic): upgrade ironic to openstack antelope

### LINKED ISSUE #610: [Story] [Parallel Group] Upgrade Ironic from Yoga to Antelope
## Description

As a CubeCOS operator, I want Ironic be in Antelope so that we could upgrade to Antelope.

## Acceptance Criteria

- [ ] Migrate configs
  - [ ] Ensure we could enroll and manage machines.
  - [ ] Ensure we could create and utilize the VM corresponding to an enrolled machine.

## Story Points: 5

We did not verify if Ironic would work or not, so we do not have a solid base to compare to. This might be the first encounter to configure and utilize Ironic capabilities.

> | -- | -- | -- |

SekiXu:
Moves ironic and ironic-inspector off the CentOS Stream 9 RDO SIG RPMs and into
the python 3.10 venv at `/opt/openstack-antelope`, following the layout the
keystone, glance, cinder, nova and neutron upgrades already established.

    ironic            20.x -> 21.4.4    (last 2023.1 revision)
    ironic-inspector  10.x -> 11.4.1    (last 2023.1 revision)

Fixes #610.

## Commits in this PR

| commit | what |
|---|---|
| `58f6ddb0` | add `ironic` to the conventional commit scope list |
| `799f1e27` | lock the whole libvirt tree so the `-14` pin actually holds |
| `815823d3` | pin `networking-baremetal` to the 2023.1 series |
| `f644c09c` | the ironic upgrade itself |
| `1f10a5a6` | keep the existing ironic config when cluster addresses are unresolved |
| `da269581` | derive cube_scan's role from the tuning, not the observer static |

The three fixes are ordered before / after the upgrade deliberately: the scope
entry and the libvirt lock are prerequisites for `f644c09c` to build at all, so
they come first and every commit in the branch builds on its own.

## PR boundary

- **`799f1e27` also exists on the #608 branch** (as `d90b8771`, byte-identical
  patch). Whichever of the two merges first, the other disappears on rebase. It
  is included here because `f644c09c` cannot build without it: adding packages
  to `ROOTFS_DNF` changes the download set, which is what flips the latent
  duplicate-removal bug.
- **`58f6ddb0` touches the same `commitScopes` line as #608's `0fb65c96`**
  (which adds `heat`). Whichever merges second has to re-apply its own entry.
- **`815823d3` fixes something #615 left behind** — `neutron.mk` pip installs
  `networking-baremetal` with no version and the constraints file did not pin
  it. `networking-baremetal 8.0.0` (2026-07-28) requires `neutron>=27.0.0.0rc1`
  against our `neutron==22.2.1`, so pip had to backtrack and the resolved
  version depended on the state of PyPI at build time. It is in this PR because
  that package provides the `ironic-neutron-agent` binary ironic starts.
- **`da269581` touches `core/modules/config_cube_scan.cpp`**, which is outside
  ironic. It is here because ironic is the only service that can reach the bug
  (see below) and because the fix and its evidence belong together. Happy to
  split it out if you would rather review it separately.

## The cube_scan bug, and why only ironic hits it

`cube_scan` publishes `MGMT_ADDR`, `SHARED_ID`, `EXTERNAL` and friends from
`ParseNew()`. `Commit()` gated that call on `s_eCubeRole`, a file-static whose
only assignment is in `NotifyCube()` — an observer on cubesys that does not fire
when the commit leaves cubesys untouched. So a commit that changes only some
other module's tuning left the static unset, `Commit()` returned early,
`ParseNew()` never ran, and every `G(...)` consumer saw the compiled-in defaults
`"0.0.0.0"` / `""`.

**ironic is the only OpenStack service shipping with `enabled=false`**, so it is
the only one whose config is first written *after* bootstrap. Every other
service had its conf written during bootstrap, while the globals were valid, and
is not rewritten afterwards — `config_nova.cpp:321-336` builds its database URI
exactly the same way and is safe only for that reason.

Enabling ironic before the fix wrote `@0.0.0.0/ironic` into `ironic.conf`
(conductor stuck in a DB retry loop) and rewrote the keystone catalog to
`http://0.0.0.0:6385` and `http://:6385` (`EXTERNAL` empty drops the host), so
every baremetal call failed with `No host supplied`. Both were correct until
ironic was enabled.

`1f10a5a6` and `da269581` are complementary rather than alternatives:
`da269581` fixes the full-commit path, while a *scoped* commit
(`hex_config commit <settings> ironic`) never invokes `cube_scan` at all, and
that path stays covered by the guard in `config_ironic.cpp`.

## What was removed, and why

`enabled_hardware_types` is pinned to `ipmi` in `config_ironic.cpp:396` and
rewritten on every `Commit()`, so these had no reachable code path:

- `python3-dracclient`, `python3-scciclient` — Dell iDRAC / Fujitsu irmc drivers
- `openstack-ironic-ui` — a Horizon plugin whose host is still on Yoga
  (deferred to #609)

`sushy` stays: `ironic==21.4.4` requires it directly (`requirements.txt:50`).

Also dropped, with reasons: the RPM's `logrotate` drop-ins (replaced by hex's
`LogRotateConf`), `openstack-ironic.service` (the single-process binary; we run
api + conductor separately), the RDO dnsmasq-tftp-server unit (we ship our own
file-server and inspector-dnsmasq), `openstack-ironic-inspector-conductor.service`
(`standalone = true`), and `ironic-dist.conf` (RDO installs it under
`/usr/share/ironic/` and neither unit references it, so not carrying it is
behaviour-preserving).

Conversely, `ipmitool`, `qemu-img`, `mtools`, `dosfstools` and `xorriso` are now
declared explicitly in `ROOTFS_DNF`. They are what the conductor shells out to
and used to arrive either as RPM weak dependencies (`ipmitool` was only a
`Recommends`) or through other components' RPM sets. `mtools` was not in the
tree at all.

## Changed default that was adopted rather than neutralised

`[conductor]permitted_image_formats` was introduced in 21.4.3 as part of the
**CVE-2024-44082** fixes and defaults to `raw,qcow2,iso`. It is deliberately
left unset here, i.e. the new default is adopted, because those three formats
are the ones the ironic project states as supported and tests against, and
CubeCOS has no existing bare metal usage to stay compatible with. Setting it
back to Yoga behaviour would mean disabling a security fix.

Note for whoever runs bare metal in anger: with `[agent]image_download_source`
at `http` and a glance image, the image goes through the conductor, so the
deep-inspection path applies. Only an `image_source` pointing straight at an
external URL bypasses the conductor cache, and that is what
`[conductor]conductor_always_validates_images` exists for — not enabled here, as
it is a new option that costs extra traffic and disk.

## Verification

Built on `centos9_cubecos_seki_200.13` with `MK_TARGET=distclean iso`,
`RC_BASE=v3.1.10` — `CUBE_3.1.10_20260730-0254_f644c09.iso`.

For the requirement "Migrate configs":

```bash
[cc1 ~]# ls -l /etc/ironic/ /etc/ironic-inspector/
/etc/ironic:            drwxr-xr-x root:ironic
  -rw-r----- root:ironic  ironic.conf          (0640)
  -rw-r----- root:root    ironic.conf.def
  -rw-r----- root:ironic  rootwrap.conf
  rootwrap.d/ironic-utils.filters  0644 root:root
/etc/ironic-inspector:  drwxr-x--- root:ironic-inspector   (0750)
  -rw-r----- root:ironic-inspector  inspector.conf         (0640)
  -rw-r----- root:ironic-inspector  inspector-dist.conf
  -rw-r----- root:ironic-inspector  rootwrap.conf
  rootwrap.d/ironic-inspector.filters  0644 root:root
/etc/sudoers.d/{ironic,ironic-inspector}  0440 root:root

[cc1 ~]# grep -n mysql_wsrep_sync_wait /etc/ironic/ironic.conf /etc/ironic-inspector/inspector.conf
/etc/ironic/ironic.conf:45:mysql_wsrep_sync_wait = 1
/etc/ironic-inspector/inspector.conf:19:mysql_wsrep_sync_wait = 1

[cc1 ~]# cat /etc/logrotate.d/ironic /etc/logrotate.d/ironic-inspector
/var/log/ironic/*.log { daily copytruncate maxsize 128M missingok compress delaycompress notifempty }
/var/log/ironic-inspector/*.log { ... }
```

Database schema is at the 2023.1 head for both — the values match the alembic
heads computed from the sdists (59 and 10 revisions respectively):

```bash
[cc1 ~]# mysql -u root -D ironic -e "select version_num from alembic_version"
4dbec778866e
[cc1 ~]# mysql -u root -D ironic_inspector -e "select version_num from alembic_version"
b55109d5063a
[cc1 ~]# su -s /bin/sh -c "/usr/bin/ironic-status upgrade check" ironic
| Check: Database Index Status               | Result: Success |
| Check: Allocations Name Field Length Check | Result: Success |
| Check: Policy File JSON to YAML Migration  | Result: Success |
```

For the requirement "Adjust service arrangements to fulfill basic
functionalities":

```bash
[cc1 ~]# systemctl is-active openstack-ironic-{api,conductor,inspector} ironic-neutron-agent
active / active / active / active
[cc1 ~]# ss -lntp | grep -E ':6385|:5050'
LISTEN 0.0.0.0:6385  users:(("ironic-api",...))
LISTEN 0.0.0.0:5050  users:(("ironic-inspecto",...))

[cc1 ~]# openstack baremetal driver list
+---------------------+----------------+
| Supported driver(s) | Active host(s) |
| ipmi                | cc1            |
+---------------------+----------------+
[cc1 ~]# openstack baremetal conductor list
| Hostname | Conductor Group | Alive |
| cc1      |                 | True  |

[cc1 ~]# openstack baremetal node create --driver ipmi --name ironic-610-smoke \
           --driver-info ipmi_address=192.0.2.10
ironic-610-smoke  d3580c6e-7a78-44a3-9b82-f4c937ac6abe  enroll
[cc1 ~]# openstack baremetal node show ironic-610-smoke -c driver -c provision_state
ipmi / enroll
[cc1 ~]# openstack baremetal node delete ironic-610-smoke
Deleted node ironic-610-smoke
```

`ipmi` with an RFC 5737 address is used instead of `fake-hardware` on purpose:
`fake_hardware.py:23-32` notes that the fake implementations still have to be
enabled in the configuration, which would mean overriding
`enabled_hardware_types` — a value this module rewrites on every commit. `ipmi`
needs only `ipmi_address` (`ipmitool.py:69-71`), so node CRUD exercises the full
API → conductor → DB path without touching any product setting.

For the requirement "Pass health check":

```bash
two phase boot... [yes] / standalone services... [ready] / cube services... [ready] / cluster data... [synced]
```

The libvirt lock, verified by what the build actually removed:

| | before `799f1e27` | after |
|---|---|---|
| `removed_rpms.txt` (libvirt) | 9 files, all `-14` | 9 files, all `-16` |
| `locked_rpms.txt` (libvirt) | 0 | 27 |
| surviving `RPMS/libvirt*.rpm` | mixed `-14`/`-16` | 27, all `11.10.0-14` |

Image split stays clear of the FAT32 cap — largest `rootfs_0.cgz` is 2.72 GiB of
the 4 GiB limit.

## Not verified here

- Real bare metal enrolment, introspection and provisioning — no hardware
  available. This PR covers packaging and configuration only.
- The Horizon panel (#609) and redfish / iDRAC / irmc hardware types.
- IPA ramdisk build or upgrade — CubeCOS imports the deploy kernel/initramfs.

## Unrelated finding

`cluster check` also reports `Network NG [ neutron(3 metadata not all up) ]` on
this VM. `neutron-ovn-metadata-agent` is crash-looping on
`FailedToDropPrivileges: privsep helper command exited non-zero (1)` because
`/etc/neutron/rootwrap.conf`'s `exec_dirs` has no venv path, so rootwrap resolves
`/usr/bin/privsep-helper` (system python 3.9) for a service running out of the
3.10 venv. That is the same mechanism as #1174, which is still draft and covers
only glance and cinder. Nothing in this branch touches `core/neutron/`.

## Definition of Done

Handbook notes: **bigstack-oss/bigstack-handbook#177** — one architecture topic
for the packaging, plus two known-issues for the bugs this hop exposed
(`cube_scan` globals, and `ROOTFS_DNF` version not being a pin). `handbook
doctor` HIGH=0.

### DIFF
--- a/core/modules/config_cube_scan.cpp
+++ b/core/modules/config_cube_scan.cpp
@@ -310,7 +310,9 @@ ParseNew(void)
             return false;
     }
 
-    bool isCtrl = IsControl(s_eCubeRole);
+    // ParseOld() already derives this from the tuning; keep the two symmetric so the
+    // shared id does not silently change shape depending on whether NotifyCube() ran
+    bool isCtrl = IsControl(r);
     bool isMaster = IsMaster(isCtrl, mgmtAddr, s_ctrlAddrs.newValue());
     std::string ctrl = GetController(isCtrl, s_hostname.newValue(), s_controller.newValue());
     std::string ctrlIp = isCtrl ? mgmtAddr : s_controllerIp.newValue();
@@ -427,10 +429,14 @@ Prepare(bool modified, int dryLevel)
 static bool
 Commit(bool modified, int dryLevel)
 {
-    if (IsUndef(s_eCubeRole))
-        return true;
-
+    // read the role straight from the tuning rather than from s_eCubeRole, which is
+    // only ever assigned in NotifyCube() -- and that observer does not fire when the
+    // commit leaves cubesys untouched. Bailing out on the unset static skipped
+    // ParseNew() entirely, so every G(...) below stayed at its compiled-in default.
     CubeRole_e r = GetCubeRole(s_cubeRole.newValue());
+
+    if (IsUndef(r))
+        return true;
 
     if (IfProviderReq(r)) {
         // in case provider interface is disabled in policy
--- a/core/heavyfs/os-antelope-pip-upper-constraints.txt
+++ b/core/heavyfs/os-antelope-pip-upper-constraints.txt
@@ -591,3 +591,4 @@ sphinxcontrib-applehelp===1.0.2
 scikit-learn===1.1.2
 pytest-cov===4.0.0
 setuptools==65.7.0
+networking-baremetal===6.1.1
--- a/core/ironic/inspector-dist.conf
+++ b/core/ironic/inspector-dist.conf
@@ -1,0 +1,4 @@
+[DEFAULT]
+log_dir = /var/log/ironic-inspector
+state_path = /var/lib/ironic-inspector
+use_stderr = False
--- a/core/ironic/inspector-rootwrap.conf
+++ b/core/ironic/inspector-rootwrap.conf
@@ -1,0 +1,27 @@
+# Configuration for ironic-inspector-rootwrap
+# This file should be owned by (and only-writeable by) the root user
+
+[DEFAULT]
+# List of directories to load filter definitions from (separated by ',').
+# These directories MUST all be only writeable by root !
+filters_path=/etc/ironic-inspector/rootwrap.d,/usr/share/ironic-inspector/rootwrap
+
+# List of directories to search executables in, in case filters do not
+# explicitly specify a full path (separated by ',')
+# If not specified, defaults to system PATH environment variable.
+# These directories MUST all be only writeable by root !
+exec_dirs=/sbin,/usr/sbin,/bin,/usr/bin
+
+# Enable logging to syslog
+# Default value is False
+use_syslog=False
+
+# Which syslog facility to use.
+# Valid values include auth, authpriv, syslog, user0, user1...
+# Default value is 'syslog'
+syslog_log_facility=syslog
+
+# Which messages to log.
+# INFO means log all usage
+# ERROR means only log unsuccessful attempts
+syslog_log_level=ERROR
--- a/core/ironic/inspector.conf.sample
+++ b/core/ironic/inspector.conf.sample
@@ -1,0 +1,2046 @@
+[DEFAULT]
+
+#
+# From ironic_inspector
+#
+
+# IP to listen on. (string value)
+#listen_address = ::
+
+# Port to listen on. (port value)
+# Minimum value: 0
+# Maximum value: 65535
+#listen_port = 5050
+
+# Unix socket to listen on. Disables listen_address and listen_port.
+# (string value)
+#listen_unix_socket = <None>
+
+# File mode (an octal number) of the unix socket to listen on. Ignored
+# if listen_unix_socket is not set. (integer value)
+#listen_unix_socket_mode = <None>
+
+# Name of this node. This can be an opaque identifier. It is not
+# necessarily a hostname, FQDN, or IP address. However, the node name
+# must be valid within an AMQP key, and if using ZeroMQ, a valid
+# hostname, FQDN, or IP address. (string value)
+#
+# This option has a sample default set, which means that
+# its actual default value may vary from the one documented
+# below.
+#host = localhost
+
+# Authentication method used on the ironic-inspector API. "noauth",
+# "keystone" or "http_basic" are valid options. "noauth" will disable
+# all authentication. (string value)
+# Possible values:
+# noauth - no authentication
+# keystone - use the Identity service for authentication
+# http_basic - HTTP basic authentication
+#auth_strategy = keystone
+
+# Path to Apache format user authentication file used when
+# auth_strategy=http_basic (string value)
+#http_basic_auth_user_file = /etc/ironic-inspector/htpasswd
+
+# Timeout after which introspection is considered failed, set to 0 to
+# disable. (integer value)
+# Maximum value: 315576000
+#timeout = 3600
+
+# Amount of time in seconds, after which repeat clean up of timed out
+# nodes and old nodes status information. WARNING: If set to a value
+# of 0, then the periodic task is disabled and inspector will not sync
+# with ironic to complete the internal clean-up process. Not advisable
+# if the deployment uses a PXE filter, and will result in the ironic-
+# inspector ceasing periodic cleanup activities. (integer value)
+# Minimum value: 0
+#clean_up_period = 60
+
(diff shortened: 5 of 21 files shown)


core/modules/config_cube_scan.cpp
Eandalf-Bigstack, line 1: The issue of the blank value of `SHARED_ID` and also global tuning family parsing miss while committing a single module has been identified and resolved fully on branch develop via #1127 . However, if the fix works here, we could go with it since the code change is not affected even if the parsing is working correctly before executing those statements.

Eandalf-Bigstack (approved):

Issue #601, opened 2026-01-27
[Task] Language - Python Upgrade from 3.9 to 3.10


jdjgya:
## Description

> Overview of what should be done.

As a CubeCOS operator, I want to fix Python 3.9 EOL security issue so that we could resolve the security concern.

## Requirements

> A list of technical requirements must be met for completion. These may include, but are not limited to:
> - API data conforming to a specified structure.
> - A page delivering defined functionalities.
> - Passing all unit tests.
> - Passing a code review.

- [ ] Bump Python from 3.9 to 3.10

## Notes (Optional)
Insert content here.

Eandalf-Bigstack:
### Skyline: python 3.11 + a dedicated Caracal venv

Skyline was the last component running on **system python 3.9**. It is now moved into a dedicated Caracal venv at `/opt/openstack-caracal`.

It was **not** moved into the existing `/opt/openstack-antelope` venv. Our forks of `skyline-apiserver` and `skyline-console` branch off upstream `master` at `4.0.1-6` and `4.0.0.0rc1-56` respectively — Caracal-era, not Antelope — so putting them in the 2023.1 venv would pin 2024.1 code to a 2023.1 dependency set and drag Skyline's requirements into the other services. See #641 for the full version verification.

Landed, all verified in the build jail (`heavy_rootfs`) and on jim-1cc:

- **python 3.11.15** built alongside 3.10, same recipe shape as the 3.10 block (`/dev` bind-mounted, `--enable-optimizations`, `make altinstall`). On jim-1cc the interpreter is md5-identical to the build tree and its stdlib imports clean against the node's own libs.
- **`/opt/openstack-caracal` venv**, created from upstream's `unmaintained/2024.1` upper-constraints copied down verbatim. `sys.path` is `/usr/local/lib/python3.11` plus the venv's own site-packages — no Antelope entry, no python 3.9 entry. The Antelope venv (3.10) and system python (3.9) are unaffected.
- **`skyline-apiserver` and `skyline-console` installed into that venv**, replacing the system 3.9 install. Three call sites outside `skyline.mk` hardcoded the 3.9 path and moved with it: the unit's `ExecStart` gunicorn, nginx's `skyline_console` static root, and the alembic invocation in `config_skyline.cpp`. `ROOTFS_PIP_NC` no longer carries `gunicorn` — it existed only for Skyline, and gunicorn now comes from the venv's own requirements.

[...]

PR #1201, opened 2026-07-31
fix: close the gaps found auditing the heat and ironic antelope migrations

Eandalf-Bigstack:
#### What type of PR is this?

- bug

- Fixes four defects found while auditing the heat (#1191) and ironic (#1194) antelope migrations against the RDO antelope distgit specs, plus one pre-existing MariaDB logging bug and one cyborg connection-pool bug surfaced by a whole-service sweep of the antelope stack.
- `ironic-lib` ships a second rootwrap filter set as wheel data that RDO relocates from `python-ironic-lib.spec` rather than from the ironic spec, so porting only `openstack-ironic.spec` lost it. Without it `ironic-rootwrap` denies every disk-manipulation command the conductor needs.
- The RDO heat spec *appends* `[revision] heat_revision=%{version}` to `heat-dist.conf` at build time. Dropping that file therefore silently reverted `heat_revision` to its `unknown` default, which heat serves at `GET /v1/{tenant}/build_info`.
- `config_mysql.cpp` rotated `/var/log/mysql/*.log` while pointing `log_error` at `/var/log/mariadb/mysql_error.log`, and nothing created `/var/log/mariadb`. mariadbd could not open its error log at all, so the database had no error log on disk.
- Fixing that error log has a consequence that needs fixing with it (raised in review). While mariadbd could not open `log_error` it fell back to stderr, which systemd routes to journald and rsyslog writes into `/var/log/messages` — the first path filebeat collects. Database errors were reaching LogAnalytics through a fallback nobody chose. Once mariadbd writes its own file that route disappears, and none of the 13 paths in `config_logstash.cpp` covered `/var/log/mariadb`, so the mysql fix on its own would have removed the database from LogAnalytics entirely. `config_logstash.cpp` now collects it.
- cyborg's `model_query()` returns its query after its own `_session_for_read()` scope has exited, leaving a pooled connection checked out with no owner until a weakref finalizer resets it — under eventlet that runs in the hub and raises `do not call blocking functions from the mainloop`.
- Finally, `git-conventional-commits.yaml` never listed `cyborg` or `mysql`, so two of the commits here were technically invalid. Scanning the history showed 53 scopes in the same position, so the list is completed and tidied rather than patched twice.

- Refs #601 — **not closing.** #601 is the wrapping issue for the Yoga-to-Antelope epic and several OpenStack services are still on Yoga.

> The cyborg change is deliberately scoped to the nine read methods on the `report_data` path. Upstream fixed this by taking `context.session` instead, but that is part of the SQLAlchemy 2.0 refactor — 30 hunks migrating 27 `_session_for_read/write` uses to 23 `context.session` uses — which is too invasive to backport onto the `unmaintained/2023.1` tree we install. Sixteen further unenclosed `model_query()` call sites remain on paths `report_data` does not reach; the upstream migration covers them when cyborg is next bumped.

> The `.cpp` hunks (`config_mysql.cpp`, `config_logstash.cpp`) and the `.mk` hunks only take effect in a built image — `config_*.cpp` is statically linked into the setuid `hex_config` binary and cannot be hot-swapped onto a node. On jim-1cc the equivalent on-node state was applied by hand — `/etc/logrotate.d/mysql` rewritten to exactly what the fixed `LogRotateConf` emits, `/var/log/mariadb` created, and `/var/log/mariadb/*.log` appended to `/etc/filebeat/filebeat.yml` exactly where `config_logstash.cpp` now emits it — which is what the evidence below was captured against.

> Two things worth knowing about the environment the evidence came from: ironic's DB on jim-1cc was two alembic revisions behind and was migrated to head (`9ef41f07cb58` → `0ac0f39bc5aa` → `4dbec778866e`) before testing, and ironic itself is intentionally left disabled, so its evidence is config/rootwrap-level rather than service-level.

#### Additional documentation

```bash
[cc1 ~]# hex_cli -c cluster check
          Service  Status  Report
           IaasDb      ok  [ mysql(v) ]
            Image      ok  [ glance(v) ]
            LBaaS      ok  [ octavia(v) ]
```

For the requirement "ironic: ironic-lib's rootwrap filter set is installed and authorises the conductor's disk commands",

```bash
[cc1 ~]# ls -l /etc/ironic/rootwrap.d/
total 4
-rw-r--r-- 1 root root 230 Jul 31 14:38 ironic-utils.filters

[cc1 ~]# su -s /bin/sh -c "sudo /usr/bin/ironic-rootwrap /etc/ironic/rootwrap.conf lsblk --version" ironic
/usr/bin/ironic-rootwrap: Unauthorized command: lsblk --version (no filter matched)
[cc1 ~]# su -s /bin/sh -c "sudo /usr/bin/ironic-rootwrap /etc/ironic/rootwrap.conf blkid -V" ironic
/usr/bin/ironic-rootwrap: Unauthorized command: blkid -V (no filter matched)
[cc1 ~]# su -s /bin/sh -c "sudo /usr/bin/ironic-rootwrap /etc/ironic/rootwrap.conf mount --version" ironic
mount from util-linux 2.37.4 (libmount 2.37.4: selinux, btrfs, namespaces, assert, debug)

# ---- apply the relocation this PR adds to ironic.mk ----
[cc1 ~]# install -p -D -m 644 /opt/openstack-antelope/etc/ironic/rootwrap.d/ironic-lib.filters \
    /etc/ironic/rootwrap.d/ironic-lib.filters
[cc1 ~]# chown root:root /etc/ironic/rootwrap.d/ironic-lib.filters
[cc1 ~]# ls -l /etc/ironic/rootwrap.d/
total 8
-rw-r--r-- 1 root root 911 Jul 31 14:42 ironic-lib.filters
-rw-r--r-- 1 root root 230 Jul 31 14:38 ironic-utils.filters

[cc1 ~]# for c in "blkid -V" "blockdev --version" "lsblk --version" "wipefs -V" \
             "sgdisk --version" "partprobe --version" "parted --version" "qemu-img --version"; do
      su -s /bin/sh -c "sudo /usr/bin/ironic-rootwrap /etc/ironic/rootwrap.conf $c" ironic 2>&1 | head -1
  done
blkid from util-linux 2.37.4  (libblkid 2.37.4, 14-Feb-2022)
blockdev from util-linux 2.37.4
lsblk from util-linux 2.37.4
wipefs from util-linux 2.37.4
sgdisk from GPT fdisk (sgdisk) version 1.0.7
partprobe (GNU parted) 3.5
parted (GNU parted) 3.5
qemu-img version 10.1.0 (qemu-kvm-10.1.0)

[cc1 ~]# su -s /bin/sh -c "/usr/bin/ironic-status upgrade check" ironic
+--------------------------------------------+
| Check: Object versions                     |
| Result: Success                            |
+--------------------------------------------+
| Check: Database Index Status               |
| Result: Success                            |
+--------------------------------------------+
| Check: Allocations Name Field Length Check |
| Result: Success                            |
+--------------------------------------------+
| Check: Policy File JSON to YAML Migration  |
| Result: Success                            |
+--------------------------------------------+
```

For the requirement "ironic: pykickstart's absence does not break the conductor's imports",

```bash
[cc1 ~]# rpm -q pykickstart
package pykickstart is not installed
[cc1 ~]# grep -n pykickstart \
    /opt/openstack-antelope/lib/python3.10/site-packages/ironic/common/pxe_utils.py \
    /opt/openstack-antelope/lib/python3.10/site-packages/ironic/drivers/modules/deploy_utils.py
.../ironic/common/pxe_utils.py:1158:            "missing. Please install pykickstart package to enable "
.../ironic/drivers/modules/deploy_utils.py:1250:        # https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html

[cc1 ~]# /opt/openstack-antelope/bin/python -c \
    "import ironic.common.pxe_utils, ironic.drivers.modules.deploy_utils; print('both modules import fine')"
both modules import fine

[cc1 ~]# grep -E '^enabled_(deploy|hardware)' /etc/ironic/ironic.conf
enabled_deploy_interfaces = direct
enabled_hardware_types = ipmi
```

For the requirement "heat: build_info reports the release instead of the unknown default",

```bash
# ---- before ----
[cc1 ~]# curl -s -H "X-Auth-Token: $TOK" http://10.1.0.1:8004/v1/$PROJ/build_info
{"api": {"revision": "unknown"}, "engine": {"revision": "unknown"}}

# ---- apply what this PR adds to heat.mk, then restart heat ----
[cc1 ~]# sed -i '/^\[revision\]$/a heat_revision = 20.0.1' /etc/heat/heat.conf
[cc1 ~]# sed -i '/^\[revision\]$/a heat_revision = 20.0.1' /etc/heat/heat.conf.def
[cc1 ~]# grep -A1 '^\[revision\]$' /etc/heat/heat.conf
[revision]
heat_revision = 20.0.1
[cc1 ~]# systemctl restart openstack-heat-api openstack-heat-api-cfn openstack-heat-engine

# ---- after ----
[cc1 ~]# curl -s -H "X-Auth-Token: $TOK" http://10.1.0.1:8004/v1/$PROJ/build_info
{"api": {"revision": "20.0.1"}, "engine": {"revision": "20.0.1"}}
```

For the requirement "heat: no regression — stack lifecycle and the orchestration health path still work",

```bash
[cc1 ~]# openstack orchestration service list -f value
cc1 heat-engine 580339c2-b796-44fb-a591-918feafe4db0 cc1 engine 2026-07-31T07:44:18.000000 up

[cc1 ~]# heat-status upgrade check
+-------------------------------------------+
| Check: Policy File JSON to YAML Migration |
| Result: Success                           |
+-------------------------------------------+

[cc1 ~]# cat > /tmp/land-test-stack.yaml <<'YML'
heat_template_version: 2021-04-16
resources:
  rand:
    type: OS::Heat::RandomString
    properties:
      length: 12
outputs:
  value:
    value: { get_attr: [rand, value] }
YML
[cc1 ~]# openstack stack create -t /tmp/land-test-stack.yaml --wait land-test-1191 -f value -c stack_status
2026-07-31 07:44:53Z [land-test-1191]: CREATE_COMPLETE  Stack CREATE completed successfully
CREATE_COMPLETE
[cc1 ~]# openstack stack resource list land-test-1191 -f value
rand land-test-1191-rand-jlzjichmvuh5 OS::Heat::RandomString CREATE_COMPLETE 2026-07-31T07:44:53Z
[cc1 ~]# openstack stack output show land-test-1191 value -f value -c output_value
Cfc0VxcvjNhY
[cc1 ~]# openstack stack delete --yes --wait land-test-1191
2026-07-31 07:45:09Z [land-test-1191.rand]: DELETE_IN_PROGRESS  state changed
[cc1 ~]# openstack stack list -f value | wc -l
0
```

For the requirement "mysql: mariadb has an error log on disk and logrotate rotates it",

```bash
# ---- before ----
[cc1 ~]# mysql -N -e "select @@log_error"
/var/log/mariadb/mysql_error.log
[cc1 ~]# ls -l /var/log/mariadb/
ls: cannot access '/var/log/mariadb/': No such file or directory
[cc1 ~]# head -1 /etc/logrotate.d/mysql
/var/log/mysql/*.log {
[cc1 ~]# ls -d /var/log/mysql
ls: cannot access '/var/log/mysql': No such file or directory

# ---- apply what this PR adds to mysql.mk and config_mysql.cpp ----
[cc1 ~]# install -d -m 750 /var/log/mariadb
[cc1 ~]# chown mysql:mysql /var/log/mariadb
[cc1 ~]# cat /etc/logrotate.d/mysql
/var/log/mariadb/*.log {
  daily
  copytruncate
  maxsize 128M
  missingok
  compress
  delaycompress
  notifempty
}

# ---- FLUSH ERROR LOGS reopens the error log in place, no restart needed ----
[cc1 ~]# mysql -e "FLUSH ERROR LOGS"
[cc1 ~]# ls -l /var/log/mariadb/
total 0
-rw-rw---- 1 mysql mysql 0 Jul 31 17:00 mysql_error.log
[cc1 ~]# ls -l /proc/$(systemctl show mariadb -p MainPID --value)/fd | grep mysql_error
l-wx------ 1 mysql mysql 64 Jul 29 12:40 2 -> /var/log/mariadb/mysql_error.log
l-wx------ 1 mysql mysql 64 Jul 29 12:34 979 -> /var/log/mariadb/mysql_error.log

# ---- prove writes land there (log_warnings 1 -> 2 for the test, restored after) ----
[cc1 ~]# mysql -e "SET GLOBAL log_warnings=2"
[cc1 ~]# mysql -h 10.1.0.1 -e "select sleep(8)" >/dev/null 2>&1 & sleep 2; kill -9 %1
[cc1 ~]# cat /var/log/mariadb/mysql_error.log
2026-07-31 17:02:24 5339 [Warning] Access denied for user 'root'@'10.1.0.1' (using password: NO)
[cc1 ~]# mysql -e "SET GLOBAL log_warnings=1"

[cc1 ~]# logrotate --debug /etc/logrotate.d/mysql
rotating pattern: /var/log/mariadb/*.log  after 1 days (no old logs will be kept)
considering log /var/log/mariadb/mysql_error.log
[cc1 ~]# systemctl start logrotate
[cc1 ~]# systemctl show logrotate -p Result -p ExecMainStatus --value
success
0
[cc1 ~]# systemctl list-units --state=failed --no-legend | wc -l
0
[cc1 ~]# systemctl is-active mariadb; mysql -N -e "show status like 'Uptime'"
active
Uptime	189886
```

For the requirement "logstash: mariadb's error log still reaches LogAnalytics once it stops passing through journald",

```bash
# ---- before: the mysql fix above is already applied, so mariadbd is off stderr ----
[cc1 ~]# ls -l /proc/$(systemctl show mariadb -p MainPID --value)/fd/2
l-wx------ 1 mysql mysql 64 Aug  1 21:26 /proc/11607/fd/2 -> /var/log/mariadb/mysql_error.log
[cc1 ~]# grep -ci "mariadbd\|mysqld" /var/log/messages
0
[cc1 ~]# grep -c "mariadb\|mysql" /etc/filebeat/filebeat.yml
0
# -> the database now reaches neither journald nor LogAnalytics

# ---- apply what this PR adds to config_logstash.cpp ----
[cc1 ~]# sed -i '\|- /var/log/ceph/ceph.log|a\    - /var/log/mariadb/*.log' /etc/filebeat/filebeat.yml
[cc1 ~]# systemctl restart filebeat; systemctl is-active filebeat
active
[cc1 ~]# systemctl show filebeat -p User    # empty => runs as root, so 750 mysql:mysql is readable
User=

# ---- provoke real warnings (log_warnings 1 -> 2 for the test, restored after) ----
[cc1 ~]# mysql -e "SET GLOBAL log_warnings=2"
[cc1 ~]# for i in $(seq 1 15); do mysql -u root -pWrongPw-$i -h 10.1.0.1 -e "SELECT 1" >/dev/null 2>&1; done
[cc1 ~]# mysql -e "SET GLOBAL log_warnings=1"; mysql -N -e "select @@global.log_warnings"
1

# ---- filebeat harvests the file and it reaches opensearch ----
[cc1 ~]# ls -l /proc/$(systemctl show filebeat -p MainPID --value)/fd | grep -o '/var/log/mariadb.*'
/var/log/mariadb/mysql_error.log
[cc1 ~]# grep -c mariadb /var/lib/filebeat/registry/filebeat/log.json
2
[cc1 ~]# curl -s -H 'Content-Type: application/json' http://localhost:9200/_search \
    -d '{"size":3,"query":{"query_string":{"query":"\"Access denied for user\""}}}' \
  | python3 -c "import sys,json; d=json.load(sys.stdin); print('total:',d['hits']['total'])"
total: {'value': 16, 'relation': 'eq'}
```

The shipped document read back from opensearch, in `logs-20260801-%{program}` alongside the other logs whose format logstash's grok does not parse:

```json
{"log": {"file": {"path": "/var/log/mariadb/mysql_error.log", "inode": "2276529",
                  "fingerprint": "270fa3314a9d38cb..."}, "offset": 0},
 "message": "2026-08-01 23:39:48 8323 [Warning] Access denied for user 'root'@'10.1.0.1' (using password: YES)"}
```

`offset: 0` is worth noting — filebeat backfills from the start of the file, so the lines written before the harvester attached are ingested too.

> **A threshold this change does not clear, and deliberately does not try to.** filebeat 9.2.3's `filestream` input identifies files by a 1024-byte fingerprint and will not begin ingesting one until it reaches that size; the node already logs `22 files are too small to be ingested` on every 10-second scan. `mysql_error.log` sat at 98 bytes and was skipped outright — it was only harvested after the warning above was repeated 15 times to push it past 1568 bytes. A quiet day (the previous day's entire error log was 97 bytes) therefore rotates away un-ingested. This fails in the harmless direction: a database in real trouble clears 1 KB easily, so collection works precisely when it matters. Lowering `prospector.scanner.fingerprint.length` on a dedicated `filestream` input would close the remainder, but the threshold is pre-existing platform behaviour affecting two dozen files across every service, so it belongs in its own change rather than here.

For the requirement "cyborg: pooled connections are returned instead of abandoned",

```bash
# The measurement has to take the pool from the session _session_for_read() actually
# uses. cyborg's _session_for_read() goes through the GLOBAL enginefacade while
# get_session() goes through main_context_manager, and those are two engines with two
# pools -- sampling the wrong one hides the leak completely. checkedout() is read while
# the result is still alive, because the finalizer returns the connection at GC.

# ---- before ----
[cc1 ~]# /opt/openstack-antelope/bin/python pooltest.py
    engine in use: Engine(mysql+pymysql://cyborg:***@10.1.0.1/cyborg)
    main_context_manager engine: Engine(mysql+pymysql://cyborg:***@10.1.0.1/cyborg)  DIFFERENT POOL
    baseline                   checkedout=0  size=5  overflow=-4
    device_list                rows=0    checkedout before=0 held=1 after_gc=0   LEAKED (held outside any scope)
    device_list_by_filters     rows=0    checkedout before=0 held=1 after_gc=0   LEAKED (held outside any scope)
    deployable_list            rows=0    checkedout before=0 held=1 after_gc=0   LEAKED (held outside any scope)
    attach_handle_list         rows=0    checkedout before=0 held=1 after_gc=0   LEAKED (held outside any scope)
  RESULT: 4 of 4 calls left a connection checked out

# ---- after installing this PR's patched api.py ----
[cc1 ~]# /opt/openstack-antelope/bin/python pooltest.py
    baseline                   checkedout=0  size=5  overflow=-4
    device_list                rows=0    checkedout before=0 held=0 after_gc=0   clean
    device_list_by_filters     rows=0    checkedout before=0 held=0 after_gc=0   clean
    deployable_list            rows=0    checkedout before=0 held=0 after_gc=0   clean
    attach_handle_list         rows=0    checkedout before=0 held=0 after_gc=0   clean
  RESULT: 0 of 4 calls left a connection checked out
```

For the requirement "cyborg: the QueuePool errors stop and the services stay healthy",

```bash
# ---- before: 40 occurrences over 48h, median gap 66 min ----
[cc1 ~]# grep -c 'Exception during reset or similar' /var/log/cyborg/cyborg-conductor.log.1 \
                                                     /var/log/cyborg/cyborg-conductor.log
/var/log/cyborg/cyborg-conductor.log.1:28
/var/log/cyborg/cyborg-conductor.log:12
[cc1 ~]# grep -m1 -A3 'Exception during reset or similar' /var/log/cyborg/cyborg-conductor.log
ERROR sqlalchemy.pool.impl.QueuePool [-] Exception during reset or similar: AssertionError: do not call blocking functions from the mainloop
  File ".../sqlalchemy/pool/base.py", line 753, in _finalize_fairy
    fairy._reset(pool)
  File ".../eventlet/hubs/__init__.py", line 141, in trampoline
    assert hub.greenlet is not current, 'do not call blocking functions from the mainloop'

# ---- after ----
[cc1 ~]# systemctl restart cyborg-conductor cyborg-api cyborg-agent
[cc1 ~]# for u in cyborg-api cyborg-conductor cyborg-agent; do echo "$u $(systemctl is-active $u)"; done
cyborg-api active
cyborg-conductor active
cyborg-agent active
[cc1 ~]# openstack accelerator device list -f value; echo "exit=$?"
exit=0
[cc1 ~]# openstack accelerator device profile list -f value; echo "exit=$?"
exit=0

[cc1 ~]# awk '$0>="2026-07-31 17:29:47"' /var/log/cyborg/cyborg-conductor.log > /tmp/since.log
[cc1 ~]# grep -c 'Start differing devices'            /tmp/since.log   # report_data cycles
6
[cc1 ~]# grep -c 'Exception during reset or similar'  /tmp/since.log
0
[cc1 ~]# grep -cE ' (ERROR|CRITICAL) '                /tmp/since.log
0

# connections held in MariaDB: 6 (oldest idle 90 min) before, 2 (oldest idle 135 s) after
[cc1 ~]# mysql -N -e "select count(*), max(time) from information_schema.processlist where user='cyborg'"
2	135
```

For the requirement "the commit scope list is complete, deduplicated and free of malformed entries",

```bash
[repo]$ git log --format=%s origin/openstack-develop | wc -l
837
[repo]$ python3 - <<'PY'
import json, re, subprocess
line = next(l for l in open('git-conventional-commits.yaml') if l.strip().startswith('commitScopes:'))
s = json.loads(line.split('commitScopes:', 1)[1].strip())
subs = subprocess.run(['git','log','--format=%s','origin/openstack-develop'],
                      capture_output=True, text=True).stdout.splitlines()
pat = re.compile(r'^(feat|fix|refactor|perf|style|test|build|ops|docs|chore|merge|revert)\(([^)]+)\)!?:')
used = set()
for x in subs:
    m = pat.match(x.strip())
    if m: used.update(y.strip() for y in m.group(2).split(',') if y.strip())
print("scopes            :", len(s))
print("duplicates        :", sorted({x for x in s if s.count(x) > 1}) or "none")
print("malformed entries :", [x for x in s if not re.fullmatch(r'[a-z0-9][a-z0-9_.-]*', x)] or "none")
print("this PR's scopes  :", all(k in s for k in ('cyborg','mysql','heat','ironic')))
print("still unmatched   :", sorted(used - set(s)))
PY
scopes            : 119
duplicates        : none
malformed entries : none
this PR's scopes  : True
still unmatched   : ['ClusterReady', 'apache2', 'app frameworks', 'issue template', 'k3s cephcsi', 'keycloak-login', 'non-rolling', 'oci image', 'rc.nicdetect.sh', 'rolling']
```

The ten still-unmatched scopes are deliberate: eight are malformed or already represented (`issue template` and `k3s cephcsi` contain spaces — `issue-templates` replaces the former, and `k3s` and `cephcsi` are now listed separately; `rc.nicdetect.sh` is a filename; `ClusterReady` is CamelCase; `apache2` duplicates `httpd`; `rolling`/`non-rolling` are covered by `rolling-upgrade`; `keycloak-login` by `keycloak`). The remaining two, `app frameworks` and `oci image`, are the spaced spellings this PR retires in favour of `app-frameworks` and `oci`, which retroactively invalidates three old subjects — that is the point of removing spaces from the vocabulary, and `includeInvalidCommits: true` keeps them in the changelog regardless.


core/mysql/mysql.mk
```diff
@@ -34,6 +34,13 @@ ROOTFS_DNF_DL_FROM += $(MARIADB_URL)/$(GALERA_RPM).x86_64.rpm
 ROOTFS_DNF += rsync
 ROOTFS_DNF_NOARCH += python3-PyMySQL
 
+# config_mysql.cpp points log_error at /var/log/mariadb/mysql_error.log, but nothing
+# created that directory: the MariaDB.org rpms ship no /var/log/mariadb (the distro
+# mariadb package does, and we do not install it). mariadbd could not open its error log
+# and fell back to stderr, so the database had no error log on disk at all.
+rootfs_install::
+	$(Q)chroot $(ROOTDIR) install -d -m 750 -o mysql -g mysql /var/log/mariadb
```
SekiXu: nit — this is the only `install -d` in the tree carrying `-o`/`-g` (`grep -rn "install -d.*-o " --include="*.mk" .` is otherwise empty). The other fourteen log directories use `install -d -m` plus a separate `chroot chown`, which is what makes `grep -rh "chown.*var/log" --include="*.mk" .` list every log directory's owner in one command — that is how cinder's and nova's `:root` groups are visible. mariadb's owner hides in the arguments here, and the grep misses it without saying so.

```suggestion
	$(Q)chroot $(ROOTDIR) install -d -m 750 /var/log/mariadb
	$(Q)chroot $(ROOTDIR) chown mysql:mysql /var/log/mariadb
```

SekiXu (approved):
Approve. One thing I'd like folded into this PR, plus a nit inline on `mysql.mk`.

## The mysql fix drops mariadb's errors out of LogAnalytics

Today those messages reach opensearch: mariadbd cannot open `log_error`, falls back to stderr, stderr is journald, and that lands in `/var/log/messages` — which is the first path filebeat collects. Measured on a 1cc:

```
[cc1 ~]# MPID=$(systemctl show mariadb -p MainPID --value); ls -l /proc/$MPID/fd/2
lrwx------ 1 mysql mysql 64 Aug  1 00:59 /proc/11701/fd/2 -> socket:[21877]
[cc1 ~]# grep -ci "mariadbd\|mysqld" /var/log/messages
64
[cc1 ~]# grep -n "var/log/messages" /etc/filebeat/filebeat.yml
4:    - /var/log/messages
```

After the fix mariadbd writes `/var/log/mariadb/mysql_error.log` directly and stops going through stderr, so journald and `/var/log/messages` lose it — and the new path is not collected (`grep -c "mariadb\|mysql" /etc/filebeat/filebeat.yml` returns 0; `config_logstash.cpp:86-113` lists 25 paths, none of them mariadb). Net effect is that database errors disappear from LogAnalytics.

One line in that same list closes it:

```c
    fprintf(fout, "    - /var/log/mariadb/*.log\n");
```

filebeat runs as root (no `User=` in its unit) so `750 mysql:mysql` is readable; `copytruncate` + `delaycompress` yield `.log.1` / `.log.2.gz`, which `*.log` will not double-collect; and `config_mysql.cpp` is already here, so this costs no extra hex_config build.

SekiXu:
For the record, the things I checked against the tree rather than taking the body's
word for.

**heat**
- `heat.conf.sample` has exactly one `[revision]` section (line 2383), so the `sed`
  has a landing site. A missing section would have made it a silent no-op.
- `config_heat.cpp:319` `LoadConfig(CONF ".def")` → `:426` `WriteConfig(CONF)`: the
  `.def` really is parsed into key/values, so a key the cpp does not touch survives
  regeneration, as your comment says.
- In `heat.mk` the `sed` runs before the `cp` to `.def` — the order that matters.

**ironic**
- Taking `ironic-lib.filters` from the venv prefix matches what `heat.mk:76-78`
  already does for `api-paste.ini` / `environment.d` / `templates`, and the install
  step is after the pip install that puts it there.

**mysql**
- `config_mysql.cpp:24` (`ERROR_LOG`) and `:46` (the logrotate glob) did disagree.
- `hex/src/hex_sdk_library/logrotate/logrotate.cpp:102` writes `missingok`
  unconditionally — which is why a glob matching nothing is indistinguishable from
  one that works.
- Creating `/var/log/<svc>` inside the image is what twelve other `.mk` files do,
  so nothing mounts over it at runtime.
- `mysql` exists as a user by the time that `install -d` runs — the pre-existing
  first `rootfs_install::` already tars `/var/lib/mysql`.

**cyborg**
- `_paginate_query()` ends in `return query.all()`, so the new scopes do materialize
  inside themselves. Had it returned the query, wrapping would have changed nothing.
- Of 45 `model_query()` call sites: 15 already sit under `_session_for_write()`,
  9 are wrapped here, and the rest are reached only by an API request. The quota
  helpers look unenclosed but run under `quota_reserve()`'s and
  `reservation_commit()`'s write scope, and `_get_quota_usages()` is a
  `SELECT FOR UPDATE` that would be wrong inside a read scope. The boundary is right.

Eandalf-Bigstack:
@SekiXu Great catch! Thanks. Both the issue filebeat not capturing mariadb logs and the issue mariadb install script styling have been resolved.

Btw, I appreciate the enthusiasm, but please don't feel the need to review PRs over the weekend! Go log off, enjoy your downtime, and we can dive back into this on Monday.

PR #1202, opened 2026-08-01
feat(monasca): upgrade monasca from yoga to antelope

Eandalf-Bigstack:
#### What type of PR is this?

- feature

#### What this PR does / why we need it

- Move Monasca out of the system python 3.9 and into the OpenStack Antelope venv at `/opt/openstack-antelope`, with its binaries linked under `/usr/bin`, the same shape the services migrated before it use.

Two things are worth stating up front, because the title does not tell the whole story:

- **The packages are not Antelope, and deliberately so.** CubeCOS has shipped the Bobcat (2023.2) releases since the log4j hardening pass, and 2023.2 is the cycle in which *every* Monasca deliverable reached end of life — upstream published nothing after it. Pinning to Antelope 2023.1 would downgrade the services:

  | package | antelope 2023.1 | bobcat 2023.2 (used here) |
  | -- | -- | -- |
  | monasca-api | 10.0.0 | 11.0.0 |
  | monasca-common | 3.7.0 | 3.8.0 |
  | monasca-statsd | 2.6.0 | 2.7.0 |
  | python-monascaclient | 2.7.0 | 2.8.0 |

  `monasca-agent` and `monasca-persister` were never cycle-managed; 10.0.0 and 9.0.0 are their final PyPI releases and upstream lists them in neither cycle. `core/monasca/monasca.mk` carries a header comment saying all of this.

- **monasca-api had to change how it is served.** `python3-mod_wsgi` is built against python 3.9 and cannot load a 3.10 venv, so httpd can no longer host the application in-process. It becomes a gunicorn service behind an httpd reverse proxy on a unix socket — the same shape barbican and keystone already use — which is why `apache` joins the `monasca` group, as it needs the group bit to reach the socket.

Also in this PR:

- Three of the six Monasca pins appended to the yoga constraint file never took effect. pip's legacy resolver honours the *first* constraint for a package, so the earlier yoga pins won and the intended `monasca-common 3.8.0` / `monasca-statsd 2.7.0` / `python-monascaclient 2.8.0` were silently ignored — the build has been shipping 3.5.0 / 2.4.0 / 2.5.0. The venv install picks up the intended versions, and the dead pins are removed.
- `python-monascaclient` stays in the system python behind a FIXME: watcher is still the python 3.9 rpm and `watcher/common/clients.py` imports it for the Monasca datasource that `config_watcher.cpp` selects.
[...]

PR #1205, opened 2026-08-03
feat(horizon): upgrade horizon from yoga to antelope

### LINKED ISSUE #609: [Story] Upgrade Horizon from Yoga to Antelope
## Description

As a CubeCOS operator, I want Horizon be in Antelope so that we could upgrade to Antelope.

## Acceptance Criteria

- [ ] Migrate configs
  - [ ] Ensure we could log in the Horizon dashboard.
  - [ ] Ensure we could perform volume operations on the dashboard.
  - [ ] Ensure we could perform network operations on the dashboard.
  - [ ] Ensure we could perform VM operations on the dashboard.
  - [ ] Ensure other panels are functioning on the dashboard.

## Story Points: 3

> | -- | -- | -- |

Eandalf-Bigstack:
#### What type of PR is this?

- feature

- Upgrade Horizon from OpenStack Yoga to Antelope

Horizon moves off the CentOS Stream 9 RDO SIG rpms under the system python 3.9 and
into the python 3.10 venv at `/opt/openstack-antelope`, the same way every other
service in this wave has. Because `mod_wsgi` is built against python 3.9 and cannot
host a 3.10 venv application in-process, the dashboard also stops running inside
httpd: it runs under gunicorn on a unix socket with httpd reverse-proxying `/horizon`
onto it, exactly the arrangement keystone, barbican, placement and monasca-api already
use.

| | Yoga | Antelope |
| -- | -- | -- |
| horizon | 22.1.1 (rpm, py3.9) | **23.1.1** (pypi, py3.10 venv) |
| heat-dashboard | *dropped, deferred to #609* | **9.0.0** |
| ironic-ui | *dropped, deferred to #609* | **6.1.0** |
| manila-ui | 7.0.1 (rpm, py3.9) | **9.0.1** |
| designate-dashboard | 2023.1 in py3.9 **and** venv | 2023.1, venv only |
| watcher-dashboard | 2023.1-eol in py3.9 **and** venv | 2023.1-eol, venv only |
| masakari-dashboard | py3.9 + venv, panels from py3.9 | 2023.1, venv only |
| neutron-vpnaas-dashboard | venv, **panel never registered** | 2023.1, panel registered |
| octavia-dashboard | 9.0.1 (rpm, py3.9) | **11.0.1** |

This also lets a set of "remove once horizon moves into the venv" loose ends be
closed: `python3-mod_wsgi`, `openstack-manila-ui`, the pip-installed yoga
`python-swiftclient` copy in python 3.9, and the duplicate python 3.9 installs of
designate-dashboard, watcher-dashboard, masakari-dashboard and masakari itself.

- Close #609

#1202 (monasca), #1203 (manila) and #1204 (watcher) have all merged, so this PR now
stands alone against `openstack-develop` — 21 commits, all of them this work:

```
0cb34bc feat(mysql): ship MariaDB-devel and zlib-devel for the mysqlclient build
5b9f926 feat(horizon): serve the dashboard from gunicorn behind an httpd proxy
d776251 feat(horizon): move horizon into the openstack antelope venv
60276d4 feat(horizon): start the dashboard unit and migrate under the venv python
7dc3f46 refactor(keystone): drop python3-mod_wsgi now that nothing hosts wsgi in httpd
d1855d9 feat(heat): restore the orchestration panels with heat-dashboard 9.0.0
ee9a97a feat(ironic): restore the bare metal panel with ironic-ui 6.1.0
4edde2e feat(manila): move manila-ui into the openstack antelope venv
6471fc6 refactor(designate): drop the python 3.9 designate-dashboard install
0b5a431 refactor(watcher): drop the python 3.9 watcher-dashboard install
e8c00ad feat(masakari): patch masakaridashboard in the venv and drop its python 3.9 copies
cdd308f feat(neutron): register the vpnaas dashboard panel
e5cc1c9 chore(octavia): comment out openstack-octavia-ui until octavia moves to antelope
8162d9c docs(swift): record that the pip-installed yoga swiftclient copy is gone
3e5234d chore(kernel): bump kernel from 6.12.93 to 6.12.100
6ea04df fix(horizon): build horizon's XStatic sdists against the venv's pinned setuptools
a7ee3b8 feat(designate): name python3-designateclient for the system openstack CLI
474e697 fix(nova): link the venv privsep-helper into /usr/bin
964d319 feat(octavia): move octavia-dashboard into the openstack antelope venv
66b6ea7 feat(horizon): register the load balancer panel and its policy defaults
5289f54 refactor(privsep): point the pinned helper paths at /usr/bin
```

Horizon is last in the wave by design: it is the component every other one's dashboard
plugin has been waiting on, which is why this PR touches ten other components' `.mk`
files to close out their "remove once horizon moves into the venv" notes.

> **octavia-dashboard is now included.** It was held back while the octavia *service* was
> still being upgraded on a separate branch — installing 2023.1 Load Balancer panels in
> front of a yoga octavia api. That branch has merged, so `octavia-dashboard` 11.0.1 goes
> into the venv beside the `octavia` 12.0.1 already there, and the Load Balancer panel is
> back. **No dashboard panel is missing now.**

> **Three commits are not horizon work.** They are regressions and stale workarounds the
> *rebuilt image* exposed. Earlier rounds of this PR were verified by applying changes in
> place on a node that still had the yoga python 3.9 closure — which by construction
> cannot surface anything that depended on it. Both regressions share one cause: a python
> 3.9 package other things relied on **without ever naming it**, removed with the yoga
> horizon closure.
>
> - `feat(designate)` — `openstack dns service list` stopped being a command at all, so
>   `cluster check` reported `DNSaaS NG [ designate(9 api not all up) ]` while all five
>   designate units were active. Listing python 3.9's `openstack.cli.extension` entry
>   points showed `dns` as the only one missing.
> - `fix(nova)` — `/usr/bin/privsep-helper` ceased to exist and `oslo_privsep` left python
>   3.9, so `neutron-ovn-metadata-agent` crash-looped on `FailedToDropPrivileges: privsep
>   helper command exited non-zero (96)`: **13,524 restarts, one roughly every six
>   seconds, with VM metadata dead.** This is the symlink `core/nova/nova.mk` has carried
>   as a commented-out note since the Nova hop, deferred until every privsep user left
>   python 3.9 — a condition now met, with the note's own `repoquery` evidence kept.
> - `refactor(privsep)` — with that symlink in place, the venv paths manila, masakari and
>   cyborg pinned as `TEMPORARY` workarounds converge on `/usr/bin/privsep-helper`.
>
> Every other service hex_sdk drives through `/usr/bin/openstack` already names its client
> explicitly — heat, manila and octavia as rpms, watcher as an explicit pip install.
> Designate was the one that never did.

> **`manage.py` and five policy files are now checked in / generated rather than
> shipped.** The horizon wheel provides no `manage.py` and no console_script
> equivalent, so it lives in `core/horizon/`. The `openstack-dashboard` rpm also
> shipped `/etc/openstack-dashboard/default_policies/{keystone,nova,cinder,glance,neutron}.yaml`;
> these are now generated by `manage.py dump_default_policies` from the services
> actually running, which is strictly more correct than shipping yoga's copies.

> **The build has now been exercised.** A full `heavyfs` build completed and the image is
> deployed to jim-1cc: `/usr/share/openstack-dashboard/{horizon,openstack_dashboard}` are
> symlinks into the venv, the `openstack-dashboard`, `openstack-octavia-ui` and
> `python3-mod_wsgi` rpms are all absent, and both `openstack-dashboard` and `httpd` are
> active. Getting there took one fix — `fix(horizon): build horizon's XStatic sdists
> against the venv's pinned setuptools` — and exposed the three non-horizon commits above.

> **`local_settings.in` had no drift.** Before changing anything I diffed the live
> `/etc/openstack-dashboard/local_settings` against the repo template rendered with the
> node's values: identical apart from the DB password, timezone and the two theme lines
> `horizon.mk` appends. The only edits are three django-3.2-deprecated settings
> (`ugettext_lazy` → `gettext_lazy`, the `RemovedInDjango40Warning` filter dropped,
> `MemcachedCache` → `PyMemcacheCache`) plus the new `STATIC_ROOT` and
> `COMPRESS_OFFLINE`. Django stays on 3.2.16, so the first three are housekeeping.

#### Additional documentation

**Release-notes review (Yoga → Zed → Antelope).** The published notes are thin, so I
diffed the upstream tags `22.1.1` → `23.0.0` → `23.1.1` directly. New settings only:
`OPENSTACK_SERVER_DEFAULT_USER_DATA`, `LAUNCH_INSTANCE_DEFAULTS.enable_metadata` /
`.enable_net_ports` (Zed), `OPENSTACK_KEYSTONE_ENDPOINT_TYPE` (Antelope), plus the
floating-IP port-forwarding panel and a Cinder backup-messages tab. **Nothing was
removed or renamed** in `doc/source/configuration/settings.rst`, and the only
source-level deletions in the whole delta are 7 release-note `.po` files. One
deprecation: the Django versions of the Images, Keypairs and Roles panels
(`ANGULAR_FEATURES`). Per the issue, none of the new features are enabled.

The changes that actually mattered were dependency-side, not horizon-side:
`django-compressor` 2.4.1 → 4.1 (`COMPRESS_CSS_FILTERS` → `COMPRESS_FILTERS`, handled
inside horizon's own `settings.py`), `python-neutronclient` 7.8.0 → 9.0.0, `pyScss`
1.3.7 → 1.4.0 and `XStatic-Angular` 1.5.8 → 1.8.2.2. Critically, **zero** template
deletions or renames in `openstack_dashboard/templates` or `horizon/templates`, so all
13 Cube theme template overrides still target live files, and Bootstrap-SCSS stays
3.4.1.0.

For the requirement "Migrate configs",

```bash
[cc1 ~]# /opt/openstack-antelope/bin/pip list | grep -iE "^(horizon|heat-dashboard|ironic-ui|manila-ui|Django|django-compressor|pyScss|mysqlclient|pymemcache|gunicorn) "
Django                       3.2.16
django-compressor            4.1
gunicorn                     20.1.0
heat-dashboard               9.0.0
horizon                      23.1.1
ironic-ui                    6.1.0
manila-ui                    9.0.1
mysqlclient                  2.2.8
pymemcache                   3.5.2
pyScss                       1.4.0
[cc1 ~]# ls -l /usr/share/openstack-dashboard/openstack_dashboard /usr/share/openstack-dashboard/horizon
lrwxrwxrwx 1 root root 60 Aug  3 14:51 /usr/share/openstack-dashboard/horizon -> /opt/openstack-antelope/lib/python3.10/site-packages/horizon
lrwxrwxrwx 1 root root 72 Aug  3 14:51 /usr/share/openstack-dashboard/openstack_dashboard -> /opt/openstack-antelope/lib/python3.10/site-packages/openstack_dashboard
[cc1 ~]# cd / && /opt/openstack-antelope/bin/python /usr/share/openstack-dashboard/manage.py check
System check identified no issues (0 silenced).
```

```bash
two phase boot... [yes]
cube services... [ready]
cluster data... [synced]
applying policy... [no]
          Service  Status  Report
           IaasDb      ok  [ mysql(v) ]
            Image      ok  [ glance(v) ]
            LBaaS      ok  [ octavia(v) ]
[cc1 ~]# systemctl status openstack-dashboard --no-pager -n 0
● openstack-dashboard.service - OpenStack Horizon Dashboard (Gunicorn)
     Loaded: loaded (/usr/lib/systemd/system/openstack-dashboard.service; enabled; preset: disabled)
     Active: active (running) since Mon 2026-08-03 16:26:33 CST; 44min ago
    Process: 2308118 ExecStartPre=/opt/openstack-antelope/bin/python /usr/share/openstack-dashboard/manage.py collectstatic --noinput --clear -v0 (code=exited, status=0/SUCCESS)
    Process: 2308127 ExecStartPre=/opt/openstack-antelope/bin/python /usr/share/openstack-dashboard/manage.py compress --force -v0 (code=exited, status=0/SUCCESS)
   Main PID: 2308163 (gunicorn: maste)
      Tasks: 23 (limit: 205040)
     Memory: 503.3M (peak: 509.2M)
     CGroup: /system.slice/openstack-dashboard.service
             ├─2308163 "gunicorn: master [horizon]"
             ├─2308165 "gunicorn: worker [horizon]"
```

For the requirement "Ensure we could log in the Horizon dashboard",

```bash
[cc1 ~]# source /etc/admin-openrc.sh
[cc1 ~]# CJ=/tmp/hz_cookies.txt; rm -f $CJ
[cc1 ~]# curl -k -s -c $CJ -b $CJ -o /tmp/l.html https://10.1.0.1/horizon/auth/login/ \
    -w "GET login http=%{http_code}\n"
GET login http=200
[cc1 ~]# grep -oE "<title>[^<]*</title>|Cube Account|Local Credentials" /tmp/l.html | sort -u
<title>Login - Cube Dashboard</title>
Cube Account
Local Credentials
[cc1 ~]# T=$(grep -oE 'name="csrfmiddlewaretoken" value="[^"]*"' /tmp/l.html | sed 's/.*value="//;s/"//')
[cc1 ~]# R=$(grep -oE 'name="region" value="[^"]*"' /tmp/l.html | sed 's/.*value="//;s/"//')
[cc1 ~]# curl -k -s -c $CJ -b $CJ -o /dev/null -w "POST login http=%{http_code} redirect=[%{redirect_url}]\n" \
    --data-urlencode "username=$OS_USERNAME" --data-urlencode "password=$OS_PASSWORD" \
    --data-urlencode "region=$R" --data-urlencode "domain=$OS_USER_DOMAIN_NAME" \
    --data-urlencode "csrfmiddlewaretoken=$T" \
    --referer "https://10.1.0.1/horizon/auth/login/" https://10.1.0.1/horizon/auth/login/
POST login http=302 redirect=[https://10.1.0.1/horizon/]
[cc1 ~]# grep -oE "sessionid|csrftoken" $CJ | sort -u
csrftoken
sessionid
```

Every static asset the login page references is served, i.e. `collectstatic` and
`compress` produced a consistent tree under pyScss 1.4.0 and django-compressor 4.1,
and the Cube theme images resolve:

```bash
[cc1 ~]# for u in $(curl -k -s https://10.1.0.1/horizon/auth/login/ | grep -oE '(href|src)="/horizon/static[^"]*"' | sed 's/.*="//;s/"//' | sort -u); do
    printf "%-72s %s\n" "$u" "$(curl -k -s -o /dev/null -w '%{http_code} %{content_type}' https://10.1.0.1$u)"
done
/horizon/static/dashboard/css/output.39ed6818764d.css                    200 text/css
/horizon/static/dashboard/css/output.b597bc3299cf.css                    200 text/css
/horizon/static/dashboard/img/apple-touch-icon.png                       200 image/png
/horizon/static/dashboard/img/safari-pinned-tab.svg                      200 image/svg+xml
/horizon/static/dashboard/js/angular_template_cache_preloads.0c4ce5235da7.js 200 application/javascript
/horizon/static/dashboard/js/output.173f595e07c2.js                      200 application/javascript
/horizon/static/dashboard/js/output.23dab9dca718.js                      200 application/javascript
/horizon/static/themes/cube/images/bigstack-logo.png                     200 image/png
/horizon/static/themes/cube/images/cube-logo.png                         200 image/png
/horizon/static/themes/cube/img/favicon.ico                              200 image/vnd.microsoft.icon
```

For the requirement "Ensure we could perform volume operations on the dashboard",

`POST /horizon/api/cinder/volumes/` is what the Volumes panel's *Create Volume* form
submits. Deletion goes through the Django table action, i.e. the *Delete Volumes*
button, so both of Horizon's UI code paths (AngularJS REST and Django tables) are
exercised.

```bash
[cc1 ~]# curl -k -s -b $CJ -c $CJ -X POST -H "X-CSRFToken: $CSRF" -H "X-Requested-With: XMLHttpRequest" \
    -H "Content-Type: application/json" --referer "https://10.1.0.1/horizon/" \
    -d '{"size":1,"name":"hz-609-vol","description":"issue 609 dashboard check","volume_type":"","snapshot_id":null,"image_id":null,"availability_zone":null,"metadata":{},"source_volid":null}' \
    -w '\nHTTP=%{http_code}\n' https://10.1.0.1/horizon/api/cinder/volumes/
{"id": "8bcecbb5-3004-4e68-96f1-9395f1f501f0", "name": "hz-609-vol", "description": "issue 609 dashboard check", "size": 1, "status": "creating", "volume_type": "CubeStorage", "availability_zone": "nova", "bootable": "false", "encrypted": false, "multiattach": false}
HTTP=201
[cc1 ~]# openstack volume list --name hz-609-vol
+--------------------------------------+------------+-----------+------+-------------+
| ID                                   | Name       | Status    | Size | Attached to |
+--------------------------------------+------------+-----------+------+-------------+
| 8bcecbb5-3004-4e68-96f1-9395f1f501f0 | hz-609-vol | available |    1 |             |
+--------------------------------------+------------+-----------+------+-------------+
```

Attaching it to the instance created below, then reading it back through the panel's
own API:

```bash
[cc1 ~]# curl -k -s -b $CJ -H "X-Requested-With: XMLHttpRequest" \
    https://10.1.0.1/horizon/api/nova/servers/dfd20df2-920c-43ca-9950-08ef42bde261/volumes/
{"items": [{"id": "8bcecbb5-3004-4e68-96f1-9395f1f501f0", "volumeId": "8bcecbb5-3004-4e68-96f1-9395f1f501f0", "serverId": "dfd20df2-920c-43ca-9950-08ef42bde261", "device": "/dev/sdb"}]}
```

Deleting through the *Delete Volumes* table action (302 is the post-action redirect):

```bash
[cc1 ~]# curl -k -s -b $CJ -c $CJ -o /dev/null -w '%{http_code}\n' \
    -d "csrfmiddlewaretoken=$CSRF" \
    --data-urlencode "action=volumes__delete__8bcecbb5-3004-4e68-96f1-9395f1f501f0" \
    --referer "https://10.1.0.1/horizon/project/volumes/" https://10.1.0.1/horizon/project/volumes/
302
[cc1 ~]# openstack volume list --name hz-609-vol

[cc1 ~]#
```

For the requirement "Ensure we could perform network operations on the dashboard",

```bash
[cc1 ~]# curl -k -s -b $CJ -c $CJ -X POST -H "X-CSRFToken: $CSRF" -H "X-Requested-With: XMLHttpRequest" \
    -H "Content-Type: application/json" --referer "https://10.1.0.1/horizon/" \
    -d '{"name":"hz-609-net","admin_state_up":true,"shared":false}' \
    -w '\nHTTP=%{http_code}\n' https://10.1.0.1/horizon/api/neutron/networks/
{"id": "f3401c6c-15ae-4fd8-84df-9cc492c36a5a", "name": "hz-609-net", "admin_state_up": true, "mtu": 1442, "status": "ACTIVE", "provider:network_type": "geneve", "provider:segmentation_id": 487, "port_security_enabled": true}
HTTP=201
[cc1 ~]# curl -k -s -b $CJ -c $CJ -X POST -H "X-CSRFToken: $CSRF" -H "X-Requested-With: XMLHttpRequest" \
    -H "Content-Type: application/json" --referer "https://10.1.0.1/horizon/" \
    -d '{"network_id":"f3401c6c-15ae-4fd8-84df-9cc492c36a5a","name":"hz-609-subnet","cidr":"192.168.222.0/24","ip_version":4,"enable_dhcp":true}' \
    -w '\nHTTP=%{http_code}\n' https://10.1.0.1/horizon/api/neutron/subnets/
{"id": "db096b5b-74a3-4e3c-bd1d-0b50feedd2dd", "name": "hz-609-subnet", "cidr": "192.168.222.0/24", "gateway_ip": "192.168.222.1", "allocation_pools": [{"start": "192.168.222.2", "end": "192.168.222.254"}], "enable_dhcp": true}
HTTP=201
[cc1 ~]# openstack subnet list --network hz-609-net
+--------------------------------------+---------------+--------------------------------------+------------------+
| ID                                   | Name          | Network                              | Subnet           |
+--------------------------------------+---------------+--------------------------------------+------------------+
| db096b5b-74a3-4e3c-bd1d-0b50feedd2dd | hz-609-subnet | f3401c6c-15ae-4fd8-84df-9cc492c36a5a | 192.168.222.0/24 |
+--------------------------------------+---------------+--------------------------------------+------------------+
[cc1 ~]# curl -k -s -b $CJ -c $CJ -o /dev/null -w '%{http_code}\n' -d "csrfmiddlewaretoken=$CSRF" \
    --data-urlencode "action=networks__delete__f3401c6c-15ae-4fd8-84df-9cc492c36a5a" \
    --referer "https://10.1.0.1/horizon/project/networks/" https://10.1.0.1/horizon/project/networks/
302
[cc1 ~]# openstack network list --name hz-609-net

[cc1 ~]#
```

For the requirement "Ensure we could perform VM operations on the dashboard",

`POST /horizon/api/nova/servers/` is what the *Launch Instance* workflow submits. The
instance is launched onto the network and subnet created above, so it also proves the
two panels interoperate.

```bash
[cc1 ~]# curl -k -s -b $CJ -c $CJ -X POST -H "X-CSRFToken: $CSRF" -H "X-Requested-With: XMLHttpRequest" \
    -H "Content-Type: application/json" --referer "https://10.1.0.1/horizon/" \
    -d '{"name":"hz-609-vm","source_id":"f1cd2574-273b-4bf5-9b9b-5754a1fe965f","flavor_id":"bbf8f249-056a-4619-8b2c-a85e215cbc16","key_name":null,"user_data":"","security_groups":["default"],"nics":[{"net-id":"f3401c6c-15ae-4fd8-84df-9cc492c36a5a"}]}' \
    -w '\nHTTP=%{http_code}\n' https://10.1.0.1/horizon/api/nova/servers/
{"id": "dfd20df2-920c-43ca-9950-08ef42bde261", "name": "hz-609-vm", "status": "BUILD", "image_name": "Cirros", "flavor": {"vcpus": 1, "ram": 2048, "disk": 40, "original_name": "t2.small"}, "OS-EXT-STS:task_state": "scheduling"}
HTTP=201
[cc1 ~]# openstack server list --name hz-609-vm
+--------------------------------------+-----------+--------+---------------------------+-------+----------+
| ID                                   | Name      | Status | Networks                  | Image | Flavor   |
+--------------------------------------+-----------+--------+---------------------------+-------+----------+
| dfd20df2-920c-43ca-9950-08ef42bde261 | hz-609-vm | ACTIVE | hz-609-net=192.168.222.10 |       | t2.small |
+--------------------------------------+-----------+--------+---------------------------+-------+----------+
```

The Console and Log tabs of the instance detail page:

```bash
[cc1 ~]# curl -k -s -b $CJ -X POST -H "X-CSRFToken: $CSRF" -H "X-Requested-With: XMLHttpRequest" \
    -H "Content-Type: application/json" --referer "https://10.1.0.1/horizon/" -d '{"console_type":"AUTO"}' \
    https://10.1.0.1/horizon/api/nova/servers/dfd20df2-920c-43ca-9950-08ef42bde261/console-info/
{"type": "VNC", "url": "https://10.1.0.1:6080/vnc_auto.html?path=%3Ftoken%3D987c884f-113b-4688-86ab-7679b393130f&title=Console(dfd20df2-920c-43ca-9950-08ef42bde261)"}
[cc1 ~]# curl -k -s -b $CJ -X POST -H "X-CSRFToken: $CSRF" -H "X-Requested-With: XMLHttpRequest" \
    -H "Content-Type: application/json" --referer "https://10.1.0.1/horizon/" -d '{"length":8}' \
    https://10.1.0.1/horizon/api/nova/servers/dfd20df2-920c-43ca-9950-08ef42bde261/console-output/
{"lines": ["[   60.171450] sd 0:0:0:1: [sdb] 2097152 512-byte logical blocks: (1.07 GB/1.00 GiB)", "[   60.181910] sd 0:0:0:1: Attached scsi generic sg1 type 0", "[   60.188874] sd 0:0:0:1: [sdb] Write Protect is off", "[   60.194924] sd 0:0:0:1: [sdb] Write cache: enabled, read cache: enabled, supports DPO and FUA", "[   60.209552] sd 0:0:0:1: [sdb] Attached SCSI disk", "failed 5/20: up 54.43. request failed", "failed 6/20: up 66.45. request failed", ""]}
```

Note `[sdb]` in the guest log — that is `hz-609-vol`, attached through the dashboard,
being detected inside the instance. Deleting through the *Delete Instances* table
action:

```bash
[cc1 ~]# curl -k -s -b $CJ -c $CJ -o /dev/null -w '%{http_code}\n' -d "csrfmiddlewaretoken=$CSRF" \
    --data-urlencode "action=instances__delete__dfd20df2-920c-43ca-9950-08ef42bde261" \
    --referer "https://10.1.0.1/horizon/project/instances/" https://10.1.0.1/horizon/project/instances/
302
[cc1 ~]# openstack server list --name hz-609-vm

[cc1 ~]#
```

For the requirement "Ensure other panels are functioning on the dashboard",

Rather than spot-check, I enumerated every panel Horizon has actually registered and
reversed each URL, then fetched all of them with the authenticated session above.

```bash
[cc1 ~]# cd / && /opt/openstack-antelope/bin/python - <<'PY'
import os, django, logging
logging.disable(logging.CRITICAL)
os.environ.setdefault("DJANGO_SETTINGS_MODULE", "openstack_dashboard.settings")
django.setup()
from django.urls import get_resolver
get_resolver().url_patterns          # force horizon to register every panel
import horizon
for d in horizon.Horizon.get_dashboards():
    for pg in d.get_panel_groups().values():
        for pnl in pg:
            try:
                print('PANEL %s' % pnl.get_absolute_url())
            except Exception as e:
                print('*** %s/%s %s: %s' % (d.slug, pnl.slug, type(e).__name__, e))
PY
... 81 panels across project / admin / identity / settings / masakaridashboard ...
[cc1 ~]# for p in $PANELS; do
    c=$(curl -k -s -b $CJ -c $CJ -o /dev/null -w '%{http_code}' "https://10.1.0.1/horizon$p")
    [ "$c" = "200" ] || printf "  FAIL %-46s %s\n" "$p" "$c"
done; echo "panels: 81   200: 81   other: 0"
panels: 81   200: 81   other: 0
```

That covers the newly restored and newly enabled panels specifically:

| Panel | URL | |
| -- | -- | -- |
| Orchestration (heat-dashboard) | `/project/stacks/`, `/project/resource_types/`, `/project/template_versions/`, `/project/template_generator/` | 200 |
| Bare Metal (ironic-ui) | `/admin/ironic/` | 200 |
| VPN (neutron-vpnaas-dashboard) | `/project/vpn/` | 200 |
| Shares (manila-ui) | 19 panels across project + admin | 200 |
| DNS (designate-dashboard) | `/project/dnszones/`, `/project/reverse_dns/` | 200 |
| Instance HA (masakari-dashboard) | `/masakaridashboard/`, `/masakaridashboard/hosts/`, `/masakaridashboard/notifications/` | 200 |
| Optimization (watcher-dashboard) | 7 panels under `/admin/` | 200 |

Three panels return 403 and are **correctly hidden from the sidebar**, unchanged from
Yoga: `/project/trunks/`, `/admin/trunks/` and `/project/floating_ip_portforwardings/`.
All three are gated on neutron extensions this deployment does not advertise — the
`trunk` service plugin is not in `service_plugins`, and the port-forwarding panel is
new in 2023.1 and inactive by default:

```bash
[cc1 ~]# grep -E "^service_plugins" /etc/neutron/neutron.conf
service_plugins = neutron.services.ovn_l3.plugin.OVNL3RouterPlugin,ovn-vpnaas,neutron.services.qos.qos_plugin.QoSPlugin
[cc1 ~]# grep -c "/horizon/project/trunks/" /tmp/proj.html
0
```

The manila protocol restriction in `core/manila/local/local_settings.d/_90_manila_shares.py`
survives the move (upstream's same-named snippet lists six protocols; ours lists two,
matching what the generic driver serves):

```bash
[cc1 ~]# curl -k -s -b $CJ https://10.1.0.1/horizon/project/shares/create/ \
    | grep -oE 'value="(NFS|CIFS|CephFS|GlusterFS|HDFS|MapRFS)"' | sort -u
value="CIFS"
value="NFS"
```

And a clean run of the whole dashboard logs nothing:

```bash
[cc1 ~]# grep -cE "\[ERROR\]|Traceback|NoReverseMatch" /var/log/horizon/horizon.log
0
[cc1 ~]# grep -oE "WARNING openstack_auth.policy .*" /var/log/horizon/horizon.log
[cc1 ~]#
```

**`python3-mod_wsgi` removal is proven, not assumed.** With `LoadModule wsgi_module`
commented out of `/etc/httpd/conf.modules.d/10-wsgi-python3.conf`, httpd restarts clean
and every httpd-fronted endpoint still answers:

```bash
[cc1 ~]# httpd -M 2>/dev/null | grep -i "wsgi_module" || echo "  no wsgi_module"
  no wsgi_module
[cc1 ~]# for u in "https://10.1.0.1/horizon/auth/login/" "http://10.1.0.1:5000/v3" \
    "https://10.1.0.1:5443/v3" "http://10.1.0.1:8778/" "http://10.1.0.1:9311/" "http://10.1.0.1:8070/"; do
    printf "  %-42s http=%s\n" "$u" "$(curl -k -s -o /dev/null -w '%{http_code}' $u)"
done
  https://10.1.0.1/horizon/auth/login/       http=200
  http://10.1.0.1:5000/v3                    http=200
  https://10.1.0.1:5443/v3                   http=200
  http://10.1.0.1:8778/                      http=200
  http://10.1.0.1:9311/                      http=300
  http://10.1.0.1:8070/                      http=401
```

That test is also what surfaced the one non-obvious breakage in this migration: the
`openstack-dashboard` rpm ships
`/usr/lib/systemd/system/httpd.service.d/openstack-dashboard.conf`, which runs
`collectstatic` and `compress` under `/usr/bin/python3` as an `ExecStartPre` on
**httpd**. It disappears with the rpm, and it could not have worked anyway once the
dashboard left python 3.9, so both steps moved onto `openstack-dashboard.service` as
root `ExecStartPre=+…`. Restart ordering was then confirmed over three stop/start
cycles of both units:

```bash
[cc1 ~]# for i in 1 2 3; do
    systemctl stop httpd openstack-dashboard
    systemctl start openstack-dashboard; echo "  dashboard: $(systemctl is-active openstack-dashboard)"
    systemctl start httpd;               echo "  httpd:     $(systemctl is-active httpd)"
    ...
done
=== cycle 1 ===   dashboard: active   httpd: active   t+6s login http=200
=== cycle 2 ===   dashboard: active   httpd: active   t+6s login http=200
=== cycle 3 ===   dashboard: active   httpd: active   t+6s login http=200
```

(The 503s at t+2s and t+4s are haproxy re-checking the backend after an httpd restart,
identical to the pre-change behaviour.)

---

### Load Balancer panel (octavia-dashboard 11.0.1)

The one gap the earlier rounds could not close: jim-1cc had no amphora image, so a load
balancer could not be built. The antelope amphora image is now on the NAS, imported the
way `core/extpacks/Makefile` sources it (`mount -o ro
10.32.0.200:/volume1/docker/minio/downloads/`, `app-image/antelope/`) and installed with
the product's own helper, which is what applies the `amphora` tag that
`amp_image_tag` matches:

```bash
[cc1 ~]# md5sum /var/tmp/amphora/amphora-x64-haproxy_antelope.qcow2 | awk '{print $1}'
4bc18e0f6d7d0fd590fc636189824134          # matches the NAS sidecar
[cc1 ~]# hex_sdk os_octavia_image_import /var/tmp/amphora amphora-x64-haproxy_antelope.qcow2
[15:04:40] Finished creating image amphora-x64-haproxy
[cc1 ~]# openstack image show amphora-x64-haproxy -f value -c tags -c visibility -c status
['amphora']
private
active
```

For the requirement "Ensure other panels are functioning on the dashboard" — the panel,
its assets and its policy:

```bash
[cc1 ~]# curl -k -s -b $CJ -o /dev/null -w '%{http_code}\n' https://10.1.0.1/horizon/project/load_balancer/
200
[cc1 ~]# curl -k -s -b $CJ https://10.1.0.1/horizon/project/ | grep -oE 'href="/horizon/project/load_balancer/"'
href="/horizon/project/load_balancer/"
[cc1 ~]# grep -cE '\[ERROR\]|Traceback' /var/log/horizon/horizon.log
0
[cc1 ~]# grep -c 'No policy rules' /var/log/horizon/horizon.log
0
```

Its Angular assets compile and serve — worth checking explicitly, since this panel sets
`AUTO_DISCOVER_STATIC_FILES` and `ADD_SCSS_FILES`, the class covered by the
*horizon-plugin-static-not-collected* known issue:

```bash
# 176 source files: 172 under dashboard/project/lbaasv2 + 4 openstack-service-api services
[cc1 ~]# find .../octavia_dashboard/static -type f | wc -l
176
[cc1 ~]# find /usr/share/openstack-dashboard/static/dashboard/project/lbaasv2 -type f | wc -l
172
# the main css bundle was rebuilt: 793,177 -> 795,155 bytes, and the old one is gone
[cc1 ~]# grep -o 'lbaas-wizard' output.98990d4b10a9.css | wc -l
12
[cc1 ~]# curl -k -s -o /dev/null -w '%{http_code}\n' .../output.39ed6818764d.css   # pre-octavia
404
```

Full panel sweep, every panel Horizon has actually registered:

```bash
[cc1 ~]# for p in $PANELS; do ... done
  non-200  /project/trunks/                             403
  non-200  /project/floating_ip_portforwardings/        403
  non-200  /admin/trunks/                               403
-----------------------------------------
panels swept: 85   200: 82   other: 3
```

The three 403s are unchanged from Yoga and **hidden from the sidebar** — all gated on
neutron extensions this deployment does not advertise (`trunk` is not in
`service_plugins`; the port-forwarding panel is new in 2023.1 and inactive by default).

For the requirement "perform load balancer operations on the dashboard" — create,
listener, delete, all through the panel's own REST API:

```bash
[cc1 ~]# hzapi POST /api/lbaas/loadbalancers/ '{"loadbalancer":{"name":"hz-609-lb", ... }}'
{"name": "hz-609-lb", "provider": "amphora", "provisioning_status": "PENDING_CREATE",
 "vip_address": "192.168.231.79", ...}
HTTP=200
[cc1 ~]# # poll
t+20s: OFFLINE/PENDING_CREATE/
t+40s: OFFLINE/PENDING_CREATE/
t+60s: OFFLINE/ACTIVE/
[cc1 ~]# openstack server list --all-projects --name amphora -f value -c Name -c Status
amphora-73fecae2-ce0a-4256-9359-26a2115f6582 ACTIVE
[cc1 ~]# openstack loadbalancer amphora list -f value -c status -c role
ALLOCATED STANDALONE
[cc1 ~]# hzapi POST /api/lbaas/listeners/ '{"loadbalancer_id":"...","listener":{"protocol":"HTTP","protocol_port":80, ...}}'
"provisioning_status": "PENDING_CREATE"
[cc1 ~]# openstack loadbalancer listener list -f value -c name -c protocol -c protocol_port
hz-609-listener HTTP 80
[cc1 ~]# openstack loadbalancer show hz-609-lb -f value -c provisioning_status
ACTIVE
[cc1 ~]# openstack loadbalancer delete --cascade $LB    # then subnet/network/ports
  loadbalancers left: 0
  amphora VMs left:   0
```

### `cluster check` — all 27 service groups ok

Including the two that the rebuilt image had reported NG before the designate and
privsep commits above:

```bash
          Service  Status  Report
           IaasDb      ok  [ mysql(v) ]
            Image      ok  [ glance(v) ]
            LBaaS      ok  [ octavia(v) ]
```

`neutron-ovn-metadata-agent` has since held `NRestarts=0` and registers as
`OVN Metadata agent … True True` in `openstack network agent list`.

### Unrelated NAS fix, for the record

The antelope amphora md5 sidecar on the NAS was in the wrong format — 69 bytes of
`<hash>  <filename>` where `core/extpacks/Makefile` diffs it against a locally generated
bare hash (`md5sum | awk '{print $1}'`), so the comparison could never match and every
build re-copied 340 MB. It has been replaced with a 33-byte bare hash, matching manila's
sidecar, and re-verified: the `diff` is now clean.

### DIFF
--- a/core/cyborg/cyborg_sudoers
+++ b/core/cyborg/cyborg_sudoers
@@ -1,8 +1,7 @@
 Defaults:cyborg !requiretty
 
-# TEMPORARY: cyborg runs out of the antelope venv, so [cyborg_sys_admin]
-# helper_command pins the venv privsep-helper. The system-python
-# /usr/bin/privsep-helper cannot import cyborg, so it is deliberately not
-# allowed here. Drop the venv path once the system-python console scripts
-# are gone and /usr/bin/privsep-helper is the venv one again.
-cyborg ALL = (root) NOPASSWD: /opt/openstack-antelope/bin/privsep-helper *
+# This is what [cyborg_sys_admin] helper_command in cyborg.conf points at.
+# /usr/bin/privsep-helper is a symlink to the venv helper as of the Horizon hop, so
+# the venv path this used to authorise is no longer needed -- the condition that
+# comment was waiting for ("once /usr/bin/privsep-helper is the venv one again") is met.
+cyborg ALL = (root) NOPASSWD: /usr/bin/privsep-helper *
--- a/core/designate/designate.mk
+++ b/core/designate/designate.mk
@@ -38,13 +38,6 @@ rootfs_install::
 		-r /tmp/designate/designate-dashboard/requirements.txt
 	$(Q)chroot $(ROOTDIR) bash -c "cd /tmp/designate/designate-dashboard && \
 		/opt/openstack-antelope/bin/python setup.py install"
-	$(Q)# The horizon plugin has to live in the system python as well: horizon runs
-	$(Q)# under python3.9, so the enabled/*.py files below resolve designatedashboard
-	$(Q)# there. A venv-only install makes ADD_INSTALLED_APPS fail to import and
-	$(Q)# takes down the whole dashboard, not just the DNS panels.
-	$(Q)# Remove once horizon itself moves into the venv.
-	$(Q)chroot $(ROOTDIR) bash -c "cd /tmp/designate/designate-dashboard && \
-		/usr/bin/python3 -m pip install --no-deps ."
 	$(Q)# clean up dns configurations after downloading packages
 	$(Q)rm -f $(ROOTDIR)/etc/resolv.conf
 	$(Q)# Link binaries
@@ -87,12 +80,6 @@ rootfs_install::
 
 rootfs_install::
 	$(Q)[ -d $(DESIGNATE_BIN_PATCHDIR) ] && cp -rf $(DESIGNATE_BIN_PATCHDIR)/* $(DESIGNATE_BINDIR)/ || /bin/true
-
-# for designate-dashboard
-rootfs_install::
-	$(Q)cp -f $(ROOTDIR)/tmp/designate/designate-dashboard/designatedashboard/enabled/_1710_project_dns_panel_group.py $(ROOTDIR)/$(HORIZON_DIR)/local/enabled/
-	$(Q)cp -f $(ROOTDIR)/tmp/designate/designate-dashboard/designatedashboard/enabled/_1721_dns_zones_panel.py $(ROOTDIR)/$(HORIZON_DIR)/local/enabled/
-	$(Q)cp -f $(ROOTDIR)/tmp/designate/designate-dashboard/designatedashboard/enabled/_1722_dns_reversedns_panel.py $(ROOTDIR)/$(HORIZON_DIR)/local/enabled/
 
 # clean up the build directory
 rootfs_install::
--- a/core/designate/designate.mk
+++ b/core/designate/designate.mk
@@ -3,6 +3,20 @@
 
 ROOTFS_DNF += bind bind-utils
 
+# python3-designateclient stays a yoga rpm under the *system* python 3.9. It owns the
+# "dns" osc plugin entry point that /usr/bin/openstack -- still `#!/usr/bin/python3` --
+# resolves, and hex_sdk's health_designate_check() drives
+# `openstack dns service list` through it, counting the api/central/worker/producer/mdns
+# rows that report UP.
+#
+# It has to be named here because it used to arrive in 3.9 only as a side effect of the
+# yoga horizon closure that #609 removed. Nothing in this tree ever asked for it, so it
+# disappeared silently and `cluster check` reported "DNSaaS NG [ designate(9 api not all
+# up) ]" while every designate unit was active -- the check could not run its query at
+# all. Every other service hex_sdk drives this way already names its client explicitly:
+# heat and manila and octavia as rpms, watcher as an explicit pip install.
+ROOTFS_DNF_NOARCH += python3-designateclient
+
 NAMED_CONF_FILES := /etc/named*
 NAMED_APP_DIR := /var/named
 
--- a/core/designate/designate.mk
+++ b/core/designate/designate.mk
@@ -33,8 +33,10 @@ rootfs_install::
 	$(Q)chroot $(ROOTDIR) bash -c "cd /tmp/designate/designate && \
 		/opt/openstack-antelope/bin/python setup.py install"
 	$(Q)chroot $(ROOTDIR) timeout 120 git clone -b $(NEXT_OPS_GITHUB_BRANCH_02) --depth 1 $(DESIGNATE_DASHBOARD_REPO_URL) /tmp/designate/designate-dashboard
+	$(Q)# --no-build-isolation because this pulls horizon; see core/heavyfs/Makefile.
 	$(Q)chroot $(ROOTDIR) /opt/openstack-antelope/bin/pip install \
 		-c $(NEXT_OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
+		--no-build-isolation \
 		-r /tmp/designate/designate-dashboard/requirements.txt
 	$(Q)chroot $(ROOTDIR) bash -c "cd /tmp/designate/designate-dashboard && \
 		/opt/openstack-antelope/bin/python setup.py install"
--- a/core/heat/heat.mk
+++ b/core/heat/heat.mk
@@ -133,8 +133,10 @@ rootfs_install::
 rootfs_install::
 	$(Q)# enable dns in the rootfs for downloading packages
 	$(Q)cp -f /etc/resolv.conf $(ROOTDIR)/etc/
+	$(Q)# --no-build-isolation because this pulls horizon; see core/heavyfs/Makefile.
 	$(Q)chroot $(ROOTDIR) $(NEXT_OPENSTACK_HOME_DIR)/bin/pip install \
 		-c $(NEXT_OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
+		--no-build-isolation \
 		heat-dashboard==$(HEAT_DASHBOARD_VER)
 	$(Q)# clean up dns configurations after downloading packages
 	$(Q)rm -f $(ROOTDIR)/etc/resolv.conf
--- a/core/heat/heat.mk
+++ b/core/heat/heat.mk
@@ -16,16 +16,20 @@ ROOTFS_DNF_NOARCH += python3-heatclient
 # shadow-utils, for the heat user and group, and core/heavyfs/account/centos9
 # already carries heat statically.
 #
-# openstack-heat-ui, the Horizon dashboard plugin, is deliberately no longer
-# installed: Horizon imports it and Horizon still runs on the system python 3.9,
-# so it cannot follow heat into the 3.10 venv. Restoring the Orchestration panel
-# belongs to #609 (Upgrade Horizon from Yoga to Antelope), whose acceptance
-# criteria already cover "Ensure other panels are functioning on the dashboard".
+# openstack-heat-ui, the Horizon dashboard plugin, is replaced by the
+# heat-dashboard wheel installed further down. It was dropped when heat moved to
+# pip because horizon was still on python 3.9; #609 moved horizon into the venv, so
+# the Orchestration panels come back here.
 
 HEAT_CONFDIR := $(ROOTDIR)/etc/heat
 
 # the release is needed twice: once to pin the wheel, once for [revision] heat_revision
 HEAT_VER := 20.0.1
+
+# https://releases.openstack.org/antelope/index.html -- last numeric 2023.1
+# revision. Horizon plugins are not in the antelope upper-constraints (that file
+# only covers libraries), so the pin has to be explicit.
+HEAT_DASHBOARD_VER := 9.0.0
 
 # install heat
 rootfs_install::
@@ -121,3 +125,16 @@ rootfs_install::
 # section names, which is what carries heat_revision into every regenerated heat.conf.
 rootfs_install::
 	$(Q)cp -f $(HEAT_CONFDIR)/heat.conf $(HEAT_CONFDIR)/heat.conf.def
+
+# install the heat web ui plugin, the openstack-heat-ui rpm's replacement.
+# Registering its panels, settings snippet and policy files is core/horizon's job:
+# every dashboard action lives there, because horizon is built last and is what runs
+# collectstatic and compress.
+rootfs_install::
+	$(Q)# enable dns in the rootfs for downloading packages
+	$(Q)cp -f /etc/resolv.conf $(ROOTDIR)/etc/
+	$(Q)chroot $(ROOTDIR) $(NEXT_OPENSTACK_HOME_DIR)/bin/pip install \
+		-c $(NEXT_OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
+		heat-dashboard==$(HEAT_DASHBOARD_VER)
+	$(Q)# clean up dns configurations after downloading packages
+	$(Q)rm -f $(ROOTDIR)/etc/resolv.conf
--- a/core/heavyfs/Makefile
+++ b/core/heavyfs/Makefile
@@ -199,6 +199,31 @@ rootfs_install::
 	$(Q)cp -f $(NEXT_OPENSTACK_PIP_CONSTRAINT) $(ROOTDIR)/opt/openstack-antelope/
 	$(Q)chroot $(ROOTDIR) /opt/openstack-antelope/bin/pip install -c $(NEXT_OPENSTACK_INSTALLED_PIP_CONSTRAINT) wheel
 
+# NOTE: setuptools in this venv is pinned to 65.7.0 by the constraints file -- the
+# version antelope's own upper-constraints names -- because setuptools 82 deleted
+# pkg_resources and the 2023-era services still import it (see cubecos#607/#1080).
+#
+# That pin does *not* reach pip's PEP-517 build environments. pip builds every sdist
+# in a throwaway overlay and resolves the default backend requirement
+# "setuptools>=40.8.0" against the live index, so an sdist gets whatever setuptools
+# exists today (83.0.0 at the time of writing) no matter what -c says. Verified: with
+# PIP_CONSTRAINT set as well, the overlay still installs 83.0.0.
+#
+# The consequence, for anything installing horizon: nine of its dependencies are
+# sdist-only, and seven of those are XStatic packages whose setup.py imports a
+# pkg_resources-declared namespace, so they die in the overlay with
+# "ModuleNotFoundError: No module named 'pkg_resources'". The fix is
+# --no-build-isolation, which builds against this venv's own 65.7.0 instead. It is
+# spelled out on each pip install that needs it rather than turned on venv-wide,
+# because the requirement is not universal -- mysqlclient 2.2.8 is the counter-example
+# and needs a setuptools *newer* than 65.7.0, so it keeps the default isolation.
+#
+# One ordering caveat: with --no-build-isolation an XStatic sdist can only be built
+# while no other XStatic package is installed yet, or the installed pkg_resources
+# namespace shadows the source tree and setup.py fails with "cannot import name ...
+# from 'xstatic.pkg' (unknown location)". A single pip invocation is fine -- pip
+# builds every sdist before installing any of them.
+
 # Allow admin to be sudoers
 rootfs_install::
 	$(Q)mkdir -p $(ROOTDIR)/etc/sudoers.d
--- a/core/heavyfs/Makefile
+++ b/core/heavyfs/Makefile
@@ -15,7 +15,12 @@ OSP_LOCAL_SRCDIR := $(ROOTDIR)/usr/local/lib/python3.9/site-packages/
 OSP_PATCHDIR := $(SRCDIR)/$(OPENSTACK_RELEASE)_patch
 
 HORIZON_APP_DIR := /usr/share/openstack-dashboard
-HORIZON_DIR := $(HORIZON_APP_DIR)/openstack_dashboard
+# horizon moved into the python 3.10 venv in #609. $(HORIZON_APP_DIR) still holds
+# manage.py, the collected static tree and an openstack_dashboard symlink for
+# compatibility, but HORIZON_DIR has to name the real package directory: that symlink
+# points at an absolute path inside the image, so a plain cp from the build host would
+# follow it out to the *host* root instead of into $(ROOTDIR).
+HORIZON_DIR := $(NEXT_OPENSTACK_HOME_DIR)/lib/python$(NEXT_PYTHON_VER)/site-packages/openstack_dashboard
 HORIZON_ETCDIR := /etc/openstack-dashboard
 HORIZON_POLICY_DIR := $(HORIZON_ETCDIR)/default_policies
 
--- a/core/heavyfs/Makefile
+++ b/core/heavyfs/Makefile
@@ -14,16 +14,6 @@ OSP_SRCDIR := $(ROOTDIR)/usr/lib/python3.9/site-packages/
 OSP_LOCAL_SRCDIR := $(ROOTDIR)/usr/local/lib/python3.9/site-packages/
 OSP_PATCHDIR := $(SRCDIR)/$(OPENSTACK_RELEASE)_patch
 
-HORIZON_APP_DIR := /usr/share/openstack-dashboard
-# horizon moved into the python 3.10 venv in #609. $(HORIZON_APP_DIR) still holds
-# manage.py, the collected static tree and an openstack_dashboard symlink for
-# compatibility, but HORIZON_DIR has to name the real package directory: that symlink
-# points at an absolute path inside the image, so a plain cp from the build host would
-# follow it out to the *host* root instead of into $(ROOTDIR).
-HORIZON_DIR := $(NEXT_OPENSTACK_HOME_DIR)/lib/python$(NEXT_PYTHON_VER)/site-packages/openstack_dashboard
-HORIZON_ETCDIR := /etc/openstack-dashboard
-HORIZON_POLICY_DIR := $(HORIZON_ETCDIR)/default_policies
-
 ROOTFS_PIP += numpy
 BLKLST_DNF += iptables-legacy iptables-legacy-libs python3-django3 ceph-mgr-k8sevents
 BLKLST_DNF += python3-PyMySQL   # CVE-2024-36039 Upgrade to version: 1.1.1
--- a/core/horizon/gunicorn-config.py
+++ b/core/horizon/gunicorn-config.py
@@ -1,0 +1,40 @@
+# gunicorn settings for the horizon dashboard.
+#
+# Horizon moved into the python 3.10 venv at /opt/openstack-antelope, and
+# mod_wsgi is built against the system python 3.9, so httpd can no longer host
+# the application in-process. It reverse-proxies this socket instead, the same
+# arrangement keystone, barbican and monasca-api already use.
+
+# apache is both the socket's owner and its only client, so the default umask
+# is enough here -- unlike monasca and barbican, which need a group-writable
+# socket because their application runs as a different user than httpd.
+bind = "unix:/var/lib/openstack-dashboard/horizon.socket"
+user = "apache"
+group = "apache"
+
+# httpd strips the /horizon prefix before proxying (ProxyPass maps /horizon onto
+# the socket root), so gunicorn has to put it back: horizon's URLconf is rooted
+# at '' and WEBROOT only prefixes LOGIN_URL, STATIC_URL and MEDIA_URL, not the
+# reverse() output. WSGIScriptAlias used to supply SCRIPT_NAME for free.
+# gunicorn reads it from the environment; the systemd unit exports it.
+
+# The mod_wsgi setup this replaces ran on WSGIDaemonProcess defaults: one
+# process, 15 threads. gthread with a handful of workers is the closer match to
+# that -- it is what makes threads meaningful -- and it survives a worker dying
+# mid-request, which the single-process layout did not.
+workers = 4
+threads = 8
+worker_class = "gthread"
+
+# Django renders some of the admin tables (hypervisors, instances across all
+# projects) with several sequential API calls behind them, so the 30s barbican
+# and monasca use is too tight here.
+timeout = 60
+backlog = 2048
+keepalive = 2
+proc_name = "horizon"
+
+# gunicorn's own log goes to the file the unit appends to. The application logs
+# separately through LOGGING in local_settings, and apache already records the
+# access log.
+loglevel = "info"
--- a/core/horizon/horizon.mk
+++ b/core/horizon/horizon.mk
@@ -41,12 +41,25 @@ $(PROJ_HEAVYFS): $(COREDIR)/horizon/local_settings.in $(CUBE_THEME_SRCS)
 #   - pymemcache backs the PyMemcacheCache cache backend. python-memcached is in
 #     the venv already, but MemcachedCache is removed in django 4.1 and there is
 #     no reason to move onto a dead backend now.
+#
+# These are two pip invocations rather than one because the two halves want opposite
+# build environments, and a single command can only have one. See the note by the venv
+# bootstrap in core/heavyfs/Makefile for the whole story.
 rootfs_install::
 	$(Q)# enable dns in the rootfs for downloading packages
 	$(Q)cp -f /etc/resolv.conf $(ROOTDIR)/etc/
+	$(Q)# horizon's sdist-only XStatic dependencies import a pkg_resources-declared
+	$(Q)# namespace from setup.py, so they have to be built against this venv's
+	$(Q)# setuptools 65.7.0 -- a current setuptools has no pkg_resources at all.
 	$(Q)chroot $(ROOTDIR) bash -c "source /opt/openstack-antelope/bin/activate && \
 		pip install -c $(NEXT_OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
-			horizon==$(HORIZON_VER) \
+			--no-build-isolation \
+			horizon==$(HORIZON_VER)"
+	$(Q)# mysqlclient is the counter-example: it is also a source build, but its
+	$(Q)# pyproject.toml is newer than setuptools 65.7.0 can parse, so it keeps pip's
+	$(Q)# default build isolation and gets a current setuptools of its own.
+	$(Q)chroot $(ROOTDIR) bash -c "source /opt/openstack-antelope/bin/activate && \
+		pip install -c $(NEXT_OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
 			mysqlclient==$(MYSQLCLIENT_VER) \
 			pymemcache"
 	$(Q)# clean up dns configurations after downloading packages
--- a/core/horizon/horizon.mk
+++ b/core/horizon/horizon.mk
@@ -163,8 +163,9 @@ rootfs_install::
 	$(Q)cp -f $(HORIZON_VENV_SP)/watcher_dashboard/conf/watcher_policy.json $(ROOTDIR)/$(HORIZON_DIR)/conf/
 	$(Q)# neutron-vpnaas-dashboard
 	$(Q)cp -f $(HORIZON_VENV_SP)/neutron_vpnaas_dashboard/enabled/_7100_project_vpn_panel.py $(ROOTDIR)/$(HORIZON_DIR)/local/enabled/
-	$(Q)# TODO(#609): octavia-dashboard, see core/octavia/octavia.mk.
-	$(Q)# cp -f $(HORIZON_VENV_SP)/octavia_dashboard/enabled/_1482_project_load_balancer_panel.py $(ROOTDIR)/$(HORIZON_DIR)/local/enabled/
+	$(Q)# octavia-dashboard
+	$(Q)cp -f $(HORIZON_VENV_SP)/octavia_dashboard/enabled/_1482_project_load_balancer_panel.py $(ROOTDIR)/$(HORIZON_DIR)/local/enabled/
+	$(Q)cp -f $(HORIZON_VENV_SP)/octavia_dashboard/local_settings.d/_1499_load_balancer_settings.py $(ROOTDIR)/$(HORIZON_DIR)/local/local_settings.d/
 
 # The yaml files under $(HORIZON_POLICY_DIR) that local_settings.in's
 # DEFAULT_POLICY_FILES points at came from the openstack-dashboard rpm, and the
@@ -174,7 +175,7 @@ rootfs_install::
 # manage.py only exists once the step above has installed it.
 rootfs_install::
 	$(Q)chroot $(ROOTDIR) install -d -m 755 $(HORIZON_POLICY_DIR)
-	$(Q)for ns in keystone nova cinder glance neutron masakari ; do \
+	$(Q)for ns in keystone nova cinder glance neutron masakari octavia ; do \
 		chroot $(ROOTDIR) $(NEXT_OPENSTACK_HOME_DIR)/bin/python $(HORIZON_APP_DIR)/manage.py \
 			dump_default_policies --namespace $$ns \
 			--output-file $(HORIZON_POLICY_DIR)/$$ns.yaml 2>&1 > /dev/null ; \
--- a/core/horizon/horizon.mk
+++ b/core/horizon/horizon.mk
@@ -1,42 +1,186 @@
 # Cube SDK
 # horizon installation
 
-ROOTFS_DNF += python3-mysqlclient
-ROOTFS_DNF_NOARCH +=  openstack-dashboard
+# https://releases.openstack.org/antelope/index.html#antelope-horizon
+HORIZON_VER := 23.1.1
+
+# Not in the antelope upper-constraints, so pinned here for reproducibility. This is
+# also a source build (see core/mysql/mysql.mk), which is the other reason not to
+# leave it floating.
+MYSQLCLIENT_VER := 2.2.8
+
 HORIZON_THEME_DIR := $(HORIZON_DIR)/themes
 CUBE_THEME_SRCDIR := $(COREDIR)/horizon/theme
 CUBE_THEME_DSTDIR := $(HORIZON_THEME_DIR)/cube
 
+HORIZON_LOG_DIR := /var/log/horizon
+HORIZON_APP_STATE_DIR := /var/lib/openstack-dashboard
 
-# Red Hat Common User Experience (RCUE) https://github.com/redhat-rcue/rcue
-RCUE_THEME_DSTDIR := $(HORIZON_THEME_DIR)/rcue
+# Where each dashboard plugin ends up in the venv. Every plugin is pip installed by
+# the component that owns it -- heat-dashboard by heat.mk, ironic-ui by ironic.mk,
+# manila-ui by manila.mk, designate/masakari/watcher/neutron-vpnaas from git by
+# theirs -- and all of them land here.
+HORIZON_VENV_SP := $(ROOTDIR)$(NEXT_OPENSTACK_HOME_DIR)/lib/python$(NEXT_PYTHON_VER)/site-packages
 
 CUBE_THEME_SRCS := $(shell find $(CUBE_THEME_SRCDIR) -type f 2>/dev/null)
 
 $(PROJ_HEAVYFS): $(COREDIR)/horizon/local_settings.in $(CUBE_THEME_SRCS)
 
+# install horizon inside the python 3.10 virtual environment
+#
+# The rpms this replaces are openstack-dashboard, openstack-dashboard-theme and
+# python3-django-horizon. Each dashboard plugin is pip installed by the component
+# that owns it; registering the panels and running manage.py is this file's job, and
+# horizon is built last so all of it happens once every plugin is in the venv.
+#
+# Two dependencies that are not horizon requirements and so have to be named:
+#   - mysqlclient replaces the python3-mysqlclient rpm. local_settings.in uses
+#     django.db.backends.mysql for the cached_db session store, and that backend
+#     imports MySQLdb. PyMySQL is already in the venv but django 3.2 rejects it:
+#     it wants MySQLdb >= 1.4.0 and PyMySQL reports 1.0.2.
+#   - pymemcache backs the PyMemcacheCache cache backend. python-memcached is in
+#     the venv already, but MemcachedCache is removed in django 4.1 and there is
+#     no reason to move onto a dead backend now.
 rootfs_install::
-# FIXME:
-#	$(Q)cd $(ROOTDIR) && rpm2cpio $(BLDDIR)/RPMS/openstack-dashboard-$(RHOSP_RELEASE)*.rpm | cpio -idmv "*/locale/*"
-	$(Q)chroot $(ROOTDIR) ln -sf /usr/local/lib/python3.9/site-packages/horizon /usr/share/openstack-dashboard/horizon
-	$(Q)mkdir -p $(ROOTDIR)/$(HORIZON_DIR)/conf
-	$(Q)cp -f $(ROOTDIR)/$(HORIZON_ETCDIR)/local_settings $(ROOTDIR)/$(HORIZON_ETCDIR)/local_settings.orig
-	$(Q)cp -f $(ROOTDIR)/etc/httpd/conf.d/openstack-dashboard.conf $(ROOTDIR)/etc/httpd/conf.d/openstack-dashboard.conf.orig
+	$(Q)# enable dns in the rootfs for downloading packages
+	$(Q)cp -f /etc/resolv.conf $(ROOTDIR)/etc/
+	$(Q)chroot $(ROOTDIR) bash -c "source /opt/openstack-antelope/bin/activate && \
+		pip install -c $(NEXT_OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
+			horizon==$(HORIZON_VER) \
+			mysqlclient==$(MYSQLCLIENT_VER) \
+			pymemcache"
+	$(Q)# clean up dns configurations after downloading packages
+	$(Q)rm -f $(ROOTDIR)/etc/resolv.conf
+
+# lay out /usr/share/openstack-dashboard
+#
+# horizon and openstack_dashboard are symlinks into the venv rather than copies:
+# gunicorn imports them from site-packages either way, so a copy would only go
+# stale. The horizon symlink already worked this way when the target was python
+# 3.9.
+#
+# manage.py is installed from this directory. The openstack-dashboard rpm shipped
+# it; the horizon wheel does not, and there is no console_script equivalent.
+#
+# STATIC_ROOT is pinned to $(HORIZON_APP_DIR)/static in local_settings.in. Horizon
+# would otherwise derive it from the package location and collect 51MB of static
+# assets into the venv, where openstack-dashboard.conf does not alias it.
+rootfs_install::
+	$(Q)chroot $(ROOTDIR) install -d -m 755 $(HORIZON_APP_DIR)
+	$(Q)$(INSTALL_DATA) -f $(ROOTDIR) $(COREDIR)/horizon/manage.py .$(HORIZON_APP_DIR)/manage.py
+	$(Q)chroot $(ROOTDIR) chmod 755 $(HORIZON_APP_DIR)/manage.py
+	$(Q)chroot $(ROOTDIR) ln -sfn $(NEXT_OPENSTACK_HOME_DIR)/lib/python$(NEXT_PYTHON_VER)/site-packages/horizon $(HORIZON_APP_DIR)/horizon
+	$(Q)chroot $(ROOTDIR) ln -sfn $(HORIZON_DIR) $(HORIZON_APP_DIR)/openstack_dashboard
+
+# settings
+#
+# The local_settings.py symlink is what makes settings.py's `from
+# local.local_settings import *` pick up /etc/openstack-dashboard/local_settings.
+# The rpm created the same link; nothing in the wheel does.
+rootfs_install::
+	$(Q)chroot $(ROOTDIR) install -d -m 755 $(HORIZON_ETCDIR)
+	$(Q)chroot $(ROOTDIR) ln -sfn $(HORIZON_ETCDIR)/local_settings $(HORIZON_DIR)/local/local_settings.py
 	$(Q)cp -f $(COREDIR)/horizon/local_settings.in $(ROOTDIR)/$(HORIZON_ETCDIR)/
-	$(Q)cp -f $(COREDIR)/horizon/openstack-dashboard.conf $(ROOTDIR)/etc/httpd/conf.d/
 	$(Q)echo "AVAILABLE_THEMES = [ ('cube', 'Cube Theme', 'themes/cube') ]" >> $(ROOTDIR)/$(HORIZON_ETCDIR)/local_settings.in
 	$(Q)echo "DEFAULT_THEME = 'cube'" >> $(ROOTDIR)/$(HORIZON_ETCDIR)/local_settings.in
+	$(Q)# config_horizon.cpp regenerates local_settings from local_settings.in on
+	$(Q)# every Commit(), so these build-time values never reach a running node.
+	$(Q)# They do have to parse, though: collectstatic and compress below import
+	$(Q)# the settings module, and PyMemcacheCache splits LOCATION on ":" where the
+	$(Q)# MemcachedCache backend it replaces accepted anything.
 	$(Q)sed \
 		-e 's/@CONTROLLER@/localhost/' -e 's/@TIME_ZONE@/America\/New_York/' \
+		-e 's/@SHARED_ID@/localhost/' -e "s/'@CACHE_SERVERS@'/'localhost:11211'/" \
+		-e 's/@DOMAIN@/Default/' -e 's/@HORIZON_DB_PASSWORD@/horizon_dbpass/' \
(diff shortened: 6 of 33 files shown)


Eandalf-Bigstack:
## Condensed summary of this PR description

The description above stays as the full record; this is the short read of the same points, in the
same order, with command transcripts replaced by their results.

---

#### What type of PR is this?

- feature

Horizon leaves the CentOS Stream 9 RDO rpms under system python 3.9 for the python 3.10 venv at
`/opt/openstack-antelope`, like every other service in this wave. Since `mod_wsgi` is built against
python 3.9 and cannot host a 3.10 venv app in-process, the dashboard also leaves httpd: gunicorn on
a unix socket, httpd reverse-proxying `/horizon` onto it — the arrangement keystone, barbican,
placement and monasca-api already use. Horizon is last in the wave by design, which is why this PR
touches ten other components' `.mk` files and closes their "remove once horizon moves into the
venv" notes: `python3-mod_wsgi`, `openstack-manila-ui`, the pip-installed yoga `python-swiftclient`,
and the duplicate python 3.9 installs of designate-, watcher- and masakari-dashboard plus masakari.

| | Yoga | Antelope |
| -- | -- | -- |
| horizon | 22.1.1 (rpm, py3.9) | **23.1.1** (pypi, py3.10 venv) |
| heat-dashboard | *dropped, deferred to #609* | **9.0.0** |
| ironic-ui | *dropped, deferred to #609* | **6.1.0** |
| manila-ui | 7.0.1 (rpm, py3.9) | **9.0.1** |
| designate-dashboard | 2023.1 in py3.9 **and** venv | 2023.1, venv only |
| watcher-dashboard | 2023.1-eol in py3.9 **and** venv | 2023.1-eol, venv only |
| masakari-dashboard | py3.9 + venv, panels from py3.9 | 2023.1, venv only |
| neutron-vpnaas-dashboard | venv, **panel never registered** | 2023.1, panel registered |
| octavia-dashboard | 9.0.1 (rpm, py3.9) | **11.0.1** |

- Close #609

**Scope.** #1202, #1203 and #1204 have merged, so this stands alone against `openstack-develop` at
21 commits: the horizon move, ten dependent plugin cleanups, and the three called out below.

**octavia-dashboard is now included.** It was held back while the octavia *service* was upgraded on
a separate branch, which would have put 2023.1 Load Balancer panels in front of a yoga API. That
branch merged, so octavia-dashboard 11.0.1 joins the octavia 12.0.1 already in the venv — **no
dashboard panel is missing now.**

**Three commits are not horizon work.** They are regressions only the *rebuilt image* could expose;
earlier in-place rounds still had the yoga python 3.9 closure and by construction could not surface
them. Both share one cause — a python 3.9 package other things relied on without ever naming it:
`feat(designate)` restores `openstack dns service list` (its loss made `cluster check` report DNSaaS
NG while all five units were active), and `fix(nova)` restores `/usr/bin/privsep-helper` (its loss
crash-looped `neutron-ovn-metadata-agent` 13,524 times with VM metadata dead — the symlink
`core/nova/nova.mk` has carried as a note since the Nova hop, deferred until every privsep user left
python 3.9, now true). `refactor(privsep)` then converges the `TEMPORARY` venv paths manila,
masakari and cyborg had pinned.

**`manage.py` and five policy files are checked in / generated rather than shipped.** The wheel has
no `manage.py` or console_script equivalent, so it lives in `core/horizon/`; the five
`default_policies/*.yaml` the rpm shipped are now generated by `manage.py dump_default_policies`
from the running services, which beats shipping yoga's copies.

**`local_settings.in` had no drift.** The live file matched the rendered template apart from DB
password, timezone and the two theme lines `horizon.mk` appends. Edits are three django-3.2
deprecations (`ugettext_lazy`, the `RemovedInDjango40Warning` filter, `MemcachedCache`) plus
`STATIC_ROOT` and `COMPRESS_OFFLINE`; Django stays 3.2.16, so the first three are housekeeping.

#### Additional documentation

**Release-notes review (Yoga → Zed → Antelope).** Published notes are thin, so tags `22.1.1` →
`23.0.0` → `23.1.1` were diffed directly: new settings only, nothing removed or renamed in
`settings.rst`, one deprecation (`ANGULAR_FEATURES` Django panels), and none of the new features
enabled. What mattered was dependency-side — django-compressor 2.4.1 → 4.1, python-neutronclient
7.8.0 → 9.0.0, pyScss 1.3.7 → 1.4.0, XStatic-Angular 1.5.8 → 1.8.2.2 — with **zero** template
deletions or renames, so all 13 Cube theme overrides still target live files.

Verified on a full `heavyfs` build deployed to jim-1cc, results only:

- **Build / configs** — the two `/usr/share/openstack-dashboard` paths are symlinks into the venv;
  `openstack-dashboard`, `openstack-octavia-ui` and `python3-mod_wsgi` rpms absent; horizon 23.1.1,
  heat-dashboard 9.0.0, ironic-ui 6.1.0, manila-ui 9.0.1, Django 3.2.16 in the venv; `manage.py
  check` clean. One fix was needed: XStatic sdists built against the venv's pinned setuptools.
- **Health** — `boot_mode status` clean, `cluster check` **all 27 groups ok** (including the two the
  rebuilt image reported NG before the designate and privsep commits), dashboard active under
  gunicorn, `neutron-ovn-metadata-agent` at `NRestarts=0`.
- **Login** — login page 200 in the Cube theme, credential POST 302 to `/horizon/` with `csrftoken`
  and `sessionid` set, and all 10 referenced static assets 200 with correct content types.
- **Volume / network / VM** — driven through the panels' own APIs, so both AngularJS REST and Django
  table actions are exercised. Create returned 201 in each case, the VM launched onto the new
  network reached ACTIVE, Console returned a working VNC URL, the Log tab showed the guest detecting
  the dashboard-attached volume as `[sdb]`, and all three deleted cleanly via *Delete …* actions.
- **Panel sweep** — every registered panel enumerated and fetched: **85 swept, 82 return 200**,
  covering Orchestration, Bare Metal, VPN, Shares, DNS, Instance HA, Optimization and Load Balancer.
  The three 403s are unchanged from Yoga and hidden from the sidebar, gated on neutron extensions
  this deployment does not advertise. Manila's two-protocol restriction survives; logs show zero
  errors, tracebacks or policy warnings.
- **`python3-mod_wsgi` removal proven** — with the module unloaded, httpd restarts clean and all six
  httpd-fronted endpoints still answer as before. That test surfaced the one non-obvious breakage:
  the rpm's httpd drop-in ran `collectstatic`/`compress` under `/usr/bin/python3` as an
  `ExecStartPre` on **httpd**; both moved onto `openstack-dashboard.service`, confirmed over three
  stop/start cycles reaching login 200 within six seconds.
- **Load Balancer panel** — the antelope amphora image is now on the NAS and imported with the
  product's own helper, closing the one gap earlier rounds could not. Panel 200 and linked in the
  sidebar, its Angular assets collected (172 of 176 files) and the CSS bundle rebuilt — worth an
  explicit check, since this panel is the *horizon-plugin-static-not-collected* class. A load
  balancer reached ACTIVE with a STANDALONE amphora, took an HTTP:80 listener, and cascade-deleted
  to zero.
- **Unrelated NAS fix** — the amphora md5 sidecar was `<hash>  <filename>` where
  `core/extpacks/Makefile` diffs a bare hash, so every build re-copied 340 MB; replaced with a bare
  hash and the diff is clean.

traviswu-bigstack (approved):

SekiXu (changes requested):
Reviewed the full diff against the branch (`5289f54`). The write-up, the panel sweep and the httpd-without-`wsgi_module` proof are all well above bar — the arrangement itself (gunicorn on a socket, httpd proxying, plugins pip-installed by their owner and registered by `core/horizon`) is the right shape.

Six things below. The first three I'd like resolved before merge; they share a property — each lives on a path the 1cc verification could not reach.

## Blocking

### 1. WebSSO origin drops to `http://` behind the proxy, and keystone matches it exactly

`local_settings.in:56` sets `SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')`, but `mod_proxy_http` only adds `X-Forwarded-For` / `-Host` / `-Server` — it never sends `X-Forwarded-Proto`, and there is no `RequestHeader` anywhere under `core/`. Under `mod_wsgi` the app ran inside the SSL vhost, so `wsgi.url_scheme` was `https` natively; over the unix socket gunicorn now sees plain http and `request.is_secure()` is `False`.

That matters because WebSSO is on — `local_settings.in:483` `WEBSSO_ENABLED = True`, `:494` `WEBSSO_INITIAL_CHOICE = "cube_idp_mapped"`. `openstack_auth.utils.build_absolute_uri()` builds the `origin` parameter from `request.build_absolute_uri()`, so it becomes `http://<host>/horizon/auth/websso/`, while `config_keystone.cpp:321-322` pins `trusted_dashboard` to `https://<sharedId>/horizon/auth/websso/`. Keystone's federation check is an exact string comparison, so this should fail with "Origin is not a trusted dashboard host".

Second effect: Django 3.2 only runs the strict CSRF Referer check when `is_secure()` is true, so that check is now silently off — a regression relative to mod_wsgi.

Suggested fix in `core/horizon/openstack-dashboard.conf`:

```apache
RequestHeader set X-Forwarded-Proto expr=%{REQUEST_SCHEME}
```

`expr=%{REQUEST_SCHEME}` rather than a literal `"https"`, because this file is in `conf.d/` at server scope and applies to every vhost.

The login evidence in the description goes through Local Credentials, which bypasses this path entirely. Could you re-run it once via the "Cube Account" button?

### 2. `/var/log/horizon/horizon.log` has no rotation

`openstack-dashboard.service:15-16` appends gunicorn's stdout/stderr there, and `LOGGING` in `local_settings.in` has no `filename`, so the console handler lands in the same file. Previously that output went to `/var/log/httpd/horizon_error.log` and was covered by httpd's own logrotate; the `openstack-dashboard` rpm's `/etc/logrotate.d/openstack-dashboard` is also gone with the rpm. With `loglevel = "info"` this grows unbounded on a control node.

`config_manila.cpp:21-24` already handles exactly this case, and says so:

```cpp
// manila no longer arrives as an rpm, so /etc/logrotate.d/openstack-manila -- which
// python3-manila owned -- is gone. Declare the rotation through hex instead...
static LogRotateConf log_conf("manila", "/var/log/manila/*.log", DAILY, 128, 0, true);
```

`config_horizon.cpp` should do the same — `#include <hex/logrotate.h>`, a `LogRotateConf`, and `WriteLogRotateConf(log_conf)` at the end of `Commit()`, the way `config_barbican.cpp` closes out.

### 3. `cube_edge_cfg()` was not updated

`core/main/proj_functions:230-233`:

```bash
remote_run $ctrl python3 /usr/share/openstack-dashboard/manage.py compilemessages 2>&1 > /dev/null
remote_run $ctrl python3 /usr/share/openstack-dashboard/manage.py collectstatic --noinput 2>&1 > /dev/null
remote_run $ctrl python3 /usr/share/openstack-dashboard/manage.py compress --force 2>&1 > /dev/null
remote_run $ctrl systemctl restart httpd
```

Three problems now:

- `python3` is the system 3.9 and can no longer import `openstack_dashboard`. Should be `/opt/openstack-antelope/bin/python`, matching `config_horizon.cpp:88`.
- `systemctl restart httpd` no longer reruns collectstatic/compress — those moved onto `openstack-dashboard.service`. It needs to restart `openstack-dashboard`.
- `HORIZON_APP_DIR` at `proj_functions:13` still points at `/usr/share/openstack-dashboard/openstack_dashboard/local/enabled`, which now resolves through the symlink into the venv's site-packages. Functionally fine, but the blast radius changed and is worth a comment.

Entry point is `config_cube_scan.cpp:517`, which is why a single-node 1cc run does not exercise it.

## Should fix

### 4. The `dump_default_policies` loop swallows failures

`core/horizon/horizon.mk:178-182`. A `for` loop returns the status of its *last* iteration, so a failure in keystone through masakari leaves an empty or missing yaml and the build still passes — surfacing at runtime as `No policy rules for service '<x>'` and a denied panel.

This is weaker than what it replaces. The masakari comment removed in this PR spelled the old guarantee out: "Without this the namespace is not found, that command exits 1 **and the build fails**." Adding `|| exit 1` inside the loop restores it.

### 5. `collectstatic --clear` runs on every start while httpd keeps serving

`openstack-dashboard.service:37-38` runs `collectstatic --noinput --clear` and `compress --force` as `ExecStartPre` on every start. `--clear` empties `STATIC_ROOT` first, and httpd serves `/horizon/static` off that tree independently — `Before=httpd.service` is boot ordering only, so a bare `systemctl restart openstack-dashboard` leaves httpd 404ing every asset for the duration. `Restart=on-failure` with `RestartSec=5s` turns a gunicorn crash-loop into a repeated teardown of the whole 51MB tree.

The three verification cycles in the description stop both units before starting them, which avoids the window. Dropping `--clear` (collectstatic overwrites anyway) would be enough.

### 6. `MariaDB-devel` + `zlib-devel` ship in the product image

`core/mysql/mysql.mk:26` and `:41`. The rationale for choosing the MariaDB.org build over appstream's `mariadb-connector-c-devel` is clear and correct, but both are build-time-only deps for the `mysqlclient` wheel and now stay in the shipped image. Can they be removed in a trailing `rootfs_install::`? If `LOCKED_DNF` makes that impractical, worth stating in the comment that they are deliberately retained.

### 7. The kernel bump is a product change, not a jail-only one

`jail/jail.ubi9.dockerfile:92` moves `KER_VER` from 6.12.93 to 6.12.100. `KER_VER` is defined nowhere else, and the product kernel is read back out of the jail image's `/kernel-info` at `hex/make/kernel_definitions.mk:7` — so this ships a different kernel and rebuilds the nvidia kmod against it. The commit message is a bare `chore(kernel)`. If the driver is that the SIG's rolling `Packages/k/` directory dropped the old rpm, please say so in the commit; otherwise this belongs in its own PR, and the GPU/vGPU work should be told.

## Checked and clear

Listing these so nobody re-derives them:

- `ProxyPass /horizon/static !` at `openstack-dashboard.conf:15` does precede `ProxyPass /horizon` at `:25`, and mod_proxy does take precedence over mod_alias — the exclusion is both correct and necessary.
- gunicorn 20.1.0's `http/wsgi.py` really is `path_info.split(script_name, 1)[1]`, so keeping the `/horizon` prefix on the proxy target is right.
- Plugins installing before horizon cannot drag in a wrong horizon: `core/heavyfs/os-antelope-pip-upper-constraints.txt:120` carries `horizon===23.1.1`, and every plugin install passes `-c`.
- `WriteDjangoConf()` does not lose the build-time `640 root:apache` on `local_settings` — shell `>` truncates the existing inode.
- Nothing in the tree references `local_settings.orig` or `openstack-dashboard.conf.orig`, so dropping them is safe.
- `HORIZON_DIR` has no consumer outside `core/heavyfs/Makefile` and `core/horizon/horizon.mk`.

## Minor

- `config_horizon.cpp:200`: `bool enabled = IsControl(s_eCubeRole)` is dead — `:170` already returns early when that is false, so `enabled` is always true here. `config_barbican.cpp` reaches `SystemdCommitService(false, …)` on non-control nodes and actively disables the unit; horizon never will, so a demoted control node keeps the dashboard enabled.
- `core/horizon/manage.py` gets `chmod 755` but has no shebang. Harmless since every caller names the interpreter, but the mode implies otherwise.
- `HORIZON_VENV_SP` (horizon.mk:20) and `HORIZON_DIR` (heavyfs/Makefile:23) spell the same path twice; `STATIC_ROOT` is hardcoded in three places.
- `<Directory …/static>` keeps `Options All` / `AllowOverride All`. Pre-existing, but the file was rewritten anyway — `Options -Indexes` / `AllowOverride None` is enough for a static tree.

Eandalf-Bigstack:
## Review responses

Thanks for the review. All seven items are addressed, plus the minor list. Nine commits pushed, `5289f54..c1b4bc0`, re-verified on e2e node jim-1cc.

One item I did not apply, and one thing the review led me to that turned out to be a real bug elsewhere — both below.

| # | Item | Outcome | Commit |
|---|---|---|---|
| 1 | `X-Forwarded-Proto` on the dashboard vhost | **not applied** — it breaks WebSSO; documented instead | `dbcb023` |
| 2 | Nothing rotates `/var/log/horizon` | fixed | `c6db08b` |
| 3 | `cube_edge_cfg()` uses the wrong interpreter | fixed | `9c3a717` |
| 4 | Policy dump loop swallows failures | fixed | `2a17dd1` |
| 5 | `collectstatic --clear` | fixed | `6bbf029` |
| 6 | `MariaDB-devel` / `zlib-devel` reaching the image | no change needed — mechanism recorded | `c8e232c` |
| 7 | Kernel bump provenance | routine, explained below — no commit | — |
| minor | dead `enabled`, `manage.py` shebang, path duplication, `<Directory>` options | fixed | `c6db08b`, `0d2dec4`, `c1b4bc0`, `dbcb023` |
| — | **the actual cause of the broken federated login** | fixed | `1fec057` |

---

### 1. `RequestHeader set X-Forwarded-Proto expr=%{REQUEST_SCHEME}` — not applied

The premise is that nothing supplies the header `SECURE_PROXY_SSL_HEADER` reads. Something does: `config_haproxy.cpp` adds it at the edge, and `mod_proxy` forwards it to gunicorn untouched.

Applying the suggestion actively breaks federated login, which I confirmed on jim-1cc rather than argued. This file is at server scope in `conf.d` and lands in a plain `:8080` vhost, so `REQUEST_SCHEME` here is `http` — the directive **overwrites** haproxy's `https`:

- `request.is_secure()` becomes `False`
- the WebSSO origin flips to `http://<host>/horizon/auth/websso/`, which no longer matches the `trusted_dashboard` value `config_keystone.cpp` pins, so login fails with *"Origin is not a trusted dashboard host"*
- Django 3.2's strict CSRF `Referer` check is silently disabled, since it only runs when `is_secure()` is true

I reverted it. With the tree as it stands, the origin is correct in a now-passing SSO flow:

```
2. Cube Account -> keystone websso   http=302
   origin=https://10.1.0.1/horizon/auth/websso/
```

`dbcb023` leaves a comment in the file recording all of the above, so the directive does not get re-added by the next reader who wonders where the header comes from.

That said — the review instinct was pointing at something real. The login *was* broken; the header just was not what broke it. See the last section.

### 2. `/var/log/horizon` rotation — `c6db08b`

Correct, nothing rotated it. gunicorn's stdout/stderr and horizon's own `LOGGING` console handler both land in `horizon.log`, which the unit appends to; that output used to go to `/var/log/httpd/horizon_error.log` under httpd's logrotate, and the rpm's `/etc/logrotate.d/openstack-dashboard` left with the rpm.

Declared through hex the way manila/cinder/heat/ironic do — `LogRotateConf` + `WriteLogRotateConf()`. `copytruncate` matters here specifically: gunicorn holds the descriptor open through systemd's `append:` redirection and does not reopen on SIGHUP. The stanza this produces, from the existing hex-generated `/etc/logrotate.d/manila`:

```
daily / missingok / compress / delaycompress / notifempty / copytruncate / maxsize 128M
```

### 3. `cube_edge_cfg()` — `9c3a717`

Both halves were wrong: `/usr/bin/python3` (3.9) cannot import `openstack_dashboard` or its plugins, and httpd is only the reverse proxy now, so restarting it no longer reruns collectstatic/compress the way the rpm's httpd drop-in did. An edge configuration would have left stale assets with no obvious error. Now uses the venv interpreter and restarts `openstack-dashboard`.

```
compilemessages exit=0
systemctl restart openstack-dashboard -> active
login page http=200
```

### 4. Policy dump loop — `2a17dd1`

Right, and worth stating why it matters: a shell `for` loop reports only its last iteration's status, so a failure in any earlier namespace left an empty or missing yaml and the build still passed — surfacing much later as `No policy rules for service '<x>'` and a denied panel. `masakari.mk` used to state the guarantee explicitly and folding the dumps into a loop dropped it. Added `|| exit 1` per iteration:

```
old loop shape, deliberately bad namespace -> exit=0
new loop shape, deliberately bad namespace -> exit=1
new loop shape, good run                   -> exit=0
```

### 5. `collectstatic --clear` — `6bbf029`

Dropped. httpd serves `/horizon/static` straight off `STATIC_ROOT` and is a separate unit, so `--clear` opened a window where httpd 404s all 51 MB of assets; `Restart=on-failure` turns a gunicorn crash-loop into a repeated teardown of that tree, and `Before=httpd.service` only orders boot, not a bare `systemctl restart openstack-dashboard`. `collectstatic` overwrites what it collects, so `--clear` only bought removal of files a plugin stopped shipping.

Verified: an asset stayed `200` across eight samples three seconds apart, spanning a restart.

### 6. `MariaDB-devel` / `zlib-devel` in the image — `c8e232c`

They do not reach the product image, so there is nothing to strip here — but the mechanism is not obvious, so I traced it in the build tree rather than inferring from a node, and recorded it in the comment. `core/main/cube-post.mk` autoremoves every installed package matching `devel|headers` (bar `python3-devel`) as its final cleanup:

| | `core/heavyfs/heavy_rootfs` | `core/main/rootfs` (ships) |
|---|---|---|
| `MariaDB-devel` / `zlib-devel` / `gcc` | installed | not installed |
| `mysql.h`, `zlib.h`, `/usr/bin/gcc`, `libmariadb.so` | present | absent |
| rpm count | 1311 | 1214 |

`gcc` is not matched by that egrep and goes as a dependent — from the shipped rootfs's own dnf history:

```
Command Line   : autoremove -y glibc-devel.x86_64
    Removed gcc-11.5.0-15.el9.x86_64
    Removed gcc-c++-11.5.0-15.el9.x86_64
    Removed glibc-devel-2.34-275.el9.x86_64
    Removed libstdc++-devel-11.5.0-15.el9.x86_64
    Removed annobin / gcc-plugin-annobin / libxcrypt-devel
```

What survives is exactly what `mysqlclient` links at import time: `MariaDB-shared` owning `libmariadb.so.3`, `zlib` owning `libz.so.1`, both confirmed present in the shipped rootfs.

### 7. Kernel bump

Routine, and forced rather than chosen: `kernel-6.12.93*` is no longer in the repos (`No matching Packages to list`), so the jail moved to `6.12.100-1.el9`. The mechanism you described is right — jail `ARG KER_VER` → `/kernel-info` → `hex/make/kernel_definitions.mk` `KERNEL_VERS` → product kernel and the nvidia kmod. Since the minor version is unchanged, the nvidia driver build is unaffected, so there is nothing GPU/vGPU-specific to call out. Left as-is rather than rewording the commit, which would need a force-push.

### Minor items

- **dead `enabled` variable** (`c6db08b`) — removed; `SystemdCommitService()` takes an unconditional `true`, which is what the value always was, since `Commit()` returns at the top for every non-control role. Left a comment on the one case that shape gets wrong (a *demoted* control node keeps the dashboard enabled) and noted `config_apache2.cpp` has the identical shape for httpd — fixing one alone would leave the proxy enabled in front of a disabled backend.
- **`manage.py` shebang** (`0d2dec4`) — `chmod 755` promised something the file could not do; now names the venv interpreter. `manage.py check` runs directly: *System check identified no issues*.
- **path duplication** (`c1b4bc0`) — `HORIZON_VENV_SITE_PACKAGES` named once, `HORIZON_DIR`/`HORIZON_VENV_SP` derived. Same commit moves the whole `HORIZON_*` block out of `core/heavyfs/Makefile` into `core/horizon/horizon.mk`; heavyfs is left with zero `HORIZON_` references and no other makefile reads one.
- **static `<Directory>`** (`dbcb023`) — `Options -Indexes` / `AllowOverride None`; the rpm-era block carried `Options All` / `AllowOverride All`, which also cost a `.htaccess` stat per request.

---

### The actual cause of the broken federated login — `1fec057`

Chasing item 1 surfaced why WebSSO was failing, and it is not horizon's: **federated login has been broken cluster-wide since `d1b0699`** ("change request flow from httpd wsgi -> keystone to httpd proxy -> gunicorn -> keystone").

`mod_auth_mellon` publishes the SAML assertion as httpd *environment variables*, which only reach an application httpd hosts in-process. `mod_proxy` forwards headers, not the request environment, so every WebSSO attempt arrived at keystone carrying no assertion and got a bare 401, with keystone logging only:

```
WARNING keystone.server.flask.application [req-ec08c5d9...] Authorization failed.
The request you have made requires authentication. from 10.1.0.1:
keystone.exception.Unauthorized
```

Skyline's "Cube Account" button fails identically, which is what rules out a dashboard-side cause:

| entry point | before | after |
|---|---|---|
| horizon via `:5000` (what `local_settings` uses) | 401 | token issued |
| horizon via `:5443` (the mellon vhost) | 401 | token issued |
| **skyline via `:5000`** (control) | 401 | token issued |

httpd had everything needed all along — a temporary `CustomLog` probe on the websso request:

```
MELLON_IDP=https://10.1.0.1:10443/auth/realms/master | user=admin | groups=cube-admins
  | nameid=G-840e1358-... | /v3/auth/OS-FEDERATION/.../mapped/websso
```

`MELLON_IDP` matches the IdP's `remote_ids`; `MELLON_groups` matches the mapping's `any_one_of: [cube-admins]`. So the fix carries those three across the socket as `X-Mellon-*` headers and restores them on the far side in a small WSGI wrapper (`cube_mellon_wsgi.py`) around `initialize_public_application()` — `federation.utils.get_assertion_params_from_env()` harvests them from `flask.request.environ`, and `remote_id_attribute` names `MELLON_IDP`.

**The security half deserves review attention.** A header channel into an authentication decision is forgeable, so the config `unset`s all three at **server scope** — deliberately not inside `<Location /v3>`, because `keystone-wsgi.conf` also proxies `/identity` to the same socket and mellon does not guard that prefix. That placement is load-bearing; with the unsets commented out, a plain unauthenticated `curl` carrying three self-supplied headers is issued a token for any user the mapping can name:

```
# without the unsets
POST /identity/v3/OS-FEDERATION/identity_providers/cube_idp/protocols/mapped/auth
  -> http=201   user=admin (IAM)

# with them
  -> http=401
```

Both the exploit and the guard are recorded in the config comment so the unsets are not tidied away later.

**Result** — Cube Account login as `admin`/`admin` on jim-1cc:

```
5. relay SAMLResponse -> mellon ACS  https://10.1.0.1:5443/v3/mellon/postResponse  http=200
6. relay keystone token -> horizon   https://10.1.0.1/horizon/auth/websso/          http=200
   GET /horizon/project/ http=200 title='Instance Overview - Cube Dashboard'
   ==> KEYCLOAK SSO LOGIN OK
```

**Regression checks**, since all keystone traffic now passes through the wrapper: `token issue` succeeds, 21 users / 22 services / 66 endpoints list, and local password login as `admin_cli` reaches the dashboard.

This is a keystone auth-path change riding in a horizon PR because it is what makes horizon login work; it is self-contained in `core/keystone/` and reviewable on its own. Happy to split it into its own PR if you would rather review it separately.

SekiXu (approved):
LGTM. Re-checked all nine commits against the tree (`5289f54..c1b4bc0`).

Two of my items were wrong and your evidence settles both:

- **#1** — `config_haproxy.cpp:205` supplies the header at the 443 frontend and `config_apache2.cpp:17` puts this file in a plain `:8080` vhost, so the directive I asked for would have overwritten `https` with `http` and caused exactly the failure I was predicting. Good call not applying it, and better call verifying it on a node rather than arguing it.
- **#6** — `core/main/cube-post.mk:22` already autoremoves `devel|headers`. I should have traced the build tree before raising it.

`1fec057d` is the find of this round. Federated login being broken since `d1b0699` — for horizon and skyline alike — was never going to surface from a dashboard-side reading, and the forgery check that put the `RequestHeader unset` at server scope rather than inside `<Location /v3>` is the part that makes the header channel safe. Keeping it in this PR is fine by me; it is self-contained in `core/keystone/` and reads on its own.

Also confirmed the variable move in `c1b4bc0` is clean: no `.mk` outside `core/horizon/horizon.mk` references `HORIZON_*` any more (the `HORIZON_APP_DIR` in `proj_functions:16` is an unrelated shell variable of the same name).

Thanks for the thorough turnaround.

Issue #579, opened 2026-01-16
[EPIC] Refactor Network Analysis


jdjgya:
# High-Performance eBPF Network Telemetry Engine

## Background

Modern cloud infrastructure requires precise, real-time network visibility for capacity planning, security, and downstream billing systems. However, **traditional network monitoring tools (like OVS flow tracking, NetFlow, or iptables) lead to high CPU overhead, packet drops at 10G+ speeds, and a lack of granular tenant-awareness**, especially in multi-tenant OpenStack environments.

In the current CubeCOS network stack, accurately classifying traffic types is a major challenge. This introduces several critical limitations:

- **Lack of Granular Classification:** The platform struggles to clearly differentiate between external internet traffic (North-South) and internal workload traffic (East-West) at the network border.
- **Tenant Context Loss:** Traffic passing through infrastructure proxies (like Load Balancers or NAT routers) loses its original tenant context, often attributing customer bandwidth to the admin/service project.
- **Performance Bottlenecks:** Parsing packets in user-space at line-rate causes severe CPU starvation for actual compute workloads.

To address these gaps, CubeCOS needs to productize a **high-performance, eBPF-based network telemetry engine**, enabling line-rate traffic classification directly inside the Linux kernel. This component will serve as the single source of truth for network observability, categorizing traffic into distinct profiles (Ingress, Egress, Intra-tenant, Inter-tenant) without manual intervention or heavy compute overhead.

## Goals

- Enable **multi-tenant, byte-level network traffic classification** for OpenStack workloads.
- Accurately categorize all traffic into distinct network profiles: **Ingress, Egress, Intra-Tenant, and Inter-Tenant**.
- Utilize **eBPF (XDP/TC)** for high-speed packet interception and in-kernel aggregation to ensure near-zero CPU overhead at 10Gbps+ speeds.
- Synchronize OpenStack metadata in real-time (via MySQL cold-starts and Kafka event streams) to map network flows to precise Tenant IDs.
- Expose highly compressed, cardinality-safe metrics via **Prometheus** for consumption by downstream billing and monitoring systems.
- Ensure **crash resiliency** (zero lost packets if the user-space agent restarts).

### 1. Traffic Classification (Core)

- [ ] Traffic is accurately categorized into four primary profiles:
  - **Egress (North-South Out):** Tenant traffic leaving the cloud to the external internet.
  - **Ingress (North-South In):** External internet traffic entering a tenant's environment.
[...]

PR #1257, opened 2026-08-13
feat(telegraf): upgrade to 1.24.4 and give sflow a read-only openstack db account

### LINKED ISSUE #557: [Bug]: Telegraf enters NG state after enabling sflow
## Describe the bug

Telegraf enters NG state after enabling sflow

## Version

Version 3.0.0 RC4 - 3cc, 1cc
Version 2.5.0 GA - 1cc
Version 2.4.3 GA - 1ec

## To Reproduce

1. CLI > iaas > flowdata > set_flowdata
2. Set status "on"
3. Login Web UI > Home/Chart page >footer Extra Monitor "Network" and "Device"
4. No data display

## Expected behavior

Extra Monitor "Network" and "Device" page will displayed data, and won't make telegraf NG

## Additional context

```
Jul 07 15:15:14 dell200 telegraf[943501]: 2025-07-07T07:15:14Z I! Starting Telegraf
Jul 07 15:15:14 dell200 telegraf[943501]: 2025-07-07T07:15:14Z I! Loaded inputs: ceph cpu disk diskio exec (11x) ipmi_sensor kernel mem net processes sflow swap system
Jul 07 15:15:14 dell200 telegraf[943501]: 2025-07-07T07:15:14Z I! Loaded aggregators:
Jul 07 15:15:14 dell200 telegraf[943501]: 2025-07-07T07:15:14Z I! Loaded processors:
Jul 07 15:15:14 dell200 telegraf[943501]: 2025-07-07T07:15:14Z I! Loaded outputs: kafka (3x)
Jul 07 15:15:14 dell200 telegraf[943501]: 2025-07-07T07:15:14Z I! Tags enabled: host=dell200
Jul 07 15:15:14 dell200 telegraf[943501]: 2025-07-07T07:15:14Z I! [agent] Config: Interval:1m0s, Quiet:false, Hostname:"dell200", Flush Interval:10s
Jul 07 15:15:14 dell200 telegraf[943501]: 2025-07-07T07:15:14Z I! [inputs.sflow] Listening on udp://[::]:6343
Jul 07 15:15:14 dell200 telegraf[943501]: 2025-07-07T07:15:14Z E! [inputs.sflow] Failed to read openstack port info: Error 1698: Access denied for user 'root'@'localhost'
Jul 07 15:15:14 dell200 telegraf[943501]: panic: runtime error: invalid memory address or nil pointer dereference
Jul 07 15:15:14 dell200 telegraf[943501]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x61 pc=0x15a8d2e]
Jul 07 15:15:14 dell200 telegraf[943501]: goroutine 112 [running]:
Jul 07 15:15:14 dell200 telegraf[943501]: database/sql.(*Rows).closemuRUnlockIfHeldByScan(0x1?)
Jul 07 15:15:14 dell200 telegraf[943501]:         /usr/local/go/src/database/sql/sql.go:3412 +0xe
Jul 07 15:15:14 dell200 telegraf[943501]: database/sql.(*Rows).Next(0x0)
Jul 07 15:15:14 dell200 telegraf[943501]:         /usr/local/go/src/database/sql/sql.go:3033 +0x1c
Jul 07 15:15:14 dell200 telegraf[943501]: github.com/influxdata/telegraf/plugins/inputs/sflow.(*SFlow).cacheOpenstackPortInfo(0xc00097d110)
Jul 07 15:15:14 dell200 telegraf[943501]:         /home/jenkins/workspace/centos9_cubecos_prod/build/core/telegraf/telegraf/plugins/inputs/sflow/sflow.go:161 +0x255
Jul 07 15:15:14 dell200 telegraf[943501]: github.com/influxdata/telegraf/plugins/inputs/sflow.(*SFlow).read(0xc00097d110, {0x2d684c0, 0xc00090dce0}, {0x2d65410, 0xc0000d4db8})
Jul 07 15:15:14 dell200 telegraf[943501]:         /home/jenkins/workspace/centos9_cubecos_prod/build/core/telegraf/telegraf/plugins/inputs/sflow/sflow.go:132 +0xa5
Jul 07 15:15:14 dell200 telegraf[943501]: github.com/influxdata/telegraf/plugins/inputs/sflow.(*SFlow).Start.func2()
Jul 07 15:15:14 dell200 telegraf[943501]:         /home/jenkins/workspace/centos9_cubecos_prod/build/core/telegraf/telegraf/plugins/inputs/sflow/sflow.go:100 +0x5c
Jul 07 15:15:14 dell200 telegraf[943501]: created by github.com/influxdata/telegraf/plugins/inputs/sflow.(*SFlow).Start in goroutine 1
Jul 07 15:15:14 dell200 telegraf[943501]:         /home/jenkins/workspace/centos9_cubecos_prod/build/core/telegraf/telegraf/plugins/inputs/sflow/sflow.go:98 +0x2e5
Jul 07 15:15:14 dell200 systemd[1]: telegraf.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Jul 07 15:15:14 dell200 systemd[1]: telegraf.service: Failed with result 'exit-code'.
Jul 07 15:15:15 dell200 systemd[1]: telegraf.service: Scheduled restart job, restart counter is at 5.
Jul 07 15:15:15 dell200 systemd[1]: Stopped The plugin-driven server agent for reporting metrics into InfluxDB.
Jul 07 15:15:15 dell200 systemd[1]: telegraf.service: Start request repeated too quickly.
Jul 07 15:15:15 dell200 systemd[1]: telegraf.service: Failed with result 'exit-code'.
Jul 07 15:15:15 dell200 systemd[1]: Failed to start The plugin-driven server agent for reporting metrics into InfluxDB.
```

- [x] **Handbook knowledge update** — landed in bigstack-handbook via [bigstack-handbook#262](https://github.com/bigstack-oss/bigstack-handbook/pull/262) and corrected in [bigstack-handbook#270](https://github.com/bigstack-oss/bigstack-handbook/pull/270): [`kb/cubecos/known-issues/telegraf-sflow-panics-on-openstack-db-failure.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/telegraf-sflow-panics-on-openstack-db-failure.md) ("telegraf sflow panics on a nil *sql.Rows when the openstack port lookup fails"), plus [`kb/cubecos/runbooks/bump-the-telegraf-fork.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/runbooks/bump-the-telegraf-fork.md) (how the fork is maintained now that it carries only the crypto pin), cross-linked to [PR bigstack-oss/cubecos#1257](https://github.com/bigstack-oss/cubecos/pull/1257).

> **How this was resolved.** Not by the fix originally proposed. Review on cubecos#1257 reverted the whole sflow enrichment patch, so the plugin has no database code left and the panic is structurally impossible rather than guarded against. The Network and Device pages therefore have no data source until Lachesis ships in 3.2.0 — tracked under [cubecos#579](https://github.com/bigstack-oss/cubecos/issues/579). Investigation and root-cause analysis remain @arasHi87's.

### LINKED ISSUE #622: [Story] [Optional] Upgrade Telegraf from 1.17.2 to 1.24.3
## Description

As a CubeCOS operator, I want to resolve potential security concerns with Telegraf so that we could pass audits.

## Acceptance Criteria

- [x] Upgrade Telegraf from 1.17.2 to 1.24.3
- [x] Ensure the metrics collection mechanism from Telegraf is still working as expected
- [x] Ensure Telegraf could send metrics to our monitoring pipeline, especially Kafka

## Story Points: 3

> | -- | -- | -- |

- [x] **Handbook knowledge update** — landed in bigstack-handbook via [bigstack-handbook#262](https://github.com/bigstack-oss/bigstack-handbook/pull/262) and corrected in [bigstack-handbook#270](https://github.com/bigstack-oss/bigstack-handbook/pull/270): [`kb/cubecos/runbooks/bump-the-telegraf-fork.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/runbooks/bump-the-telegraf-fork.md) ("Bump the CubeCOS telegraf fork to a new upstream version"), plus [`kb/cubecos/known-issues/telegraf-sflow-panics-on-openstack-db-failure.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/telegraf-sflow-panics-on-openstack-db-failure.md) (sflow nil-rows panic on the openstack port lookup), cross-linked to [PR #1257](https://github.com/bigstack-oss/cubecos/pull/1257).

> **Scope note.** The sflow enrichment patch was reverted during review of #1257 — Lachesis replaces it in 3.2.0, so hardening it was work with a known expiry date. The fork now carries only the `x/crypto` security pin, and full sflow removal is owned by #579. Both handbook entries were rewritten in #270 to describe removal rather than the fix originally documented.

Eandalf-Bigstack:
#### What type of PR is this?

/kind feature

Upgrades Telegraf from 1.17.2 to **1.24.4**, and fixes the sflow crash that has been
taking the agent down whenever flowdata is enabled.

**Why 1.24.4 and not 1.24.3.** The story asks for 1.24.3. 1.24.4 is the newest tag on
the 1.24 line and is a maintenance release that does not touch the sflow plugin at all,
so taking it costs nothing. Flagging it explicitly so the acceptance criteria do not
read as unmet.

**Patches now live in one place.** `core/telegraf/Makefile` used to mutate the freshly
cloned tree on every build (`go get golang.org/x/crypto@v0.31.0`, a `go mod edit
-replace` for `nats-io/jwt`, then `go mod tidy`). That meant the fork's own `go.mod`
never described what actually shipped, and two repos had to agree about it. Those steps
are deleted here; the crypto bump is pinned in `go.mod` on the fork's `1.24.4-cube`
branch instead. The `nats-io/jwt` replace is dropped rather than moved — 1.24.4 uses the
`jwt/v2` module path at v2.3.0, and the v1 path appears in `go.sum` only as `/go.mod`
hash lines with no `h1:` content hash, so it is not linked into the binary.

**sflow no longer needs root.** The plugin reads neutron/nova/keystone directly to tag
flows with their owning tenant and instance, using a DSN hardcoded to
`root@unix(/var/lib/mysql/mysql.sock)/`. That only authenticates when the *OS* user is
root, and telegraf runs as the unprivileged `telegraf` user — so every lookup was
rejected with `Error 1698: Access denied`, and then a nil `*sql.Rows` was dereferenced
and killed the process.

The reported workaround was a systemd drop-in setting `User=root`. Since #622 is an
audit story, running a UDP-facing daemon as root seemed like the wrong direction, so
this provisions a dedicated `telegraf` MySQL account with `SELECT` only and passes its
DSN to the plugin through a new `openstack_dsn` option. The agent stays unprivileged.

Companion fork branch: `bigstack-oss/telegraf@1.24.4-cube` (4 commits, already pushed;
the fork's default branch now points at it).

Merging this PR should close **both** issues:

Fixes #622
Fixes bigstack-oss/cube-cos-ui#557

- bigstack-oss/cubecos#622 — Telegraf is upgraded off 1.17.2, metrics collection is
  intact, and Kafka delivery is confirmed on jim-1cc.
- bigstack-oss/cube-cos-ui#557 — the sflow panic is gone, and the port cache now
  populates so the Network and Device monitor pages have data to render.

> Note

**Credit.** The root-cause analysis for the crash — both the root-vs-telegraf user
mismatch and the missing `return` on the nil `*sql.Rows` — is @arasHi87's, from
[this comment](https://github.com/bigstack-oss/cube-cos-ui/issues/557#issuecomment-4379531009).
The code fix on the fork carries them as co-author; this PR is the least-privilege
version of their verified workaround.

**Grants are database level, not table level.** MariaDB rejects a table-level grant on a
table that does not exist yet, and this module can commit before those schemas are
populated. `keystone.*` therefore also covers the credential tables. Narrowing to the
four tables actually read (`neutron.ports`, `neutron.floatingips`, `nova.instances`,
`keystone.project`) is worth revisiting once commit ordering is guaranteed. Called out
so it is a conscious trade rather than an oversight.

**Upstream had already fixed part of our patch.** Porting `576ff076` produced two
conflicts in `types.go`, both because upstream independently added the same nil-header
guards Travis had hand-rolled in 2021. Upstream's versions won; only the
`header_protocol` tag suppression was kept.

**Pre-existing test debt, deliberately not touched.** The sflow package has 7 failing
tests, because our patch intentionally drops tags (`agent_address`, `header_protocol`,
`input_ifindex`, `output_ifindex`, `sample_direction`, `source_id_*`) and adds `proto`,
while the upstream tests still assert the upstream tag set. This is not new — the
`1.17.2-cube` branch already carried 85 + 10 stale expectations against the same
commented-out assignments, and baseline v1.24.4 passes clean. Fixing it means updating
~134 expectation sites across three files, which belongs in its own change.

**e2e node state.** jim-1cc (cc1) is already running this build, applied in place.
Rollback: `/usr/bin/telegraf.bak-1.17.2-20260813-151802`,
`/etc/telegraf/telegraf.conf.bak-20260813-151802`, and
`DROP USER 'telegraf'@'localhost'`. Its `openstack_dsn` currently holds the raw default
password rather than the salted one; the next `hex_config` commit rewrites it correctly.

#### Additional documentation

```docs
Verified on jim-1cc (cc1), built in the centos9_cubecos_jim_200.13-1cc jail.

Build
  go build ./plugins/inputs/sflow/   clean
  go vet   ./plugins/inputs/sflow/   clean
  make                                Telegraf 1.24.4
  go mod verify                       all modules verified
  go directive                        stays at 1.19 after the crypto bump
  make -C core/modules all            CXX config_telegraf.o clean under -Wall -Werror
  make -C core/main hex_config V=2    links clean

Runtime
  Loaded inputs: ceph cpu disk diskio exec (12x) kernel mem net processes sflow swap system
  Loaded outputs: kafka (3x)
  I! [inputs.sflow] Listening on udp://[::]:6343
  I! [inputs.sflow] Updated Openstack Port Cache
  running as uid=995(telegraf), not root
  panics since upgrade: 0

Regression test for cube-cos-ui#557 (deliberately wrong DSN password)
  E! [inputs.sflow] Failed to read openstack port info:
     Error 1045: Access denied for user 'telegraf'@'localhost' (using password: YES)
  systemctl is-active  -> active
  NRestarts             -> 0
  previous build on this path: SIGSEGV, then "Start request repeated too quickly"

Kafka delivery
  consumed live from telegraf-metrics:
    swap,host=cc1,role=cube total=0i,used=0i,free=0i,used_percent=0 1786605660000000000
    kernel,host=cc1,role=cube entropy_avail=256i,... 1786605660000000000

Not covered
  - telegraf-hc-metrics produced no message inside a 30s window; it is fed by exec
    plugins on a slower interval, so telegraf-metrics is the delivery evidence.
  - No live sFlow source exists on cc1, so the read loop was triggered with a synthetic
    UDP packet. That fully exercises cacheOpenstackPortInfo(), but real flow decoding
    and the resulting tenant/instance tags were not observed end to end.
```

### DIFF
--- a/core/modules/config_telegraf.cpp
+++ b/core/modules/config_telegraf.cpp
@@ -20,6 +20,7 @@
 #include <cluster.hpp>
 
 #include "include/role_cubesys.h"
+#include "mysql_util.h"
 
 #define DELL_POWEREDGE_XR_MARKER "/etc/appliance/state/dell_poweredge_xr_detected"
 
@@ -36,6 +37,10 @@
 
 const static char NAME[] = "telegraf";
 
+// read-only account used by the sflow input to look up openstack port info
+static const char DBUSER[] = "telegraf";
+static const char DBPASS[] = "cIkpxxCLUFBbw4iO";
+
 static bool s_bCubeModified = false;
 static bool s_bKapacitorModified = false;
 
@@ -48,14 +53,22 @@ static Configs cfg;
 CONFIG_GLOBAL_STR_REF(CTRL_IP);
 CONFIG_GLOBAL_STR_REF(SHARED_ID);
 
+// private tunings
+CONFIG_TUNING_STR(TELEGRAF_DBPASS, "telegraf.db.password", TUNING_UNPUB, "Set telegraf database password.", DBPASS, ValidateRegex, DFT_REGEX_STR);
+
 // using external tunings
 CONFIG_TUNING_SPEC_STR(CUBESYS_ROLE);
+CONFIG_TUNING_SPEC_STR(CUBESYS_SEED);
+CONFIG_TUNING_SPEC_BOOL(CUBESYS_SALTKEY);
 CONFIG_TUNING_SPEC_STR(KAPACITOR_ALERT_FLOW_BASE);
 CONFIG_TUNING_SPEC_STR(KAPACITOR_ALERT_FLOW_UNIT);
 CONFIG_TUNING_SPEC_INT(KAPACITOR_ALERT_FLOW_THRESHOLD);
 
 // parse tunings
+PARSE_TUNING_STR(s_dbPass, TELEGRAF_DBPASS);
 PARSE_TUNING_X_STR(s_cubeRole, CUBESYS_ROLE, 1);
+PARSE_TUNING_X_STR(s_seed, CUBESYS_SEED, 1);
+PARSE_TUNING_X_BOOL(s_saltkey, CUBESYS_SALTKEY, 1);
 PARSE_TUNING_X_STR(s_alertFlowBase, KAPACITOR_ALERT_FLOW_BASE, 2);
 PARSE_TUNING_X_STR(s_alertFlowUnit, KAPACITOR_ALERT_FLOW_UNIT, 2);
 PARSE_TUNING_X_INT(s_alertFlowThreshold, KAPACITOR_ALERT_FLOW_THRESHOLD, 2);
@@ -247,6 +260,26 @@ CommitCheck(bool modified, int dryLevel)
     return s_bConfigChanged;
 }
 
+// The sflow input reads neutron/nova/keystone directly to tag flows with the
+// owning tenant and instance. Give it a dedicated read-only account so the agent
+// keeps running as the unprivileged telegraf user instead of root.
+static bool
+SetupCheck(const std::string &dbPass)
+{
+    static const char *dbs[] = { "neutron", "nova", "keystone" };
+
+    for (size_t i = 0; i < sizeof(dbs) / sizeof(dbs[0]); i++) {
+        std::string sql = "GRANT SELECT ON " + std::string(dbs[i]) + ".* TO '" + DBUSER +
+                          "'@'localhost' IDENTIFIED BY '" + dbPass + "'";
+        if (!MysqlUtilRunSQL(sql.c_str())) {
+            HexLogError("failed to grant %s read access to %s", DBUSER, dbs[i]);
+            return false;
+        }
+    }
+
+    return true;
+}
+
 static bool
 Init()
 {
@@ -294,14 +327,25 @@ Commit(bool modified, int dryLevel)
         WriteDeviceConfig();
 
         if (IsControl(s_eCubeRole)) {
-            if (HexSystemF(0, "sed -e 's/@BASE@/%s/' -e 's/@UNIT@/%s/' -e 's/@THRESHOLD@/%d/' %s > %s",
+            std::string dbPass = GetSaltKey(s_saltkey, s_dbPass.newValue(), s_seed.newValue());
+
+            if (!SetupCheck(dbPass))
+                return false;
+
+            if (HexSystemF(0, "sed -e 's/@BASE@/%s/' -e 's/@UNIT@/%s/' -e 's/@THRESHOLD@/%d/' -e 's|@DBPASS@|%s|' %s > %s",
                            s_alertFlowBase.c_str(), s_alertFlowUnit.c_str(), s_alertFlowThreshold.newValue(),
+                           dbPass.c_str(),
                            CTRL_CONF IN_EXT, CTRL_CONF) != 0) {
                 HexLogError("failed to update %s", CTRL_CONF);
                 return false;
             }
             HexSystemF(0, "cat %s >> %s", CTRL_CONF, CONF);
         }
+
+        // both files now carry the sflow database password
+        HexSetFileMode(CONF, "root", "telegraf", 0640);
+        if (IsControl(s_eCubeRole))
+            HexSetFileMode(CTRL_CONF, "root", "telegraf", 0640);
     }
 
     SystemdCommitService(true, NAME);
@@ -368,6 +412,7 @@ CONFIG_MODULE(telegraf, Init, 0, 0, 0, Commit);
 // startup sequence
 CONFIG_REQUIRES(telegraf, cube_scan);
 CONFIG_REQUIRES(telegraf, kafka);
+CONFIG_REQUIRES(telegraf, mysql);
 
 // extra tunings
 CONFIG_OBSERVES(telegraf, cubesys, ParseCube, NotifyCube);
--- a/core/modules/config_telegraf.cpp
+++ b/core/modules/config_telegraf.cpp
@@ -45,6 +45,7 @@ static bool s_bCubeModified = false;
 static bool s_bKapacitorModified = false;
 
 static bool s_bConfigChanged = false;
+static bool s_bDbPassChanged = false;
 
 static CubeRole_e s_eCubeRole;
 static Configs cfg;
@@ -254,6 +255,8 @@ CommitCheck(bool modified, int dryLevel)
         return true;
     }
 
+    s_bDbPassChanged = s_dbPass.modified() | s_bCubeModified;
+
     s_bConfigChanged = modified | s_bCubeModified | s_bKapacitorModified |
                        G_MOD(CTRL_IP) | G_MOD(SHARED_ID);
 
@@ -264,15 +267,38 @@ CommitCheck(bool modified, int dryLevel)
 // owning tenant and instance. Give it a dedicated read-only account so the agent
 // keeps running as the unprivileged telegraf user instead of root.
 static bool
-SetupCheck(const std::string &dbPass)
+SetupCheck(const std::string &dbPass, bool dbPassChanged)
 {
     static const char *dbs[] = { "neutron", "nova", "keystone" };
 
+    // Create the account once and only write the password at creation. A bare
+    // GRANT ... IDENTIFIED BY would reset it on every commit, and again for each
+    // extra control node joining a converged cluster.
+    std::string user = "CREATE USER IF NOT EXISTS '" + std::string(DBUSER) +
+                       "'@'localhost' IDENTIFIED BY '" + dbPass + "'";
+    if (!MysqlUtilRunSQL(user.c_str())) {
+        HexLogError("failed to create the %s database account", DBUSER);
+        return false;
+    }
+
+    // Re-granting an existing privilege is a no-op, so this stays idempotent.
     for (size_t i = 0; i < sizeof(dbs) / sizeof(dbs[0]); i++) {
-        std::string sql = "GRANT SELECT ON " + std::string(dbs[i]) + ".* TO '" + DBUSER +
-                          "'@'localhost' IDENTIFIED BY '" + dbPass + "'";
+        std::string sql = "GRANT SELECT ON " + std::string(dbs[i]) + ".* TO '" +
+                          std::string(DBUSER) + "'@'localhost'";
         if (!MysqlUtilRunSQL(sql.c_str())) {
             HexLogError("failed to grant %s read access to %s", DBUSER, dbs[i]);
+            return false;
+        }
+    }
+
+    // An existing account keeps its old password, so rotate it explicitly when the
+    // tuning moves. MysqlUtilUpdateDbPass is not usable here: it also targets
+    // '<user>'@'%', which this localhost-only account does not have.
+    if (dbPassChanged) {
+        std::string sql = "SET PASSWORD FOR '" + std::string(DBUSER) +
+                          "'@'localhost' = PASSWORD('" + dbPass + "')";
+        if (!MysqlUtilRunSQL(sql.c_str()) || !MysqlUtilRunSQL("FLUSH PRIVILEGES")) {
+            HexLogError("failed to update the %s database password", DBUSER);
             return false;
         }
     }
@@ -329,7 +355,7 @@ Commit(bool modified, int dryLevel)
         if (IsControl(s_eCubeRole)) {
             std::string dbPass = GetSaltKey(s_saltkey, s_dbPass.newValue(), s_seed.newValue());
 
-            if (!SetupCheck(dbPass))
+            if (!SetupCheck(dbPass, s_bDbPassChanged))
                 return false;
 
             if (HexSystemF(0, "sed -e 's/@BASE@/%s/' -e 's/@UNIT@/%s/' -e 's/@THRESHOLD@/%d/' -e 's|@DBPASS@|%s|' %s > %s",
--- a/core/modules/config_telegraf.cpp
+++ b/core/modules/config_telegraf.cpp
@@ -20,7 +20,6 @@
 #include <cluster.hpp>
 
 #include "include/role_cubesys.h"
-#include "mysql_util.h"
 
 #define DELL_POWEREDGE_XR_MARKER "/etc/appliance/state/dell_poweredge_xr_detected"
 
@@ -37,15 +36,10 @@
 
 const static char NAME[] = "telegraf";
 
-// read-only account used by the sflow input to look up openstack port info
-static const char DBUSER[] = "telegraf";
-static const char DBPASS[] = "cIkpxxCLUFBbw4iO";
-
 static bool s_bCubeModified = false;
 static bool s_bKapacitorModified = false;
 
 static bool s_bConfigChanged = false;
-static bool s_bDbPassChanged = false;
 
 static CubeRole_e s_eCubeRole;
 static Configs cfg;
@@ -54,22 +48,14 @@ static Configs cfg;
 CONFIG_GLOBAL_STR_REF(CTRL_IP);
 CONFIG_GLOBAL_STR_REF(SHARED_ID);
 
-// private tunings
-CONFIG_TUNING_STR(TELEGRAF_DBPASS, "telegraf.db.password", TUNING_UNPUB, "Set telegraf database password.", DBPASS, ValidateRegex, DFT_REGEX_STR);
-
 // using external tunings
 CONFIG_TUNING_SPEC_STR(CUBESYS_ROLE);
-CONFIG_TUNING_SPEC_STR(CUBESYS_SEED);
-CONFIG_TUNING_SPEC_BOOL(CUBESYS_SALTKEY);
 CONFIG_TUNING_SPEC_STR(KAPACITOR_ALERT_FLOW_BASE);
 CONFIG_TUNING_SPEC_STR(KAPACITOR_ALERT_FLOW_UNIT);
 CONFIG_TUNING_SPEC_INT(KAPACITOR_ALERT_FLOW_THRESHOLD);
 
 // parse tunings
-PARSE_TUNING_STR(s_dbPass, TELEGRAF_DBPASS);
 PARSE_TUNING_X_STR(s_cubeRole, CUBESYS_ROLE, 1);
-PARSE_TUNING_X_STR(s_seed, CUBESYS_SEED, 1);
-PARSE_TUNING_X_BOOL(s_saltkey, CUBESYS_SALTKEY, 1);
 PARSE_TUNING_X_STR(s_alertFlowBase, KAPACITOR_ALERT_FLOW_BASE, 2);
 PARSE_TUNING_X_STR(s_alertFlowUnit, KAPACITOR_ALERT_FLOW_UNIT, 2);
 PARSE_TUNING_X_INT(s_alertFlowThreshold, KAPACITOR_ALERT_FLOW_THRESHOLD, 2);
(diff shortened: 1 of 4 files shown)


core/modules/config_telegraf.cpp
```diff
@@ -294,14 +327,25 @@ Commit(bool modified, int dryLevel)
         WriteDeviceConfig();
 
         if (IsControl(s_eCubeRole)) {
-            if (HexSystemF(0, "sed -e 's/@BASE@/%s/' -e 's/@UNIT@/%s/' -e 's/@THRESHOLD@/%d/' %s > %s",
+            std::string dbPass = GetSaltKey(s_saltkey, s_dbPass.newValue(), s_seed.newValue());
+
+            if (!SetupCheck(dbPass))
```
traviswu-bigstack: do we set password every bootstrap not just FTS? and how do we cover upgrade path?

Eandalf-Bigstack: Great catch! We did not guard re-entering this block for idempotency. Would push a commit to fix this.

Eandalf-Bigstack: Fixed in [cabb7bc](https://github.com/bigstack-oss/cubecos/pull/1257/commits/cabb7bc5413bff343eff0a43fc24e84593524f21).

Eandalf-Bigstack:
Answering the question raised about why the sflow plugin needs a MariaDB connection at all, given it "just" receives sFlow from OVN and forwards it to Kafka.

That description of **upstream** telegraf is exactly right — and that is precisely why the question is worth answering carefully. The MariaDB connection is **not upstream telegraf**. It is a Bigstack-only patch, and it exists purely to attribute each flow to an OpenStack tenant and instance.

## Provenance

```console
$ git grep -n -iE 'database/sql|go-sql-driver|mysql' v1.24.4 -- plugins/inputs/sflow/
(no output — upstream's sflow plugin has zero database code)
```

The DB code arrives with exactly one commit, ours:

```
e61a1da96  Travis Wu <travis.wu@bigstack.co>  feat(sflow): look up openstack port info
```

originally `576ff076` on `1.17.2-cube` (Feb 2021), carried forward to `1.24.4-cube` in this upgrade. The port and the conflict resolution are mine; the design being described below is @traviswu-bigstack's original work, so corrections welcome if I have mischaracterised the intent.

## The trace

**1. Every received packet refreshes the cache first** — [`plugins/inputs/sflow/sflow.go:116-128`](https://github.com/bigstack-oss/telegraf/blob/1.24.4-cube/plugins/inputs/sflow/sflow.go#L116-L128):

```go
n, _, err := conn.ReadFrom(buf)
if err != nil {
    ...
    break
}
s.cacheOpenstackPortInfo()   // ← added by the cube patch
s.process(acc, buf[:n])
```

**2. The cache is built from three OpenStack databases** — [`sflow.go:131-215`](https://github.com/bigstack-oss/telegraf/blob/1.24.4-cube/plugins/inputs/sflow/sflow.go#L131-L215). It is guarded, so it is a no-op once populated:

```go
func (s *SFlow) cacheOpenstackPortInfo() {
	if s.OpenstackDSN == "" || s.decoder.LatestCache {
		return
	}

	db, err := sql.Open("mysql", s.OpenstackDSN)
```

Two queries do the work:

- [`sflow.go:146`](https://github.com/bigstack-oss/telegraf/blob/1.24.4-cube/plugins/inputs/sflow/sflow.go#L146) joins `neutron.ports` × `nova.instances` × `keystone.project` — this is the **MAC → (tenant, instance)** mapping.
- [`sflow.go:177`](https://github.com/bigstack-oss/telegraf/blob/1.24.4-cube/plugins/inputs/sflow/sflow.go#L177) joins `neutron.floatingips` × `neutron.ports` to map floating IPs back to their fixed port.

The result is handed to the decoder at [`sflow.go:212`](https://github.com/bigstack-oss/telegraf/blob/1.24.4-cube/plugins/inputs/sflow/sflow.go#L212).

**3. Decode-time MAC match** — [`packetdecoder.go:331`](https://github.com/bigstack-oss/telegraf/blob/1.24.4-cube/plugins/inputs/sflow/packetdecoder.go#L331) calls into [`packetdecoder.go:345-380`](https://github.com/bigstack-oss/telegraf/blob/1.24.4-cube/plugins/inputs/sflow/packetdecoder.go#L345-L380):

```go
func (d *PacketDecoder) portCacheLookup(h *EthHeader) {
	// check if it's openstack port mac
	OspPortMacVender := "fa:16:3e:"          // the OpenStack/QEMU OUI
	...
	for i := range d.PortCache {
		p := d.PortCache[i]
		if SrcMac == p.MacAddr {
			h.TenantId = p.TenantId
			h.InstanceId = p.InstanceId
			h.Direction = "egress"           // src match ⇒ egress, dst match ⇒ ingress
			return
		}
	}
```

**4. The match becomes tags** — [`types.go:161-168`](https://github.com/bigstack-oss/telegraf/blob/1.24.4-cube/plugins/inputs/sflow/types.go#L161-L168):

```go
if len(h.PortType) > 0 {
	t["port_type"] = h.PortType
	t["tenant_id"] = h.TenantId
	t["tenant_name"] = h.TenantName
	t["instance_id"] = h.InstanceId
	t["instance_name"] = h.InstanceName
	t["direction"] = h.Direction
}
```

Which flow into the metric at [`metricencoder.go:30-38`](https://github.com/bigstack-oss/telegraf/blob/1.24.4-cube/plugins/inputs/sflow/metricencoder.go#L30-L38) and out to Kafka.

## What the official documentation says

Upstream's [sflow README](https://github.com/influxdata/telegraf/blob/v1.24.4/plugins/inputs/sflow/README.md) enumerates every tag the plugin produces — `agent_address`, `source_id_type`, `source_id_index`, `input_ifindex`, `output_ifindex`, `sample_direction`, `header_protocol`, `ether_type`, `src_ip`, `src_port`, `src_mac`, `src_vlan`, `dst_*`, `next_hop`, `ip_version`, `ip_protocol`, `ip_dscp`, `ip_ecn`, `tcp_urgent_pointer`.

**None of `tenant_id`, `tenant_name`, `instance_id`, `instance_name`, `port_type`, `direction` appear.** They are not upstream concepts.

The only three occurrences of the word "database" in that README are about cardinality load on the *output* store, not a data source:

> This plugin may produce a high number of series which, when not controlled for, will cause high load on your database.

So there is no upstream documentation for this behaviour, because upstream has no such behaviour. Our fork now documents it locally in [`sample.conf`](https://github.com/bigstack-oss/telegraf/blob/1.24.4-cube/plugins/inputs/sflow/sample.conf#L11-L15) and the regenerated README, which is an improvement on the previous state where the DSN was hardcoded and undocumented.

## Short answer

Receiving sFlow from OVN and forwarding to Kafka needs no database — correct. The DB is a **pure enrichment path**: sFlow carries MAC addresses, but the UI's Network and Device pages need to show *which tenant and which VM* the traffic belongs to, and that mapping only exists in the OpenStack control databases.

It also degrades cleanly now, which is worth stating explicitly: if `openstack_dsn` is unset or the DB is unreachable, `cacheOpenstackPortInfo` returns early, the cache stays empty, `portCacheLookup` matches nothing, and the `if len(h.PortType) > 0` guard simply omits those six tags. **Flow collection and Kafka delivery continue unaffected** — only the ownership attribution is lost. That is exactly the intended behaviour on non-control nodes, which do not host the databases.

Before this PR that was not true: the lookup failure was fatal (nil `*sql.Rows` dereference → SIGSEGV), which is [cube-cos-ui#557](https://github.com/bigstack-oss/cube-cos-ui/issues/557).

## If the objection is architectural

If the team would rather telegraf did not touch MariaDB at all, the legitimate alternatives are:

1. **Query the Neutron/Nova APIs** instead of the databases — proper API contract, but adds auth/token handling and per-interval API load.
2. **Move enrichment downstream** into whatever consumes `telegraf-metrics`, joining on `src_mac`/`dst_mac` there — keeps the agent dumb, at the cost of doing the join per-metric rather than per-cache-refresh.

Both are larger changes than this PR, which only moves the existing lookup off `root` and onto a `SELECT`-only account. Happy to open a follow-up story if that is the direction you would prefer.

Eandalf-Bigstack:
## Proposal on scope

I would like to suggest we keep the sFlow deprecation question out of this PR and let [#579](https://github.com/bigstack-oss/cubecos/issues/579) own it, with telegraf continuing to work against the sFlow functions that are already slated for removal.

The reasoning, in short:

**This PR is maintenance, not architecture.** What it actually does is move telegraf off a four-year-old build (1.17.2 → 1.24.4), stop the agent segfaulting when the OpenStack lookup fails, and drop that lookup from `root` to a `SELECT`-only account. It preserves the existing enrichment design rather than endorsing it. Redesigning where tenant attribution comes from is a materially larger change and would hold up a security bump behind an architecture decision.

**#579 already owns this, and has already made the call.** The epic explicitly evaluated the pattern this PR carries forward and decided against it:

> **OpenStack metadata** synced at cold-start through the **Neutron v2.0 REST API** (direct DB/MySQL access was considered and rejected — schema churn + blast radius; see DESIGN §C.3).

So the direct-DB coupling in the sFlow plugin already has a defined end-of-life, and it is not this PR. Re-opening it here would either duplicate that decision or pre-empt it.

**The replacement is not deployable yet.** In #579's own success criteria, **CubeCOS platform integration — systemd packaging + build pipeline** ([lachesis#113](https://github.com/bigstack-oss/lachesis/issues/113)) is still unchecked, as are IPv6 zone resolution, Octavia LB re-attribution, and netlink attach-presence robustness. Until lachesis is packaged into the product, sFlow is the only per-tenant network attribution that operators on current releases actually have. Removing or degrading it now would leave the Network and Device pages with no data source at all, which is the very symptom [cube-cos-ui#557](https://github.com/bigstack-oss/cube-cos-ui/issues/557) was raised about.

**Concretely, what I am proposing:**

1. Land this PR as a maintenance + security change, keeping the existing sFlow enrichment behaviour intact.
2. Track full sFlow deprecation — including retiring the MariaDB lookup entirely — under #579, sequenced behind lachesis' CubeCOS integration.
3. Treat the least-privilege account here as an interim hardening that reduces blast radius for however long sFlow remains in the product, rather than as a long-term endorsement of reading the OpenStack databases directly.

If the preference is instead to cut the DB dependency sooner without waiting for lachesis, the middle option is switching the existing enrichment to the Neutron v2.0 API — the same choice #579 made — which would be its own story rather than part of this upgrade. Happy to open one if that is the direction.

traviswu-bigstack:
The plan to release Lachesis is 3.2.0. why bother to do the patch? what we should do instead is to remove sflow completely from the current source base

Eandalf-Bigstack:
Got it, would drop the sflow support here.

traviswu-bigstack:
we forked telegraf for the sflow patch, maybe we take this chance to move back to the community telegraf version (i believe it's way much newer) and drop to maintain our own. (do it in the next release is still fine. it's your call)

Eandalf-Bigstack:
## Decisions

Thanks for the review — both points taken. Recording where we landed:

### 1. Keep the telegraf fork, for the security fixes only

We are keeping `bigstack-oss/telegraf` alive, but it now carries **exactly one commit over upstream `v1.24.4`**: the `golang.org/x/crypto` v0.31.0 pin ([`a167111`](https://github.com/bigstack-oss/telegraf/commit/a167111d1b628a8cc457aa0a646349d74f221fb1)). Upstream 1.24.4 still ships crypto from 2022-09-26, so the pin is doing real work.

The sflow patch is gone — [`993a5ca`](https://github.com/bigstack-oss/telegraf/commit/993a5ca10) reverts all three commits, and `plugins/inputs/sflow/` is now byte-identical to upstream. Notably its test suite passes for the first time since 2021; the 7 failures we had been carrying were caused entirely by that patch.

On your suggestion about returning to community telegraf and dropping the fork altogether — agreed in principle, and now much cheaper than it was, since the only thing standing in the way is that one pin. Worth doing as its own change in a later release rather than widening this one.

### 2. Full sflow removal belongs to the network-telemetry epic

Rather than remove sflow in this PR, we are leaving it to [#579](https://github.com/bigstack-oss/cubecos/issues/579), sequenced with Lachesis.

The reason is scope. Removing sflow "completely from the current source base" is not a config change — it reaches:

| File | What depends on sflow |
|---|---|
| [`core/sdk_sh/modules/sdk_security.sh`](https://github.com/bigstack-oss/cubecos/blob/openstack-develop/core/sdk_sh/modules/sdk_security.sh#L92-L107) | `security_flow_abnormal_detection` — per-tenant and per-instance traffic anomaly detection |
| [`core/sdk_sh/modules/sdk_stats.sh`](https://github.com/bigstack-oss/cubecos/blob/openstack-develop/core/sdk_sh/modules/sdk_stats.sh#L38-L48) | sflow series GC, grouped by `instance_id` |
| [`core/sdk_sh/modules/sdk_ovn.sh`](https://github.com/bigstack-oss/cubecos/blob/openstack-develop/core/sdk_sh/modules/sdk_ovn.sh#L221-L258) | `ovn_bridge_sflow_enable/disable`, OVS sampling config, the `sflow_enabled` state file |
| [`core/modules/cli_iaas_network.cpp`](https://github.com/bigstack-oss/cubecos/blob/openstack-develop/core/modules/cli_iaas_network.cpp#L423-L429) | the `iaas > flowdata > set_flowdata [on\|off]` CLI command |
| `core/sdk_sh/modules.pre/sdk_01-var-static.sh` | the `"hc"."sflow"` measurement reference |

Worth flagging explicitly: **`sdk_security.sh` and `sdk_stats.sh` consume tags that only the reverted patch produced** — they group by `tenant_name` / `instance_id` and iterate `direction` (ingress/egress) and `proto` (tcp/udp), none of which exist in upstream sflow. So those queries return nothing as of this revert, the loops iterate zero times, and flow anomaly detection stops running silently — no error, no log.

That is a real behaviour change this PR introduces, and it is worth someone deciding deliberately rather than discovering it later. It also means full removal deletes a security feature that has no replacement until Lachesis ships in 3.2.0, which is a product call and squarely epic-shaped rather than something to fold into a version bump.

### What this PR is now

Reduced to what survives both decisions:

- telegraf 1.17.2 → 1.24.4
- dependency security pins moved out of the build-time `go get` in `core/telegraf/Makefile` and into the fork's `go.mod`, so the tree describes what actually ships
- no sflow changes; `config_telegraf.cpp` and `telegraf-ctrl.conf.in` are byte-identical to `openstack-develop`

The review history for the reverted work is preserved — nothing was amended or force-pushed.

Eandalf-Bigstack:
@traviswu-bigstack Could you help review this PR?

traviswu-bigstack (approved):

PR #1267, opened 2026-08-15
fix(fts): zero-touch 3cc install fixes — manila NB-readiness guard, ovn-controller heal, git convergence, keycloak/airgap guards

### LINKED ISSUE #1272: [Bug] airgap_sim re-apply destroys CUBE_AIRGAP drop counters
## Description

`airgap_sim_apply` flushes and rebuilds the CUBE_AIRGAP chain on every call; the phone-home agent re-applies it periodically, so the DROP counters are zeroed before anyone can read them — egress violations during air-gap validation leave no evidence (observed as "0 drops" while a public curl was demonstrably blocked).

## How to Reproduce

1. Deploy with air-gap simulation; `curl https://1.1.1.1` from a node (hangs = blocked)
2. `iptables -vL CUBE_AIRGAP` → DROP counter 0; counters visibly reset between reads

## Expected behavior

Re-applying an already-correct chain leaves it (and its counters) untouched, so DROP counts accumulate as an audit trail.

## Definition of Done (DoD)

- [ ] airgap_sim_apply is idempotent when chain + OUTPUT hook match the expected shape
- [ ] DROP counters survive agent re-applies
- [x] Handbook knowledge update: the validation-round narrative is in the [live-resize blueprint](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/blueprints/nova-live-resize.md); extracted as a standalone known-issue with a scenario sidecar in [`kb/cubecos/known-issues/airgap-sim-reapply-resets-drop-counters.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/airgap-sim-reapply-resets-drop-counters.md)

### LINKED ISSUE #1273: [Bug] neutron-OVN sync does not repair missing security-group Port_Groups
## Description

When neutron operates while OVN NB is down/wedged, API-side objects can commit to the neutron DB while their NB twins never materialize (observed: a network's `Logical_Switch` and a security group's `Port_Group` missing after the FTS OVN gap). `hex_sdk migrate_neutron_ovn_sync` (neutron-ovn-db-sync) repaired the Logical_Switch but NOT the missing `Port_Group` (returned rc=0 with the PG still absent) — port creates then fail with `RuntimeError: Port group pg_<uuid> does not exist` / `MechanismDriverError`.

## How to Reproduce

1. Create a security group while OVN NB is unreachable (e.g., during the FTS gap of the zero-touch deadlock issue)
2. Restore OVN; run `hex_sdk migrate_neutron_ovn_sync`
3. Boot an instance using that SG → port create fails; `ovn-nbctl list Port_Group` lacks `pg_<sg-uuid>`

## Expected behavior

The sync reconciles security-group Port_Groups (and their ACLs) like other resources, or the repair path documents/automates the recovery (`ovn-nbctl pg-add` + touching a rule rebuilds ACLs — verified manually).

## Definition of Done (DoD)

- [ ] Sync (or a repair helper) restores missing Port_Groups
- [ ] Recovery documented as a kb known-issue/runbook until then
- [x] Handbook knowledge update: the validation-round narrative is in the [live-resize blueprint](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/blueprints/nova-live-resize.md); extracted as a standalone known-issue with a scenario sidecar in [`kb/cubecos/known-issues/ovn-sync-does-not-recreate-security-group-port-group.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/ovn-sync-does-not-recreate-security-group-port-group.md)

traviswu-bigstack:
#### What type of PR is this?

/kind bug

Fixes from the master-first zero-touch 3cc install validation (air-gapped, repair opted out, sky lab):

1. **airgap counters**: periodic airgap_sim_apply re-applies flushed the CUBE_AIRGAP chain, zeroing DROP counters and destroying egress-violation evidence — now idempotent.
2. **keycloak air-gap install guard**: build fails when the bundled registry lacks the exact `bigstack/keycloak:<chart tag>` image (the 17.2.0-vs-18.0.x drift that ImagePullBackOff'd every fresh install since Aug 12; also catches the image missing entirely, which an incremental build can produce).
3. **manila service-network ensure gated on OVN NB readiness**: the admin default security group is auto-created by neutron on the manila ensure's first write, which fires the instant the VIP answers — before pacemaker starts the OVN NB listener — so the SG lands without a port group and every later port create 500s (#1273). The ensure now probes `tcp:VIP:6641` and fails its pre-start until the NB is reachable (systemd retries it, off the FTS critical path). This prevents the strand at the source rather than repairing it after.
4. **set_ready_reconcile trigger**: ovn-controller SEGV-crash-loops into its systemd start limit during the pacemaker OVN bring-up window (upstream NULL-deref, unrelated to any cube change; see #1270). set_ready step 3 ("Starting and reconciling cluster") fires a `set_ready_reconcile` hex_config trigger on every node in parallel; config_neutron handles it role-aware and compute nodes heal their own ovn-controller. Idempotent. Validated on sky across 6 zero-touch reimages to 27/27.
5. **git background convergence**: /.git is infra plumbing, not an operator-facing service. The old per-node migrate loops ran the FULL cluster-wide git_init concurrently from every node (stomping each other's / worktrees) inside a 10-min window, and the ceph_mds health check then reported errcode 6 forever. Now each node converges only itself, unbounded, flock-guarded, in the background — and the health check no longer reports git at all.

Validated end-to-end on sky across four zero-touch driver reimages (airgap + repair opt-out), all reaching `cluster check` 27/27 with zero manual intervention. An A/B falsification run (identical branch with/without an ovn-northd handover change) showed the handover was unnecessary — the mid-round northd stop leaves NB/SB up and nothing in the round needs the translation — so it was dropped; the real FTS blockers were the keycloak image drift and the post-FTS inconsistencies the reconcile step heals. The ovn-controller SEGV is a race (13 crashes on one run, 0 on another); the reconcile heal covers both outcomes.

Full narrative + evidence: handbook `kb/cubecos/blueprints/nova-live-resize.md` (fresh-install findings section).

Fixes #1272, fixes #1273. Related: #1270 (root cause re-attributed to the keycloak drift + reconcile items by an A/B falsification run; see comments). Related: #1271 (build guard here; the keycloak 17/18 version pairing itself is tracked there)

The ovn-controller SEGV itself (crashes instead of retrying while SB/VIP are forming) deserves its own investigation — coredump preserved on sky141 from the first run. git convergence hot-validated on sky: with /.git wiped on a peer, health stays ok, the loop restores the clone in <1 min, setuid modes survive the re-init.

#### Additional documentation

```docs
bigstack-handbook kb/cubecos/blueprints/nova-live-resize.md
```

https://claude.ai/code/session_01BkeJmP4qdsDpNzyN6hQs25

### DIFF
--- a/core/keycloak/keycloak.mk
+++ b/core/keycloak/keycloak.mk
@@ -5,6 +5,17 @@
 KEYCLOAK_DIR := /opt/keycloak
 KEYCLOAK_CONF_DIR := /etc/keycloak
 
+# the bundled registry (core/docker/registry, copied into the rootfs) must
+# hold the keycloak image tag pinned in chart-values.yaml; a missing or
+# drifted tag means every fresh install ImagePullBackOffs -- fail the build.
+rootfs_install::
+	$(Q)charttag=$$(awk '/^  tag:/ {print $$2; exit}' $(COREDIR)/keycloak/chart-values.yaml); \
+	tagsdir=$(TOP_BLDDIR)/core/docker/registry/docker/registry/v2/repositories/bigstack/keycloak/_manifests/tags; \
+	if [ ! -d "$$tagsdir/$$charttag" ]; then \
+		echo "keycloak: bundled registry lacks bigstack/keycloak:$$charttag (has: $$(ls $$tagsdir 2>/dev/null | tr '\n' ' ')) -- fresh installs would ImagePullBackOff" >&2; \
+		exit 1; \
+	fi
+
 rootfs_install::
 	$(Q)chroot $(ROOTDIR) mkdir -p $(KEYCLOAK_DIR)
 	$(Q)cp -f $(COREDIR)/keycloak/chart-values.yaml $(ROOTDIR)/$(KEYCLOAK_DIR)/
--- a/core/modules/cli_cluster.cpp
+++ b/core/modules/cli_cluster.cpp
@@ -864,8 +864,10 @@ ClusterReadyMain(int argc, const char** argv)
     else
         HexSpawn(0, HEX_CFG, "-p", "trigger", "cluster_ready", ZEROCHAR_PTR);
 
-    CliPrintf("[3/6] Starting cluster");
+    CliPrintf("[3/6] Starting and reconciling cluster");
     HexUtilSystemF(0, 0, HEX_SDK " host_local_run hex_sdk cmd " HEX_SDK " cube_cluster_start");
+    // parallel per-node heals (ovn-controller start-limit recovery)
+    HexUtilSystemF(0, 0, "cubectl node exec -pn \"" HEX_CFG " trigger set_ready_reconcile\"");
 
     // the cluster is guaranteed complete here -- never import at boot, where a
     // still-forming ceph wedges the upload (#1139)
--- a/core/modules/config_neutron.cpp
+++ b/core/modules/config_neutron.cpp
@@ -967,6 +967,25 @@ DisableSflowMain(int argc, char* argv[])
     return EXIT_SUCCESS;
 }
 
+static int
+SetReadyReconcileMain(int argc, char **argv)
+{
+    if (!s_enabled)
+        return EXIT_SUCCESS;
+
+    // ovn-controller (compute-side) can SEGV-crash-loop into its systemd
+    // start limit while the SB DB / VIP are still forming; heal this node.
+    // Fired in parallel on every node from set_ready.
+    if (IsCompute(s_eCubeRole)) {
+        HexUtilSystemF(0, 0,
+                       "systemctl reset-failed ovn-controller 2>/dev/null; "
+                       "systemctl is-active ovn-controller >/dev/null || "
+                       "systemctl restart ovn-controller");
+    }
+
+    return EXIT_SUCCESS;
+}
+
 CONFIG_COMMAND_WITH_SETTINGS(restart_neutron, RestartMain, RestartUsage);
 CONFIG_COMMAND_WITH_SETTINGS(enable_sflow, EnableSflowMain, EnableSflowUsage);
 CONFIG_COMMAND_WITH_SETTINGS(disable_sflow, DisableSflowMain, DisableSflowUsage);
@@ -986,6 +1005,8 @@ CONFIG_OBSERVES(neutron, rabbitmq, ParseRabbitMQ, NotifyMQ);
 CONFIG_OBSERVES(neutron, cubesys, ParseCube, NotifyCube);
 CONFIG_OBSERVES(neutron, nova, ParseNova, NotifyNova);
 CONFIG_OBSERVES(neutron, keystone, ParseKeystone, NotifyKeystone);
+
+CONFIG_TRIGGER_WITH_SETTINGS(neutron, "set_ready_reconcile", SetReadyReconcileMain);
 
 CONFIG_MIGRATE(neutron, "/etc/openvswitch/");
 CONFIG_MIGRATE(neutron, "/var/lib/ovn");
--- a/core/sdk_sh/modules/errcodes
+++ b/core/sdk_sh/modules/errcodes
@@ -135,7 +135,6 @@ eval "ERR_DESC_$SRV[2]='cephfs not all mounted'"
 eval "ERR_DESC_$SRV[3]='nfs-ganesha fail to mount'"
 eval "ERR_DESC_$SRV[4]='nfs-ganesha fail to write'"
 eval "ERR_DESC_$SRV[5]='nfs-ganesha not all up'"
-eval "ERR_DESC_$SRV[6]='git not initialied'"
 eval for idx in \"\${!ERR_DESC_${SRV}[@]}\" \; do ERR_CODE[${SRV},\$idx]=\"\${ERR_DESC_${SRV}[\$idx]}\" \; done
 
 SRV="ceph_osd"
--- a/core/sdk_sh/modules/sdk_git.sh
+++ b/core/sdk_sh/modules/sdk_git.sh
@@ -149,6 +149,25 @@ git_init()
     git_client_init
 }
 
+git_node_init()
+{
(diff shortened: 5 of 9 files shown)


Eandalf-Bigstack:
Heads-up on `c0048d7 build(keycloak): assert login image tag matches the chart pin` — as written the assert can never fail, so #1271's "drift fails the build" DoD isn't actually covered yet.

The check reads the tag out of the rootfs:

```make
rpmver=$(chroot $(ROOTDIR) rpm -q --qf '%{VERSION}' cube-cos-login 2>/dev/null); \
charttag=$(awk '/^  tag:/ {print $2; exit}' $(COREDIR)/keycloak/chart-values.yaml); \
if [ -n "$rpmver" ] && [ "$rpmver" != "$charttag" ]; then ... exit 1; fi
```

but `cube-cos-login` is never installed in the rootfs. `core/keycloak/Makefile` installs it on the **build jail**, extracts `keycloak-image.tar` with `podman load`, and removes it again — the rpm only exists to carry the image:

```make
.load-img: $(LOGIN_RPM)
	$(Q)dnf install -y $< $(QEND)
	$(Q)podman load -i /usr/share/cube/login/keycloak-image.tar $(QEND)
	$(Q)dnf remove -y $(PKG_NAME) $(QEND)
```

So `rpmver` comes back empty, the `[ -n "$rpmver" ]` guard short-circuits, and the assert passes silently — including on exactly the drift it is meant to catch. Verified against a completed build in the `centos9_cubecos_jim_200.13` jail:

```
$ chroot heavy_rootfs rpm -q cube-cos-login
package cube-cos-login is not installed
$ chroot heavy_rootfs rpm -qa | grep -ci keycloak
0
```

Two sources in the build tree do carry the real tag, and both agree today:

```
build/core/keycloak/version    -> 18.0.0        # what IMG_BUILD interpolates when tagging
build/core/keycloak/.load-img  -> localhost:5080/bigstack/keycloak:18.0.0
core/keycloak/chart-values.yaml tag -> 18.0.0
```

`.load-img` is the stronger of the two, since it records the tag that actually reached the registry (`echo $(IMG_BUILD) > $@`), which is what the chart later tries to pull.

One trap for the fix: `core/keycloak/Makefile` hardcodes a different tag for RC builds —

```make
ifneq ($(RC),)
IMG_BUILD := localhost:5080/bigstack/keycloak:17.1.0
else
IMG_BUILD := localhost:5080/bigstack/keycloak:`cat version`
endif
```

so comparing the pushed tag against the chart pin will fail every RC build unless RC is skipped or special-cased.

Found while chasing the accept-pipeline FTS failures on `jim.lin/fix/release-wrapup`; the keycloak 18 chart pin there is `18.0.0` and matches the registry, so #1271's drift itself is resolved on that branch — this is only about the guard that is supposed to keep it from coming back. Happy to push a corrected assert against `.load-img` if that's useful, or leave it with you.

traviswu-bigstack:
Thanks Jim — you were right, and the assert you reviewed could indeed never fire. I hit the same thing live overnight (plus a second trap: `rpm -q` prints "package … is not installed" on **stdout**, so a naive `2>/dev/null` capture turns it into a bogus non-empty version that false-positives the other way).

The assert has since been reworked on this PR (`a1a61a2`): it now checks the **bundled registry tags dir** (`$(TOP_BLDDIR)/core/docker/registry/.../bigstack/keycloak/_manifests/tags/<charttag>`) rather than `.load-img`. That's one step stronger than the pushed ref: it also catches the image being **absent entirely**, which incremental builds can genuinely produce — the `cubectl → keycloak install` chain doesn't re-run on incremental builds, and I caught a build shipping a keycloak-less registry exactly that way.

On the RC trap: this assert is under `rootfs_install::`, and the RC heavyfs path runs `heavyfs_install` instead (`core/heavyfs/Makefile`), so RC builds skip it by construction — no special-casing needed, though a comment noting that is fair.

Good news that release-wrapup aligns chart 18.0.0 with the 18.0.0 image — that's the real #1271 fix and this guard should sit quietly on top of it. (The 17.2.0 pin I used for overnight lab builds was a build-branch-only TEMP and dies with #1271.)

Eandalf-Bigstack:
Since the release window of CubeCOS v3.1.10 is closing, I would like to defer this PR to v3.1.20. We could come back to this PR after the releasing of v3.1.10-rc1 and switching v3.1.10 development onto branch release_3.1.10. Please let me know if the feature and fix implemented must go into v3.1.10.

traviswu-bigstack:
This is build and FTS hygien fixes. It adds no features. I would still prefer to merge it into 3.1.10. And i think we missed one important component for the toA upgrade. its ovn + ovs. I created 3 issues for toA, toC, and toE separatedly. I am working on the toA issue. I think that one is required

traviswu-bigstack:
**Clarification — OVN/OVS are intentionally *not* bumped in this release.** 3.1.10 (Antelope) stays on **OVN 23.03 / OVS 3.1**, which is the version-correct pairing for Antelope (RDO `rdo-openvswitch/3.1/3.el9s → rdo-ovn-23.03`). The **OVN 24.03 / OVS 3.3** upgrade is the **Caracal** pairing and is deferred to that hop — tracked in #1276 (PR #1281); the Antelope-hop issue #1275 is closed as no-change.

This is why the ovn-controller SEGV in this PR is addressed with the `set_ready_reconcile` **heal**, not by upgrading OVN: the crash is an upstream NULL-deref (`pflow_output_get_debug`) that a full-tree sweep confirmed **persists in OVN 24.03, 25.03, and 26.03** (patch mailed to ovs-dev), so bumping OVN would not remove it — the heal is required at any version. Deferring the bump also keeps this release's change surface minimal and correct for Antelope.

Eandalf-Bigstack:
Got it.

Eandalf-Bigstack (approved):

Issue #1263, opened 2026-08-14
[Bug]: A pgpu card can never be switched to vGPU through the API — supportTypes goes stale under vfio-pci


SekiXu:
## Describe the bug

A GPU card whose resource type is `pgpu` reports `supportTypes: ["pgpu"]` and an **empty**
vGPU profile list, even on hardware that fully supports SR-IOV and MIG-backed vGPU. Because
`cube-cos-api` gates the resource-type update on `supportTypes`, `PUT
/nodes/:node/gpuCards/:gpuId` answers **409** for every attempt to move that card to
`sriovVgpu` or `migBackedVgpu`.

There is no way around it from the API: `gpu_resource_set` accepts only `pgpu`,
`sriovVgpu` and `migBackedVgpu` — there is no `unset` — so the card cannot be parked in a
neutral state first. **Through the UI/API a pgpu card is a one-way door.** Only calling
`hex_config` directly on the node still works.

The card has not lost any capability. Switching it with the CLI succeeds, and immediately
afterwards the same field reports all three types.

## Affected Component / Service

**The defect is in `cubecos`.** `supportTypes` is a capability field, and the component
that produces it — `hex_sdk gpu_device_list` — reports a value that depends on the card's
current mode.

- `cubecos` — `core/sdk_sh/modules/sdk_gpu.sh`, `gpu_device_list()` and
  `gpu_vgpu_profile_list()`. **This is where the fix belongs.**
- `cube-cos-api` — `internal/apis/v1/handlers/nodes/gpu.go`, `validateGpuCardUpdate()`.
  Not defective: rejecting an unsupported resource type with 409 is correct behaviour given
  a correct input. It is listed only because it is where the stale value becomes a hard
  failure for the user.
- Not affected: `hex_config gpu_resource_set` itself, which already handles this correctly.

## Version

- Product: CubeCOS 3.1.0 (rootfs snapshot `CUBE_3.1.0_20260421-024423`)
- `cube-cos-api`: `v3.1.0-1.el9.5310e06`
- Verified on: cn13 (all-in-one, 4× NVIDIA RTX PRO 6000 Blackwell Server Edition)

## To Reproduce

1. Pick a card whose type is `pgpu` and whose status is `idle`.
2. `hex_sdk gpu_device_list` — the card reports `"supportTypes": ["pgpu"]`.
3. `hex_sdk gpu_vgpu_profile_list <gpuId>` — returns `{"sriov":[],"migBacked":[]}`.
[...]

Issue #1262, opened 2026-08-14
[Bug]: cyborg advertises vGPU VFs and carved PFs as allocatable PGPU, so pgpu scheduling collides with running VMs


SekiXu:
## Describe the bug

cyborg's GPU inventory in Placement does not reflect what the hardware is actually doing.
On a node with four GPUs it registers **148 resource providers offering `PGPU`, of which
exactly one can really serve a passthrough request.** The rest are SR-IOV VFs, and PFs that
`hex_config` has already carved into vGPU.

The clearest demonstration — three sources, same instant, same device:

| Source | What it says about `0000:42:00.2` |
|---|---|
| libvirt | `<hostdev>` … `bus='0x42' slot='0x00' function='0x2'` — attached to a running VM |
| nova DB | `0000:42:00.2  allocated  d44d6e9f-d282-42f3-bbd1-4a557e8fb8f0` |
| **Placement** | **`PGPU  total=1  used=0`** — free |

Any `resources:PGPU=1` request may be scheduled onto that VF and will then fail in libvirt.

This is not a race that needs pgpu and vGPU to be mixed on one host: a node carrying
**only** vGPU cards still exposes ~150 bogus `PGPU` providers.

## Affected Component / Service

- `cubecos` — `core/modules/config_cyborg.cpp` (enables the NVIDIA driver with no device
  filtering), `core/sdk_sh/modules/sdk_gpu.sh` (the carving path, which never informs
  cyborg).
- OpenStack cyborg agent (`cyborg-agent.service`) and Placement.

## Version

- Product: CubeCOS 3.1.0 (rootfs snapshot `CUBE_3.1.0_20260421-024423`)
- Verified on: cn13 (all-in-one, 4× NVIDIA RTX PRO 6000 Blackwell Server Edition)

## To Reproduce

1. Carve at least one card into a vGPU type:

   ```bash
   hex_config gpu_resource_set <gpuId> sriovVgpu '[{"id":1528,"count":1}]'
   ```

2. Compare hex's view with Placement's:

   ```bash
   hex_sdk gpu_device_list | jq -r '.[] | "\(.pciAddress) \(.type)"'
   openstack resource provider list -f value -c name | grep -c '^cn13_'
   ```

   ```text
   hex                                    placement
   00000000:42:00.0  sriovVgpu            cn13_0000:42:00.0  PGPU total=1 used=0
[...]

Issue #828, opened 2026-04-16
[User Story] Verify K8S GPU Operator Could Request GPU Resources within VM with GPU Resources


Eandalf-Bigstack:
## Description

As a CubeCOS operator, I want workloads on K8S could leverage GPU resources so that we could provide GPU computation resources to K8S workloads.

## Acceptance Criteria

> A detailed list of conditions that must be met for the story to be accepted. These statements must focus on functional behavior and constraints rather than specific code-level implementation.

We could verify this by using CUDA vector-add test.

- [x] VM with pGPU
- [x] VM with SR-IOV vGPU
- [x] VM with MIG-backed vGPU

## Story Points: 3

>| Points | Effort / Resources | Description |
> | -- | -- | -- |
>| 1 | Low | Low complexity; can be resolved easily.
>| 3 | Medium | Moderate workload; requires some focus, but risks are well-managed.
>| 5 | High | High complexity; difficult to implement, requiring significant time and effort.
>| 8 | Very High | Extreme complexity or high uncertainty. If possible, this should be divided into multiple user stories.

steven-chiu-bigstack:
From @Eandalf-Bigstack:

>This would need the developer to setup a K8S cluster either using App Framework 2.0 or the cube-saas cluster setup script of cube-license-manager. Then we could specify the worker machine pool with the GPU VM flavor. After the VM is successfully created, we could verify if CUDA containers could utilize the passed in GPU card on the K8S cluster.

Eandalf-Bigstack:
The following could be leveraged for setting up a K8S cluster on CubeCOS.

- https://github.com/bigstack-oss/cubecos-private/issues/62
- The folder (./infra) which contains the script for setting up a K8S cluster. https://github.com/bigstack-oss/cube-license-manager/tree/jim.lin/feat/deploy-to-new-home/infra
- The official way in CubeCOS to set up a K8S cluster. https://github.com/bigstack-oss/cube-cos-app-framework

PR #1280, opened 2026-08-17
fix(gpu): stop advertising GPU resources that cannot be delivered

### LINKED ISSUE #1264: [Bug]: The vgpu.example flavor created at cluster set_ready demonstrates a mechanism CubeCOS does not implement
## Describe the bug

`hwdetect_cluster_ready` creates two example flavors alongside the `t2.*` set. Neither can
boot a VM, and the first one is actively misleading:

```bash
openstack flavor create --vcpus 2 --ram 4096 --disk 40 --public vgpu.example
openstack flavor set vgpu.example --property "resources:VGPU=1"

openstack flavor create --vcpus 2 --ram 4096 --disk 40 --public pgpu.example
openstack flavor set pgpu.example --property 'accel:device_profile=<profile_name>'
```

- **`vgpu.example`** asks for `resources:VGPU=1`. That is Nova's **mediated device (mdev)**
  mechanism. CubeCOS does not use mdev — `hex_config gpu_resource_set` hands whole SR-IOV
  **VFs** to Nova through a PCI alias. Consequently **nothing registers `VGPU` inventory on
  any resource provider**, and a request for `resources:VGPU=1` can never be satisfied.
- **`pgpu.example`** carries the literal string `<profile_name>`, angle brackets included —
  an unsubstituted placeholder.

An example flavor needing parameters filled in would be fine. The problem is that
`vgpu.example` demonstrates **the wrong mechanism**: anyone who copies it gets
`NoValidHost` and then goes looking at the scheduler, host aggregates or quota — none of
which is where the problem is.

`cubecos` — `core/sdk_sh/modules/sdk_hwdetect.sh`, `hwdetect_cluster_ready()` lines 29-32.

## Version

## To Reproduce

1. On any cluster that has completed `cluster set_ready`:

   ```bash
   openstack flavor show vgpu.example -f json | jq -c .properties
   # {"resources:VGPU":"1"}
   ```

2. Look for a resource provider that actually offers `VGPU`:

   ```bash
   for u in $(openstack resource provider list -f value -c uuid); do
     openstack resource provider inventory list "$u" -f value -c resource_class |
       grep -qx VGPU && echo "$u has VGPU"
   done
   # (no output — no provider anywhere offers VGPU)
   ```

3. Booting a server with `--flavor vgpu.example` therefore fails with `NoValidHost`.

## Expected behavior

The example flavors either demonstrate the mechanism CubeCOS actually implements, or they
are not created at all.

For SR-IOV / MIG-backed vGPU the working form is a Nova PCI alias, and the alias name is
generated by `hex_config` into `/etc/nova/nova.d/gpu.conf`:

```text
alias = { "name": "nvidia_rtx_pro_6000_blackwell_dc_12q_1528", "device_type": "type-VF", ... }
```

```bash
openstack flavor set <flavor> \
  --property "pci_passthrough:alias=nvidia_rtx_pro_6000_blackwell_dc_12q_1528:1"
```

## Root Cause (if known)

The example predates the implementation it appears to describe.

- `vgpu.example` was added in `b48c12fa` (2025-03-15), the commit that created
  `sdk_hwdetect.sh`.
- The GPU resource-type implementation (`gpu_resource_set`, the `sriovVgpu` /
  `migBackedVgpu` paths) first appears in **2026-06**.

So the flavor is a placeholder written **15 months before** the design existed. It was
never revisited when the PCI-alias implementation landed.

Both flavors are created with `Quiet -n`, so an operator has no indication of where they
came from or that they are illustrative only.

## Suggested Fix (Optional)

**Remove both example flavors.**

They cannot be made correct where they are created, because **neither property has a valid
value at `cluster set_ready` time**:

- `pci_passthrough:alias=<name>:1` — the alias is generated by `hex_config` into
  `/etc/nova/nova.d/gpu.conf` when a card is carved, and its name encodes the chosen
  profile (`nvidia_rtx_pro_6000_blackwell_dc_12q_1528`). At `set_ready` no card has been
  carved, so no alias exists.
- `accel:device_profile=<name>` — device profiles are created by the operator afterwards.
  None exists at `set_ready` either.

So a *correct* example flavor is structurally impossible at that point in the lifecycle: any
value written there is either wrong (today's case) or a placeholder that cannot be filled
in automatically. Creating them silently with `Quiet -n` makes it worse, because an operator
finds two flavors with no indication that they are illustrative.

If the didactic value is wanted, it belongs in documentation next to the carving procedure,
where the alias name is known — not in a flavor created 15 months before that procedure
exists.

## Additional context

Found while writing up #828. Severity is low — nothing breaks by itself — but the failure
it induces (`NoValidHost`) points investigators away from the real cause.

SekiXu:
Two independent fixes for the same underlying problem: CubeCOS advertising GPU
resources it cannot actually deliver. Both were found while verifying #828.

## fix(cyborg): report only cards hex assigned to passthrough — #1262

`discover_gpus()` labelled every NVIDIA function `lspci` reports as a GPU with
`rc=PGPU`, unconditionally. SR-IOV vGPU VFs report as `3D controller` exactly
like the card itself, so each of them was advertised as a whole card; and a
card already carved into `sriovVgpu` / `migBackedVgpu` stayed advertised as
available for passthrough. Either one lets the scheduler place a
`resources:PGPU=1` request onto a device another VM already holds — placement
accepts it, libvirt then refuses with
`PCI device <addr> is in use by driver QEMU`.

The filter uses two facts the product already owns, so no new state is
introduced: a VF carries a `physfn` link in sysfs, and
`/etc/cube/cos/gpu/config.json` records the type hex assigned to each card.
That file is root-owned `0600` and the agent runs as `cyborg`, so it is read
through the same privsep entrypoint `lspci` already uses — **no new
privilege**. If it cannot be read, the driver reports nothing rather than
falling back to the old behaviour: under-reporting fails scheduling cleanly,
over-reporting hands out a device that is in use.

## fix(gpu): stop creating example flavors that can never boot — #1264

`vgpu.example` asked for `resources:VGPU=1` — Nova's mediated-device
mechanism. CubeCOS hands whole VFs to Nova through a PCI alias and never
registers `VGPU` inventory, so that flavor could not be scheduled anywhere.
`pgpu.example` carried the literal string `<profile_name>`.

Neither could be made correct where they were created: a PCI alias name is
minted by `gpu_resource_set` when a card is carved, and a cyborg device
profile is created by the operator afterwards — at `set_ready` neither exists.
A comment now records why, so they do not come back.

## Verification

Deployed to **cn13** (all-in-one, 4× RTX PRO 6000 Blackwell — one `pgpu`, one
`sriovVgpu`, two `migBackedVgpu`) and exercised against the live cluster.

`#1262`, before and after restarting `cyborg-agent`:

| | resource providers named `cn13_*` |
|---|---|
| before | **148** offering `PGPU`, one of which could serve it |
| after | **1** |

The three carved cards and every VF were withdrawn, **placement reconciled the
stale providers away on its own**, the surviving provider kept its inventory
(`PGPU total=1`) and its traits, and the two running vGPU VMs kept their VFs
attached (`0x42:00.2`, `0x8c:00.2`). No traceback in `cyborg-agent`; all six
VMs stayed `ACTIVE`.

`#1264` is only reachable at `cluster set_ready`, which has already run on
every existing cluster, so it **cannot be exercised end to end on cn13**. What
was checked: `bash -n` clean on the node, `hex_sdk` still loads the module,
`hwdetect_cluster_ready` still parses, and no `*.example` creation path
remains in the deployed file.

## Notes for the reviewer

- **Existing clusters keep the two example flavors.** Removing them was left
  out deliberately — deleting flavors is destructive and belongs in an upgrade
  step, not here.
- The cyborg change ships as a `yoga_patch` file with its `.orig` alongside,
  matching the existing `api/controllers/v2/arqs.py` convention.
- Upstream carries a `TODO(yumeng) support and test VGPU rc generation soon`
  at the same spot, so this is a gap upstream knows about rather than a
  behaviour anyone depends on.
- The third issue found in the same pass, #1263 (`supportTypes` goes stale
  under vfio-pci), is **not** in this PR — it is #1282, which can be reviewed
  and merged independently of this one; the two touch different files.

Fixes #1264
Refs #1262

**#1262 is deliberately left open.** This PR implements the first half of its
suggested fix — the wrong inventory. The second half, reconciling a stale
`Bound` ARQ after a host reboot (a pgpu VM comes back with no device attached
while cyborg and placement still show it allocated), is untouched and needs
its own change.

### DIFF
--- a/core/cyborg/yoga_patch/accelerator/drivers/gpu/utils.py
+++ b/core/cyborg/yoga_patch/accelerator/drivers/gpu/utils.py
@@ -1,0 +1,274 @@
+# Copyright 2018 Beijing Lenovo Software Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+"""
+Utils for GPU driver.
+"""
+from oslo_concurrency import processutils
+from oslo_log import log as logging
+from oslo_serialization import jsonutils
+
+import os
+import re
+
+from cyborg.accelerator.common import utils
+from cyborg.common import constants
+from cyborg.conf import CONF
+from cyborg.objects.driver_objects import driver_attach_handle
+from cyborg.objects.driver_objects import driver_attribute
+from cyborg.objects.driver_objects import driver_controlpath_id
+from cyborg.objects.driver_objects import driver_deployable
+from cyborg.objects.driver_objects import driver_device
+import cyborg.privsep
+
+LOG = logging.getLogger(__name__)
+
+GPU_FLAGS = ["VGA compatible controller", "3D controller"]
+GPU_INFO_PATTERN = re.compile(r"(?P<devices>[0-9a-fA-F]{4}:[0-9a-fA-F]{2}:"
+                              r"[0-9a-fA-F]{2}\.[0-9a-fA-F]) "
+                              r"(?P<controller>.*) [\[].*]: (?P<model>.*) .*"
+                              r"[\[](?P<vendor_id>[0-9a-fA-F]"
+                              r"{4}):(?P<product_id>[0-9a-fA-F]{4})].*")
+
+VENDOR_MAPS = {"10de": "nvidia", "102b": "matrox"}
+
+# CubeCOS owns GPU resource-type configuration through hex_config. This file is
+# its source of truth: one entry per physical card, with the type the operator
+# assigned (pgpu / sriovVgpu / migBackedVgpu). A card that is not recorded here
+# as "pgpu" is not available for whole-card passthrough, so cyborg must not
+# advertise it.
+HEX_GPU_CONFIG_FILE = "/etc/cube/cos/gpu/config.json"
+SYS_PCI_DEVICES = "/sys/bus/pci/devices"
+
+
+@cyborg.privsep.sys_admin_pctxt.entrypoint
+def lspci_privileged():
+    cmd = ['lspci', '-nnn', '-D']
+    return processutils.execute(*cmd)
+
+
+@cyborg.privsep.sys_admin_pctxt.entrypoint
+def read_hex_gpu_config_privileged():
+    """Read hex's GPU truth file. It is root-owned 0600; the agent is not."""
+    with open(HEX_GPU_CONFIG_FILE, 'r') as f:
+        return f.read()
+
+
+def _normalize_pci_address(address):
+    """hex writes an 8-digit domain and upper-case bus ("00000001:C8:00.0");
+    lspci -D prints a 4-digit domain in lower case ("0001:c8:00.0"). Compare
+    on the lspci form.
+    """
+    parts = address.strip().lower().split(':')
+    if len(parts) != 3:
+        return address.strip().lower()
+    try:
+        domain = int(parts[0], 16)
+    except ValueError:
+        return address.strip().lower()
+    return "%04x:%s:%s" % (domain, parts[1], parts[2])
+
+
+def get_passthrough_addresses():
+    """Addresses of the cards hex has assigned to whole-card passthrough.
+
+    Returns None when the truth file cannot be read, which the caller treats as
+    "report nothing": under-reporting fails scheduling cleanly, while
+    over-reporting hands out a device another VM already holds.
+    """
+    try:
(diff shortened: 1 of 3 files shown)


Eandalf-Bigstack:
I would like to defer this PR after #1266 since patching the Antelope version Cyborg would be more meaningful for CubeCOS v3.1.10. Also, this could be going into v3.1.10 to prevent GPU VM from v3.1.0 being impacted with the fight between our new GPU mechanism and OpenStack Cyborg.

SekiXu:
Agreed — deferring until after #1266 and targeting v3.1.10 makes sense. Confirmed the `yoga_patch/` file would stop applying once the release moves to Antelope, so there is nothing to salvage here.

Closing this PR. The other commit in it (`sdk_hwdetect.sh`, #1264) is unrelated to cyborg and the OpenStack release, so I'll raise that separately.

SekiXu:
Deferred until after #1266; the cyborg patch will be redone against antelope_patch. Findings moved to #1262, and the unrelated sdk_hwdetect.sh change is raised separately.

Issue #1275, opened 2026-08-17
OVN version for the Antelope hop (23.03/OVS 3.1 — already version-correct, no bump)


traviswu-bigstack:
#### What this issue does / why we need it

OVN has been pinned at **ovn23.03 / openvswitch3.1** (`core/neutron/neutron.mk`) since the initial cube import (2025-02) and never bumped. This story tracks the OVN upgrade + validation for the **Antelope hop** of the Yoga→Epoxy ladder, mirroring the qemu one-story-per-hop pattern (#1150–#1152).

Backed by a real rig spike (`/root/migration-rig/ovn-spike-results.md`): c9s NFV SIG carries 22.03–26.03; chassis-first skew is the supported/tested direction; NB/SB schema conversion is in-place and lossless across 23.03→24/25/26.03; the ovn-controller SEGV (see #1270) is **not** fixed by any of these versions.

**This hop carries the concrete bump: 23.03 → 24.03** (the current LTS; single schema step 7.0.0→7.3.0, proven clean in the spike). Conservative target for the first move off 23.03.
#### Rolling-upgrade contract (mandatory)

- **Chassis first, central last**: upgrade ovn-host node-by-node (new-chassis/old-central is the tested skew); flip the pacemaker `ovndb_servers` (NB/SB + northd) last, with the ovsdb schema conversion as its own gated step.
- **Back up** ovnnb_db.db / ovnsb_db.db before the central flip — schema conversion is one-way.

#### Definition of Done

- [ ] Target OVN version resolved on the built Antelope image (`dnf --showduplicates`), pinned in neutron.mk.
- [ ] Rig prune: mixed-version control-plane matrix green (chassis registers, bindings kept, schema converts).
- [ ] **Cluster roll certification** (the authoritative sign-off, not rig-certifiable): real VMs keep networking through the node-by-node chassis roll + central flip; rollback rehearsed from a backed-up DB.
- [ ] set_ready_reconcile SEGV heal still present (bug persists at target version).

Spike + evidence: handbook `kb/cubecos/blueprints/nova-live-resize.md` and `/root/migration-rig/ovn-spike-results.md`.

traviswu-bigstack:
**No OVN bump needed for the Antelope hop — cube already ships the version-correct pairing.**

Verified the per-release OVN/OVS pairing against RDO packaging (what cube consumes) + release cadence:

| OpenStack release | OVN | Open vSwitch |
|---|---|---|
[...]

Issue #628, opened 2026-01-27
[Story] Upgrade Neutron from Antelope to Caracal


jdjgya:
## Description

As a CubeCOS operator, I want Neutron be in Caracal so that we could upgrade to Caracal. Also, we would need to upgrade OVN to resolve the security concern regarding OVN 23.03.1 EOL.

https://docs.openstack.org/releasenotes/neutron/2024.1.html

## Acceptance Criteria

> A detailed list of conditions that must be met for the story to be accepted. These statements must focus on functional behavior and constraints rather than specific code-level implementation.

- [x] Upgrade OVN from 23.03.1 to 24.03.x
- [x] Migrate configs
- [x] Adjust service arrangements to fulfill basic functionalities
  - [x] Ensure network, subnet, router, and interface creations and deletions are working.
- [x] Ignore the new features introduced in the new version

Delivered by **[PR #1315](https://github.com/bigstack-oss/cubecos/pull/1315)** — Neutron 22.2.1 → 24.2.2, moved from `/opt/openstack-antelope` (python 3.10) into `/opt/openstack-caracal` (python 3.11), the venv Skyline, Keystone, Glance, Cinder, Nova and Placement already occupy. Ten packages added, **zero** version changes to anything the other six hold. The OVN half builds on **[PR #1281](https://github.com/bigstack-oss/cubecos/pull/1281)** (#1276).

**On "upgrade OVN":** OVN 24.03 on Open vSwitch 3.3, the version-correct pairing RDO gives Caracal. #1281 did the work; this PR adds what #1276's DoD still wanted — the resolved versions **pinned**, `ovn24.03-24.03.8-31.el9s` and `openvswitch3.3-3.3.8-7.el9s` in `LOCKED_DNF`. That pin is not housekeeping: `core/neutron/Makefile` builds the DR-SNAT-patched `ovn-northd` from a *hardcoded* src rpm URL, so floating RPMs would eventually ship a northd from one OVN release beside an `ovn-controller` and `ovn-nb` schema from another, with nothing reporting the skew. The NB/SB schema conversion needs no CubeCOS code — `ovn-ctl` calls `ovs-lib`'s `upgrade_db`, which converted 7.0.0 → 7.3.0 and 20.27.0 → 20.33.0 in place with `ovn-nbctl show` / `ovn-sbctl show` byte-identical either side.

[...]

PR #1281, opened 2026-08-17
feat(neutron): upgrade OVN to 24.03 + OVS 3.3 for the Caracal hop

### LINKED ISSUE #1276: Upgrade OVN 23.03→24.03 + OVS 3.1→3.3 for the Caracal hop
#### What this issue does / why we need it

OVN has been pinned at **ovn23.03 / openvswitch3.1** (`core/neutron/neutron.mk`) since the initial cube import (2025-02) and never bumped. This story tracks the OVN upgrade + validation for the **Caracal hop** of the Yoga→Epoxy ladder, mirroring the qemu one-story-per-hop pattern (#1150–#1152).

Backed by a real rig spike (`/root/migration-rig/ovn-spike-results.md`): c9s NFV SIG carries 22.03–26.03; chassis-first skew is the supported/tested direction; NB/SB schema conversion is in-place and lossless across 23.03→24/25/26.03; the ovn-controller SEGV (see #1270) is **not** fixed by any of these versions.

Target = the max build compatible with the Caracal image's neutron at hop-build time (re-run the rig prune when the image is cut; 24.03→25.03 converts clean in the spike but certify on the real image).
#### Rolling-upgrade contract (mandatory)

- **Chassis first, central last**: upgrade ovn-host node-by-node (new-chassis/old-central is the tested skew); flip the pacemaker `ovndb_servers` (NB/SB + northd) last, with the ovsdb schema conversion as its own gated step.
- **Back up** ovnnb_db.db / ovnsb_db.db before the central flip — schema conversion is one-way.

#### Definition of Done

- [ ] Target OVN version resolved on the built Caracal image (`dnf --showduplicates`), pinned in neutron.mk.
- [ ] Rig prune: mixed-version control-plane matrix green (chassis registers, bindings kept, schema converts).
- [ ] **Cluster roll certification** (the authoritative sign-off, not rig-certifiable): real VMs keep networking through the node-by-node chassis roll + central flip; rollback rehearsed from a backed-up DB.
- [ ] set_ready_reconcile SEGV heal still present (bug persists at target version).

Spike + evidence: handbook `kb/cubecos/blueprints/nova-live-resize.md` and `/root/migration-rig/ovn-spike-results.md`.

      beside the note.
      Landed as [`kb/cubecos/known-issues/ovn-in-place-major-swap-hazards.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/ovn-in-place-major-swap-hazards.md).

traviswu-bigstack:
#### What type of PR is this?

/kind feature

Upgrades OVN to **24.03** (LTS) on **Open vSwitch 3.3** — the version-correct pairing for the **Caracal** hop of the Yoga→Epoxy ladder (RDO `rdo-openvswitch/3.3/1.el9s` → `rdo-ovn-24.03`). Antelope stays on the existing 23.03/3.1, which is already correct for it (see #1275).

Backed by the OVN upgrade spike (`/root/migration-rig/ovn-spike-results.md`): chassis-first mixed-version is the supported direction, and NB/SB schema conversion 23.03→24.03 is in-place and lossless.

**Changes:**
- `neutron.mk`: `openvswitch3.1`→`3.3`, `ovn23.03*`→`ovn24.03*`, `rdo-ovn` metapackages repointed to the `3.3/1.el9s` koji path.
- `neutron/Makefile`: custom `ovn-northd` built from `ovn24.03-24.03.8-31.el9s` src rpm; BuildRequires resolved via `dnf builddep` (24.03 adds `python3-scapy` etc. the old hardcoded list lacked), `rpmbuild -bc`, git-hash source dir globbed.
- `ovn_patch/centos9`: the downstream DR-SNAT patch **rebased** onto 24.03. 24.03's NAT-pipeline refactor merged the distributed/centralized `priority += 128` bumps into a single `if (n_l3dgw_ports)` across three flow builders — gated on `distributed_nat` to faithfully reproduce cube's 23.03 behavior (verified against the 23.03 source: both branches carried the bump; cube dropped it only on centralized SNAT). New `ovn24.03.spec`; `ovn23.03.spec` dropped; `ovn-northd` rebuilt (24.03.9, patch applied via `%autosetup`).

Fixes #1276

- The northd patch is **not** a plain deletion — see the commit body / the `distributed_nat` gate rationale. A naive `-priority += 128` would have wrongly stripped the bump from distributed (floating-IP) NAT flows.
- Rig-prune (control-plane: chassis registration, schema conversion) validated in the spike. **Cluster-roll certification** (real-VM datapath continuity through a chassis-first roll + the central flip + rollback) is the deferred DoD item, tracked on #1276.

#### Additional documentation

```docs
handbook /root/migration-rig/ovn-spike-results.md — OVN upgrade path (per-hop version pairing)
```


Eandalf-Bigstack:
### Review: verified on `jim-1cc` (cc1) as part of the neutron Caracal hop (#628)

I landed this branch on `centos9_cubecos_jim_200.13-1cc` ahead of the neutron hop so the OVN
change could be verified on its own, before neutron 24.2.2 went in. Summary: **the substance
holds up**, including the part that worried me most. One DoD item was missing and is now fixed
on `jim.lin/feat/openstack-c-upgrade-neutron`; everything else below is confirmation.

#### The `ovn-northd` binary is reproducible from the recipe in the tree

Ran the `neutron/Makefile` recipe by hand in the build jail — `dnf builddep` on the new spec,
`rpmbuild -bc`, glob the git-hash source dir — and the result is **byte-identical** to the
committed binary after `INSTALL_PROGRAM`'s strip:

```
53c6db53fccb68194be1e2d9990742a2  /root/rpmbuild/BUILD/ovn-697058.../northd/ovn-northd
53c6db53fccb68194be1e2d9990742a2  core/neutron/ovn_patch/centos9/ovn-northd
$ ovn-northd --version
ovn-northd 24.03.9
Open vSwitch Library 3.3.8
```

The `dnf builddep` switch is load-bearing, not a tidy-up: 24.03's spec pulls `python3-scapy`
and a texlive chain the old hardcoded `BuildRequires` list didn't have. And `-bc` replacing
`-ba || /bin/true` means the build now fails loudly instead of silently shipping a stale
binary. Both good changes. (The `24.03.9` against a `24.03.8-31` srpm is expected — RDO's
`ovn24.03.patch` rebases onto commit `697058648fde`, which is 24.03.9.)

#### The DR-SNAT rebase is faithful, and this was the risky part

I checked it against the real 23.03.0 and 24.03.8 sources rather than the patch alone, because
the commit message's claim ("both branches carried the bump; cube dropped it only on
centralized SNAT") is the whole basis for the `distributed_nat` gate.

Confirmed. In 23.03 `build_lrouter_out_snat_flow` had:

```c
if (od->n_l3dgw_ports) {
    if (distributed) { priority += 128; ... is_chassis_resident(nat->logical_port) }
    else             { priority += 128; ... is_chassis_resident(cr_port) }   /* cube removed THIS one */
}
```

24.03 split that one function into three — `build_lrouter_out_snat_stateless_flow`,
`build_lrouter_out_snat_in_czone_flow`, `build_lrouter_out_snat_flow` — moved the
`is_chassis_resident` match into the shared `build_lrouter_out_snat_match()`, and left a bare
`if (od->n_l3dgw_ports) priority += 128;` inline in each. `distributed_nat` is in scope in all
three and carries the same meaning as 23.03's `distributed`. Only one of the three runs for any
given NAT rule, so gating all three reproduces 23.03 exactly: bump kept on distributed
(floating-IP) NAT, dropped on centralized SNAT.

The fourth `priority += 128` in 24.03's `northd.c` is in `lrouter_nat_add_ext_ip_match` and is
correctly left alone. A naive `-priority += 128` really would have stripped the bump from
distributed NAT flows too — the commit message is right to call that out.

#### Schema conversion is in-place and lossless, and needs no cube code

`ovn-ctl` and `ovs-ctl` both call `ovs-lib`'s `upgrade_db`, which does
`ovsdb-tool needs-conversion` → backup → compact → `convert` on service start:

```
NB: 7.0.0    -> 7.3.0
SB: 20.27.0  -> 20.33.0
conf.db: 3.1 -> 3.3
```

`ovn-nbctl show` and `ovn-sbctl show` are **byte-identical** before and after; `ovs-vsctl show`
differs only in `ovs_version`. The cc1 chassis re-registered with its port bindings intact, and
pre-conversion backups are left next to the live files (`ovnnb_db.db.backup7.0.0-*`).

Worth knowing for #1276's backup step: on a *failed* convert, `upgrade_db` logs a warning and
**creates an empty database**. The backup that story mandates is guarding a real code path, not
a theoretical one.

#### Functional check, before neutron moved

With OVN 24.03.8 / OVS 3.3.8 under the *unchanged* neutron 22.2.1, `hex_cli -c cluster check`
came back all 27 groups ok with `Network ok [ neutron(v) ]`, all three agents alive, and
network / subnet / router / `router add subnet` create and delete round-tripped with the
logical router appearing in and disappearing from the OVN NB. So this branch is good on its own
merits, independent of #628.

#### One gap: #1276 asks for the version pinned, and the names went in unversioned

`ROOTFS_DNF += openvswitch3.3 ovn24.03 ovn24.03-central ovn24.03-host` floats. That matters more
than usual here, because `neutron/Makefile` builds `ovn-northd` from a *hardcoded*
`ovn24.03-24.03.8-31.el9s` srpm URL while the RPMs track newest — so the day the repo ships
24.03.10, the shipped northd is from an older release than the `ovn-controller` and NB schema
beside it, and nothing reports the skew.

Fixed in `8929da8` on `jim.lin/feat/openstack-c-upgrade-neutron`, pinning both to what the
Caracal image resolves to today, with `ROOTFS_DNF` referencing the same variables so they can't
drift apart:

```make
OVN_VER := -24.03.8-31.el9s
OVN_LOCKED_RPMS := ovn24.03$(OVN_VER) ovn24.03-central$(OVN_VER) ovn24.03-host$(OVN_VER)
OVS_LOCKED_RPMS := openvswitch3.3-3.3.8-7.el9s
LOCKED_DNF += $(OVN_LOCKED_RPMS) $(OVS_LOCKED_RPMS)
```

Neither package carries an Epoch, so the plain name-version form holds (an epoch in
`LOCKED_DNF` silently defeats the pin — see `core/heavyfs/Makefile`).

#### Two things I raised and then withdrew — recorded so they aren't re-litigated

**1. `ovn24.03` file-conflicts with `ovn23.03`.** It neither obsoletes nor conflicts with it,
yet both own `/usr/bin/ovn-nbctl`, `/usr/bin/ovn-northd`, `/usr/share/ovn/scripts/ovn-ctl` and
a dozen man pages, so an in-place install stops at:

```
Error: Transaction test error:
  file /usr/bin/ovn-appctl from install of ovn24.03-24.03.8-31.el9s.x86_64 conflicts with
  file from package ovn23.03-23.03.0-69.el9s.x86_64
```

`--allowerasing` does not help — there's no dependency reason to erase the old set, only a file
clash, so `dnf -y remove --noautoremove ovn23.03 ovn23.03-central ovn23.03-host` has to come
first (which also takes the `rdo-ovn*` metapackages with it; reinstall the 24.03 noarch ones
after). I initially proposed `BLKLST_DNF += ovn23.03 openvswitch3.1` to prune the stale rpms
`installdnf` leaves in `core/heavyfs/RPMS` — **not needed**, per Jim: a jail rebuild clears the
cache, and a clean build never has the 23.03 rpms to begin with. Noted here only for whoever
does the next in-place swap by hand.

**2. The chassis step blacking out the management plane.** On my surgical landing the
`openvswitch3.1` erase scriptlet stopped ovs, which took the `provider` bridge and cc1's mgmt IP
with it (recovered over the qemu serial console on `:11011`). I flagged this as a gap in
#1276's "chassis first, central last" contract — **withdrawn**: the product upgrades by flashing
the other A/B partition and booting it, so the node comes up already on 24.03/3.3 and
`hex_config` plus the `hex_sdk` bootstrap builds the provider bridge and mgmt IP from scratch.
There is no bridge to tear down. Only an in-place landing hits this.

#### Verdict

LGTM on the OVN change itself. The northd rebase is the part that could have been subtly wrong
and it isn't. The version pin is the one real gap and it's already fixed downstream, so nothing
here needs to change unless you'd rather carry the pin in this PR than in #628's stack.

*Reviewed with Claude Opus 5.*

Eandalf-Bigstack (approved):

PR #1282, opened 2026-08-17
fix(gpu): report vGPU capability for cards bound to vfio-pci

### LINKED ISSUE #1263: [Bug]: A pgpu card can never be switched to vGPU through the API — supportTypes goes stale under vfio-pci

SekiXu:
A `pgpu` card is bound to `vfio-pci` and is therefore not enumerable by
`nvidia-smi`. `gpu_device_list` builds its entry from `config.json` and a PCI
bus scan instead — and both of those paths **hardcoded**
`supportTypes:["pgpu"]`. `gpu_vgpu_profile_list` has the same blind spot: its
probe returns `No devices were found`, so it reported no profiles at all.

That is what the card currently *is*, not what it *can be*. `cube-cos-api`
gates the resource-type update on `supportTypes` and answers **409**, and
`gpu_resource_set` accepts no `unset` type, so a pgpu card could not be moved
to vGPU through the API at all — only by calling `hex_config` directly on the
node. The API documents the field as capability data
(`whatever mode it is in now`), which a probe-based answer cannot deliver.

## Approach

Read capability from the vGPU host driver's own static table,
`/usr/share/nvidia/vgpu/vgpuConfig.xml`. It is keyed by PCI device id, and a
card's vendor/device ids stay readable in sysfs whatever driver holds it — so
the answer stops depending on driver binding.

- **No cached state and no schema change.** An earlier sketch persisted
  capability into `config.json`, which would have needed a schema change, a
  second write point, boot-time backfill for already-carved cards, and a
  staleness story. None of that is needed.
- **No side effect on the card.** The fallback only runs where the probe
  already produced nothing; nothing is unbound or rebound.
- **Shell only** — no `hex_config` change, so nothing to compile.
- A node with no vGPU host driver has no such table, so the fallback finds
  nothing and the card reports pgpu-only — the correct answer there.

`vmCountLimit` stays `null` on that path: the table carries a per-GI limit
only, and the card-wide count depends on how the card is partitioned, which is
undefined while it is still a passthrough card. It returns on its own once the
card runs vGPU and `nvidia-smi` answers again.

## Verification

Deployed to **cn13** and exercised against `00000001:C8:00.0`, a `pgpu` bound
to `vfio-pci`:

| | before | after |
|---|---|---|
| `supportTypes` | `["pgpu"]` | `["pgpu","sriovVgpu","migBackedVgpu"]` |
| profiles | 0 sriov / 0 migBacked | **29 / 43** |

Those are the **same counts, ids and framebuffer sizes** the `nvidia-smi` path
reports for an identical card that is *not* bound to vfio-pci — `DC-12Q` id
`1528` `12288 MiB`, `DC-1-12Q` id `1558` `12288 MiB` — with `count`/`alias`
correctly `0`/`null` for a card that carries no profiles yet, versus
`1`/`nvidia_rtx_pro_6000_blackwell_dc_12q_1528` on the carved one.

`hex_sdk gpu_resource_set_check <uuid> sriovVgpu '[{"id":1528,"count":1}]'`
now exits `0` for that card. The card stayed bound to `vfio-pci` throughout,
and the three cards `nvidia-smi` can see are byte for byte unchanged.

## Note for the reviewer

The issue text attributed the stale value to the `nvidia-smi`-derived path.
That is true for the probe in the main loop, but the pgpu case never reaches
it — those two entries are built by the vfio-specific branches, where the
value was a literal. The fix is in those branches.

Fixes #1263

### DIFF
--- a/core/sdk_sh/modules/sdk_gpu.sh
+++ b/core/sdk_sh/modules/sdk_gpu.sh
@@ -10,6 +10,14 @@ GPU_CONFIG_FILE_PATH="/etc/cube/cos/gpu/config.json"
 SRIOV_PROFILE_NAME_REGEX="^[A-Za-z0-9]+-[0-9]+[A-Za-z]+$"
 MIG_PROFILE_NAME_REGEX="^[A-Za-z0-9]+-[0-9]+-[0-9]+[A-Za-z]+$"
 
+# Static vGPU type table shipped with the NVIDIA vGPU host driver, keyed by PCI
+# device id. It answers the same question `nvidia-smi vgpu -s -v` does, and is
+# needed because a pgpu is bound to vfio-pci and therefore invisible to
+# nvidia-smi while its PCI ids stay readable in sysfs. Absent when no vGPU host
+# driver is installed - in which case the card genuinely cannot run vGPU and
+# reporting pgpu-only is the correct answer.
+VGPU_CONFIG_XML="/usr/share/nvidia/vgpu/vgpuConfig.xml"
+
 gpu_iommu_list()
 {
     shopt -s nullglob
@@ -440,6 +448,81 @@ gpu_vm_stats()
 #     total: number
 #   } | null
 # }[]
+# Emits one TSV record per vGPU type the card's model supports:
+#   <typeId>\t<shortProfileName>\t<framebufferMiB>
+# Takes an nvidia-smi style bus id ("00000001:C8:00.0"). Silent (exit 0, no
+# output) when the table or the device is missing - callers treat that the same
+# way they treat an empty nvidia-smi probe.
+gpu_vgpu_types_from_xml()
+{
+    local pci_bus_id="$1"
+
+    [ -n "$pci_bus_id" ] || return 0
+    [ -r "$VGPU_CONFIG_XML" ] || return 0
+
+    # nvidia-smi prints an 8-digit domain in upper case; sysfs uses 4 digits in
+    # lower case. Same trimming as gpu_bind_vfio_pci.
+    local sysfs_pci_addr
+    sysfs_pci_addr=$(echo "$pci_bus_id" | sed 's/^[0-9a-fA-F]\{4\}//' | tr '[:upper:]' '[:lower:]')
+
+    local device_id
+    device_id=$(cat "/sys/bus/pci/devices/${sysfs_pci_addr}/device" 2>/dev/null)
+    [ -n "$device_id" ] || return 0
+
+    # sysfs prints 0x2bb5, the table spells it 0x2BB5
+    device_id="0x$(echo "${device_id#0x}" | tr '[:lower:]' '[:upper:]')"
+
+    awk -v want="$device_id" '
+        /<vgpuType / {
+            id = ""; name = ""; dev = ""; size = ""
+            if (match($0, /id="[0-9]+"/))
+                id = substr($0, RSTART + 4, RLENGTH - 5)
+            if (match($0, /name="[^"]*"/))
+                name = substr($0, RSTART + 6, RLENGTH - 7)
+        }
+        /<devId / {
+            if (match($0, /deviceId="[^"]*"/))
+                dev = substr($0, RSTART + 10, RLENGTH - 11)
+        }
+        /<profileSize>/ {
+            if (match($0, />0[xX][0-9a-fA-F]+</))
+                size = substr($0, RSTART + 1, RLENGTH - 2)
+        }
+        /<\/vgpuType>/ {
+            # The profile name callers match on is the last token of the full
+            # board name ("NVIDIA RTX Pro 6000 Blackwell DC-12Q" -> "DC-12Q"),
+            # the same slice nvidia-smi parsing takes with $NF.
+            if (dev == want && id != "" && name != "" && size != "") {
+                n = split(name, parts, " ")
+                printf "%s\t%s\t%d\n", id, parts[n], strtonum(size) / 1048576
+            }
+        }
+    ' "$VGPU_CONFIG_XML"
+}
+
+# supportTypes for a card nvidia-smi cannot enumerate. A pgpu is bound to
+# vfio-pci, so every probe-based answer collapses to ["pgpu"] - which is what
+# the card currently is, not what it can be. supportTypes is capability data,
+# so derive it from the static table instead.
+gpu_support_types_from_xml()
+{
+    local pci_bus_id="$1"
+
+    local support_types='["pgpu"]'
+    local profile_names
+    profile_names=$(gpu_vgpu_types_from_xml "$pci_bus_id" | awk -F'\t' '{print $2}')
+
+    if echo "$profile_names" | grep -qE "$SRIOV_PROFILE_NAME_REGEX"; then
+        support_types=$(echo "$support_types" | jq -c '. + ["sriovVgpu"]')
+    fi
+
+    if echo "$profile_names" | grep -qE "$MIG_PROFILE_NAME_REGEX"; then
+        support_types=$(echo "$support_types" | jq -c '. + ["migBackedVgpu"]')
+    fi
+
+    echo "$support_types"
+}
+
 gpu_device_list()
 {
     local gpu_config="[]"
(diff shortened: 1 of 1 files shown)


Eandalf-Bigstack:
The change set is small, and would be going into v3.1.10 to make GPU functionalities intact. But, I would prefer this PR being merged behind PR #1266 to avoid the huge rebase and conflict resolving works there.

SekiXu:
Sure, I'll mark this PR to draft and wait for PR#1266.

Eandalf-Bigstack (approved):

Issue #1303, opened 2026-08-20
[Bug] Rolling upgrade wedges forever: cached console list makes the boot path block on a getty-restarted serial console


traviswu-bigstack:
## Description

A rolling upgrade/restart can wedge forever, with the in-flight node stuck in `finalizing` and every remaining node stuck in `pending`, because the node's boot path blocks inside a **console write** that sits upstream of the roll relay.

`bootstrap_cube_config` resolves the console device list once at shell start (`KERNEL_CONSOLES=$(hex_sdk GetKernelConsoleDevices)`) and every `bootstrap_status` call then writes to that cached list through a single shared `tee`. The cached list goes stale within the same boot:

1. Early in boot, `/dev/ttyS0` is writable — the kernel applies the serial console's `CLOCAL` on the **first** open after boot — so `GetKernelConsoles`' 1-second probe passes and `/dev/ttyS0` enters the cached list.
2. Later in the same boot the console getty is restarted. The tty is torn down and recreated **without** `CLOCAL`, and the console's one-shot `CLOCAL` is already consumed.
3. From that point, with no carrier on the line, every `open("/dev/ttyS0")` blocks in `tty_port_block_til_ready` **forever**.
4. The next `bootstrap_status` blocks. `hex_sdk power_roll_advance "$HOSTNAME"` is downstream of it in the same shell, so the node never marks itself `done`, never clears `inflight`, and never kicks the next node. The roll is dead with no error anywhere.

Two aggravating factors:

- The shared `tee` opens `/dev/ttyS0` first (cmdline order), so the *healthy* console (`/dev/tty0`) never receives the line either — the operator sees a frozen console, not a diagnostic.
- The stall watchdog is not a safety net here: it does not exist in the image that orchestrates the roll (a pre-3.1.10 cluster), so a blown node deadline pauses nothing. `hex_sdk power_roll_status` prints `(stalled)`, but `power_roll_status_json` — what the API/UI reads — keeps reporting `state: running`, `isProcessing: true`, `90%` indefinitely.

[...]

Issue #1225, opened 2026-08-07
[Enhancement] Hidden repair opt-out marker for FTS/driver validation


traviswu-bigstack:
## Description

A hidden, persistent marker `/etc/appliance/state/cube_repair_optout` that makes `cluster_check_repair_async` (the final FTS `check_repair`) and the health `_auto_repair` loop **no-op**. It lets the driver — and QA — iterate on the **raw** post-set_ready / post-reboot cluster state, so real defects stay visible instead of being silently auto-repaired (this is how the git-init / manila / DB-OVN FTS NGs in #1195 were found).

The phone-home agent drops the marker (driver-controlled via the deploy's `optOutRepair` flag). Persistent — survives the OS-phase reboots + rolling-restart / powercycle iteration; cleared on reimage or a normal (non-opt-out) deploy. Default behavior is unchanged when the marker is absent.

**Scope:** cubecos **honors** the marker (this issue). The driver **delivers** it (cube-cos-driver — separate PR).

## Definition of Done (DoD)

- [x] `cluster_check_repair_async` and the health `_auto_repair` loop no-op when the marker is present
- [x] Marker persists across OS-phase reboots + rolling-restart / powercycle
- [x] Cleared on reimage / a normal (non-opt-out) deploy
- [x] Driver-side delivery PR linked (cube-cos-driver)
- [x] Handbook note

## Technical Notes

`core/main/proj_functions` (`cluster_check_repair_async`) + `core/sdk_sh/modules/sdk_health.sh` (`_auto_repair` gate). Marker under `/etc/appliance/state/` so it's per-slot and persists across reboots.

## Output artifacts (Definition of Done)

> Beyond code, docs, and config changes, completing this issue must also deliver:

- [x] **Handbook knowledge update** — land the durable, team-readable knowledge from this
      work into the bigstack-handbook cubecos kb (`kb/cubecos/…`) via
      `/bigstack-core:save-to-handbook` (Topic / Runbook / Known-issue / ADR as fits).
      Diagnosed or fixed a failure? Ship the scenario sidecar (`<note>.scenario.yaml`)
      beside the note.
      Landed as [`kb/cube-cos-driver/architecture/deploy-internals.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cube-cos-driver/architecture/deploy-internals.md).

traviswu-bigstack:
[...]

PR #1302, opened 2026-08-20
fix(cluster): twenty-four roll, boot and upgrade defects found validating 3.1.10-rc1

traviswu-bigstack:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it

Twenty-four defects found validating 3.1.10-rc1 on the sky 3-node lab — rolling restart, cold
powercycle, rolling upgrade, masakari host failures — plus four more from QA's 3c3p powercycle run.
They ride together because they were found as one cascade and nearly all sit on the shared
roll/boot/upgrade path.

**Roll and boot**

| # | Symptom | Fix | commit |
|---|---|---|---|
| 1 | Every roll left `Storage NG [ceph_mgr(5)]` for tens of minutes — the dashboard wedges on any mgr failover | hand the mgr off and verify the dashboard before rebooting that node | `c7ecb02` |
| 2 | A node whose console stopped draining blocked the boot indefinitely | bound every console write | `8328048` |
| 3 | Roll drained the next node before the rebooted one served, so VMs went `ERROR` and the roll paused | gate each drain on every node's neutron/nova answering directly, not via the VIP | `b2eca68` |
| 4 | `cluster poweroff` on a repair-opted-out cluster came back with every VM `SHUTOFF`, silently | the opt-out suppresses health repair, not VM restore | `a8aa168` |
| 5 | ceph commit burned 389 s inside one `ceph_fs_mds_init`, all 16 workers idle behind it | `timeout -k` the ceph CLI; skip MDS placement when the mgr is unavailable | `d04c273` |
| 6 | Whole bootstrap ran against paused storage after a cold poweroff — MDS stuck `up:replay` | lift only the client-I/O blocks once the mons answer | `e07ec05` |
| 7 | `systemctl restart nfs-ganesha` loops forever when grace-DB membership is missing | rejoin the grace DB first | `8b5923c` |

**Upgrade**

| # | Symptom | Fix | commit |
|---|---|---|---|
| 8 | Every upgraded node lost nova-compute — `/var/lib/nova/compute_id` lives in the root slot the upgrade boots away from | declare it for migration, as ceph already does for its keyring | `441257d` |
| 9 | Upgrade silently re-enabled auto-repair under an operator who had disabled it | migrate the opt-out markers | `df27cf4` |
| 10 | A deliberately hot-patched node reverted mid-boot via `_git_client_init` | hidden `cube_git_optout` marker | `54ac712` |

**Storage plumbing**

| # | Symptom | Fix | commit |
[...]

PR #1307, opened 2026-08-20
feat(git): opt-out flag so a node keeps its uncommitted rootfs changes

### LINKED ISSUE #1306: [Feature] Hidden git opt-out so a node keeps its uncommitted rootfs changes
## Description

A node's `/` is a git worktree of the image commit, with the bare repo on shared cephfs (`/mnt/cephfs/backup/cube.git`, remote `cube`). Two paths in `sdk_git.sh` throw away a node's uncommitted changes:

- `_git_client_init` — on the first boot after an upgrade it runs `git add -A` then `git stash` (then `git pull --rebase`). Because `add -A` stages untracked files too, the stash takes both modified tracked files *and* newly added ones, so the working tree snaps back to the image commit and untracked files disappear from the tree.
- `git_push` — after committing and pushing, it fans out `cmd "git stash ; git pull"`, doing the same to every peer.

Nothing is destroyed (it all lands in `stash@{0}`), but the effect is that any hot-patched system file silently reverts, mid-boot, with no warning. That is correct for production — the rootfs should match the image — and wrong for a lab or dev cluster where an engineer has deliberately patched a node and needs it to stay patched across the upgrade boot.

This adds a **hidden opt-out marker** mirroring the existing repair opt-out (`/etc/appliance/state/cube_repair_optout`, checked inline at `sdk_health.sh:135`, see #1225): when the marker is present, the node keeps its uncommitted changes and does not publish them.

Discovered while diagnosing #1303 — a hot-patched `/usr/sbin/bootstrap` was reverted at the end of the upgrade boot, with `/.git/index` mtime matching the file's to the second.

## Definition of Done (DoD)

- [ ] `/etc/appliance/state/cube_git_optout` gates the `add -A` / `stash` / `pull` trio in `_git_client_init`; `branch --track` still runs, so HEAD adopts the image commit and local edits simply read as ` M`
- [ ] `git_push` is a no-op with a warning on an opted-out node, so its local edits are never harvested into the shared repo
- [ ] `git_push`'s peer fan-out tests the marker **per node**, so an opted-out peer keeps its changes while the rest follow the new commit
- [ ] Unit test in `core/sdk_sh/tests/` covering both the gated and ungated paths of both functions; all existing `sdk_sh` tests still pass
- [ ] Passing a code review

## Technical Notes

- **Feature Owner:**
- **Frontend:** n/a
- **Backend:** n/a
- **Infra:** `core/sdk_sh/modules/sdk_git.sh`, `core/sdk_sh/tests/test_git_optout.sh`

- The marker lives under `/etc`, which the rootfs repo's own `.gitignore` excludes (`sdk_git.sh` `git_ignore_file`), so the flag cannot be swept away by the mechanism it disables.
- Dropping the `pull` on the opted-out path costs nothing: `branch --track $branch $project/$branch` already lands the local branch on the fetched tip. This also avoids `git pull --rebase` failing outright on a dirty tree.
- `/.gitignore` is now referenced through `$CUBE_GITIGNORE` (default unchanged) so the functions are testable off-node. Behaviour-preserving.
- **Scope limit:** the flag stops an existing tree being reverted to its own HEAD. It does not preserve local changes across a *new* image install — a fresh pkg legitimately replaces the rootfs.
- **Trade-off being bought deliberately:** an opted-out node diverges from the shared repo and stops following pushed commits. That is what a lab wants and what production does not; hence hidden marker rather than a documented tuning.

## UI/UX Mockups (Optional)

None

## QA Verification

|---|---|---|
| 1 | `bash core/sdk_sh/tests/test_git_optout.sh` | `PASS=16 FAIL=0`, `OK: git opt-out` |
| 2 | Default (no marker), after a rolling upgrade of a node: `git -C / status --short` | clean — pristine image tree, i.e. today's behaviour is unchanged |
| 3 | Before rolling a node: `touch /etc/appliance/state/cube_git_optout`, then append a comment line to a tracked file (e.g. `/usr/sbin/bootstrap`), then roll that node. After it returns: `grep -c '<your line>' /usr/sbin/bootstrap` and `git -C / status --short usr/sbin/bootstrap` | the edit is still present (`1`) and the path shows ` M`; `git -C / log -1` still reports the image commit |
| 4 | On an opted-out node: `hex_sdk git_push "qa probe"` then `git -C / log -1 --format=%s` | logs `git opt-out: not pushing <host>'s local changes`, exits 0, and the commit subject is unchanged (nothing committed or pushed) |
| 5 | Marker on node B only; on node A modify a tracked file and run `hex_sdk git_push "qa probe"`. Then on B: `git -C / status --short`; on C: `git -C / status --short` | A's commit is pushed; B still shows its own ` M` edits; C is clean at the new commit |
| 6 | Remove the marker on a node (`rm /etc/appliance/state/cube_git_optout`) and roll it again | the node reverts to pristine as before — the opt-out is fully reversible |

      beside the note.
  → `kb/cubecos/known-issues/rootfs-git-checkout-strips-setuid.md` + `.scenario.yaml`: bigstack-oss/bigstack-handbook#339

traviswu-bigstack:
#### What type of PR is this?

/kind feature

A node's `/` is a git worktree of the image commit, with the bare repo on shared cephfs (`/mnt/cephfs/backup/cube.git`, remote `cube`). Two paths in `sdk_git.sh` discard a node's uncommitted changes:

- `_git_client_init` — on the first boot after an upgrade: `git add -A` then `git stash` (then `git pull --rebase`). Since `add -A` stages untracked files too, the stash takes modified tracked files *and* newly added ones, so the working tree snaps back to the image commit and untracked files vanish from the tree.
- `git_push` — after commit+push it fans out `cmd "git stash ; git pull"`, doing the same on every peer.

For production that is right: the rootfs should match the image. For a lab or dev cluster it means a deliberately hot-patched node silently reverts mid-boot with no warning. Nothing is destroyed — it all lands in `stash@{0}` — but you only find that out afterwards, and by then the boot that needed the patch has already run without it.

This adds a hidden opt-out marker, `/etc/appliance/state/cube_git_optout`, mirroring the existing repair opt-out (`cube_repair_optout`, checked inline at `sdk_health.sh:135`, #1225):

- `_git_client_init` keeps `branch --track` (HEAD still adopts the image commit) and skips the `add -A` / `stash` / `pull` trio, logging a warning. Local edits then simply read as ` M` — the state a node lands in anyway when patched after its init.
- `git_push` returns early with a warning, so an opted-out node's edits are never harvested into the shared repo. Its peer fan-out tests the marker **per node**, so an opted-out peer keeps its changes while the rest follow the new commit.

Dropping the `pull` on the opted-out path costs nothing — `branch --track $branch $project/$branch` already lands the local branch on the fetched tip — and it avoids `git pull --rebase` failing outright on a dirty tree.

Found while diagnosing #1303: a hot-patched `/usr/sbin/bootstrap` was reverted at the end of the upgrade boot, with `/.git/index` mtime matching the file's to the second, and the patch plus its backup sitting in `stash@{0}`.

Fixes #1306

- **Test-first.** `core/sdk_sh/tests/test_git_optout.sh` follows the existing `sdk_sh` test convention (extract the functions with awk, mock the git/`Quiet`/`cmd` layers, no cluster or cephfs needed). It asserts both directions for both functions: without the marker the current destructive path is unchanged; with it, `branch --track` still runs while `add -A`/`stash`/`pull` do not, and `git_push` neither commits, pushes, nor tells peers to stash. It was watched failing 8/8 on the new behaviour before the gates existed, and the 8 baseline assertions passed at that point — so the test demonstrably exercises the change rather than the mocks. `PASS=16 FAIL=0` now, and the other five `sdk_sh` tests still exit 0.
- **`/.gitignore` behind `$CUBE_GITIGNORE`** (default unchanged) so `_git_client_init` can be exercised off-node — otherwise its `[ -e "/.gitignore" ]` guard makes the function untestable anywhere but a real node. Behaviour-preserving.
- **The marker survives the mechanism it disables:** it lives under `/etc`, which the rootfs repo's own `.gitignore` excludes (see `git_ignore_file`).
- **Scope limit:** this stops an existing tree being reverted to its own HEAD. It does not preserve local changes across a *new* image install — a fresh pkg legitimately replaces the rootfs.
- **Trade-off, deliberately:** an opted-out node diverges from the shared repo and stops following pushed commits. That is what a lab wants and what production does not, which is why this is a hidden marker rather than a documented tuning — same call as the repair opt-out.
- Not addressed here, worth its own look: `git_push` commits with `-a` and publishes cluster-wide, so on a non-opted-out node any local edit anywhere in the tracked rootfs rides along into the image history whenever someone pushes.

#### Additional documentation

```docs
kb/cubecos/known-issues/rootfs-git-reset-reverts-hot-patches.md (+ scenario sidecar) — to file for #1306/#1303
```

### DIFF
--- a/core/sdk_sh/modules/sdk_git.sh
+++ b/core/sdk_sh/modules/sdk_git.sh
@@ -6,9 +6,15 @@ if [ -z "$PROG" ] ; then
     exit 1
 fi
 
+# Paths named, not inlined: the tests point them somewhere writable.
+CUBE_GITIGNORE=${CUBE_GITIGNORE:-/.gitignore}
+# Hidden opt-out, like cube_repair_optout: keep this node's uncommitted changes
+# instead of stashing them away, and never publish them to the shared repo.
+CUBE_GIT_OPTOUT=${CUBE_GIT_OPTOUT:-/etc/appliance/state/cube_git_optout}
+
 git_ignore_file()
 {
-    cat >/.gitignore <<EOF
+    cat >$CUBE_GITIGNORE <<EOF
 *.pyc
 *.po
 *.mo
@@ -57,7 +63,7 @@ _git_server_init()
         Quiet -n git init -b $branch
         # local cephfs path, not ssh://VIP (stalls during set_ready reconfig). #1195
         Quiet -n $GIT remote add $project ${cube_git_dir}
-        if [ -e "/.gitignore" ] ; then
+        if [ -e "$CUBE_GITIGNORE" ] ; then
             if ! git log -1 >/dev/null 2>&1 ; then
                 Quiet -n git add -A
                 Quiet -n git commit -m "$(hex_cli -c firmware list | grep ACTIVE | awk '{print $2}')" -a
@@ -65,7 +71,7 @@ _git_server_init()
             Quiet -n git push --set-upstream $project $branch
             git -P branch -r | grep -q "${project}/${branch}"
         else
-            Error "failed to generate /.gitignore"
+            Error "failed to generate $CUBE_GITIGNORE"
         fi
         Quiet -n popd
         $GIT -P log || Error "failed to initialize GIT server"
@@ -120,19 +126,25 @@ _git_client_init()
     Quiet -n $GIT init -b $branch
     # local cephfs path, not ssh://VIP (stalls during set_ready reconfig). #1195
     Quiet -n $GIT remote add $project ${cube_git_dir}
-    if [ -e "/.gitignore" ] ; then
+    if [ -e "$CUBE_GITIGNORE" ] ; then
         if $GIT fetch ; then
             _git_suid_save
             Quiet -n $GIT branch --track $branch $project/$branch
-            Quiet -n $GIT add -A
-            Quiet -n $GIT stash
-            Quiet -n $GIT pull $project $branch --rebase
+            # --track already lands HEAD on the fetched tip, so the opt-out path
+            # needs no pull to stay current; local edits just read as modified.
+            if [ -f $CUBE_GIT_OPTOUT ] ; then
+                log_warning "git opt-out: keeping uncommitted changes on $HOSTNAME; / stays modified vs $project/$branch"
+            else
+                Quiet -n $GIT add -A
+                Quiet -n $GIT stash
+                Quiet -n $GIT pull $project $branch --rebase
+            fi
             _git_suid_restore
         else
             Error "git server (on VIP node) is not ready"
         fi
     else
-        Error "failed to generate /.gitignore"
+        Error "failed to generate $CUBE_GITIGNORE"
     fi
     Quiet -n popd
     Quiet -n $GIT -P log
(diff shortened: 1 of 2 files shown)


traviswu-bigstack:
Folded into #1302 (commit `61f076d`, unchanged content plus `core/sdk_sh/tests/test_git_optout.sh` passing 16/16). `Fixes #1306` now lives on #1302, so closing this in favour of that PR.

PR #1313, opened 2026-08-20
feat(cinder): upgrade cinder from openstack antelope to caracal

Eandalf-Bigstack:
#### What type of PR is this?

- feature

#### What this PR does / why we need it

- Upgrade OpenStack Cinder from Antelope (22.3.0) to Caracal (24.5.0)
  - Move the block storage service out of the shared Antelope venv `/opt/openstack-antelope` (python 3.10) and into `/opt/openstack-caracal` (python 3.11), the venv Skyline, Keystone and Glance already occupy
  - Repoint the `/usr/bin/cinder-*` symlinks at the Caracal venv, and drop `cinder-3`, which has been dangling since #1088
  - Carry the local `NfsDriver` patch forward: `antelope_patch/` becomes `caracal_patch/`, re-applied on top of 24.5.0's `nfs.py`
- Escalate privsep through the Caracal venv's own helper. Cinder is the first service to leave the venv `/usr/bin/privsep-helper` points at, so `cinder_sys_admin` and `privsep_osbrick` name the helper explicitly and `/etc/sudoers.d/cinder` authorises that path
- Migrate configs: regenerate `cinder.conf.sample` against Cinder 24.5.0
- Stop carrying `api-paste.ini`, `rootwrap.conf`, `resource_filters.json` and `volume.filters` in the tree — the Cinder wheel installs all four under the venv prefix, so install them from there instead (261 lines out)
- Ignore the new features Caracal introduces (the external storage driver options that make up most of 2024.1, the new `oslo.messaging` queue-manager options)

#### Which issue(s) this PR fixes

- Close #629

#### Special notes for your reviewer

> **This PR stacks on `jim.lin/feat/openstack-c-upgrade-glance` (#1310), which itself stacks on `jim.lin/feat/openstack-c-upgrade-keystone` (#1304).** The base here is `develop`, so until those merge this PR's commit list also shows the Keystone and Glance commits. Only the five below belong to this PR:
>
> - `feat(cinder): upgrade cinder from openstack antelope to caracal`
> - `fix(cinder): escalate privsep through the caracal venv helper`
> - `feat(cinder): migrate the sample configs to cinder 24.5.0`
> - `refactor(cinder): install the configs the wheels carry`
> - `docs(cinder): correct where cinder actually reaches privsep`
>
> Review it after #1310, or diff against `jim.lin/feat/openstack-c-upgrade-glance`.

[...]

Issue #1150, opened 2026-07-22
[User Story] Upgrade qemu-kvm from 9.1.0-24 to 10.1.0-22


traviswu-bigstack:
## Description

As a CubeCOS operator, I want qemu-kvm to be pinned at the maximum version that is live-migration-compatible with 3.1.0's qemu-kvm 9.1.0-24 so that the rolling upgrade toward the Antelope-based release can live-migrate every VM off each old node with no stream breakage.

Background: c9s has rebased qemu-kvm to 10.1.0 and dropped all 9.1.0 builds from the mirror (kojihub still archives them), and `QEMU_VER` in `core/heavyfs/Makefile` is currently **unpinned** — the image gets whatever c9s has on build day. The pin was originally introduced because 9.1.0-27/-29 broke live migration from 9.1.0-20 (#155), so compatibility must be certified, not assumed.

**Container-rig verdict (2026-07-23):** 9.1.0-24 → 10.1.0-22 live migration **PASSES** — precopy and postcopy+auto-converge (the nova hardening caps), guest memory state carried, on machine type `pc-q35-rhel9.6.0` (what 3.1.0's `q35` alias bakes into production VMs; 10.1.0-22.el9 still ships it) with virtio-scsi/net/balloon/rng and the real CUBE 3.1.0 kernel as guest. Reverse (10.1.0-22 → 9.1.0-24) also passes, giving abort headroom. Same-version controls pass.

## Acceptance Criteria

> A detailed list of conditions that must be met for the story to be accepted. These statements must focus on functional behavior and constraints rather than specific code-level implementation.

- [ ] Pin qemu-kvm to `17:10.1.0-22.el9` (`QEMU_VER` in `core/heavyfs/Makefile`) — never build unpinned again.
- [ ] Fetch the pinned rpms from a stable source (kojihub archive / vendored), not the rotating c9s mirror, so an air-gapped rebuild resolves the exact certified NVR (same pattern as the mariadb re-pin).
- [ ] New VMs boot and existing VMs run on 10.1.0-22 (q35 alias now resolves to pc-q35-rhel9.8.0 for new VMs — existing VMs keep their baked pc-q35-rhel9.6.0).

### Rolling-upgrade / mixed-version DoD

Must be safe to apply **one node at a time on a live 3-node HA cluster** — every drain live-migrates VMs from a 9.1.0-24 (old) host to a 10.1.0-22 (new) host. Verify on a real HA cluster:

[...]

PR #1124, opened 2026-07-17
feat(nova): upgrade nova from yoga to antelope

Eandalf-Bigstack:
#### What type of PR is this?

- feature

#### What this PR does / why we need it

- Upgrade Nova from OpenStack Yoga to Antelope

#### Which issue(s) this PR fixes

- Close #602 

#### Additional documentation

For the requirement "health check",

```bash
[cc1 ~]# hex_cli -v -c boot_mode status
two phase boot... [yes]
standalone services... [ready]
cube services... [ready]
cluster data... [synced]
applying policy... [no]
[cc1 ~]# hex_cli -v -c cluster check
          Service  Status  Report
           IaasDb      ok  [ mysql(v) ]
            Image      ok  [ glance(v) ]
       InstanceHa      ok  [ masakari(v) ]
            LBaaS      ok  [ octavia(v) ]
        HaCluster      ok  [ hacluster(v) ]
        Baremetal      ok  [ ironic(v) ]
       ClusterSys      ok  [ bootstrap(v) license(v) ]
       ObjectStor      ok  [ swift(v) ]
           K8SaaS      ok  [ rancher(v) ]
         FileStor      ok  [ manila(v) ]
     SingleSignOn      ok  [ k3s(v) keycloak(v) ]
    Orchestration      ok  [ heat(v) ]
          Metrics      ok  [ monasca(v) telegraf(v) grafana(v) ]
      ClusterLink      ok  [ link(v) clock(v) dns(v) ]
        BlockStor      ok  [ cinder(v) ]
           DNSaaS      ok  [ designate(v) ]
    BusinessLogic      ok  [ watcher(v) ]
         DataPipe      ok  [ zookeeper(v) kafka(v) ]
          Compute      ok  [ nova(v) cyborg(v) ]
        VirtualIp      ok  [ vip(v) haproxy_ha(v) ]
  ClusterSettings      ok  [ etcd(v) nodelist(v) mongodb(v) ]
    Notifications      ok  [ influxdb(v) kapacitor(v) ]
     LogAnalytics      ok  [ filebeat(v) auditbeat(v) logstash(v) opensearch(v) opensearch-dashboards(v) ]
       ApiService      ok  [ haproxy(v) httpd(v) nginx(v) api(v) skyline(v) memcache(v) ]
          Network      ok  [ neutron(v) ]
[...]

PR #1314, opened 2026-08-20
feat(nova): upgrade nova and placement from openstack antelope to caracal

### LINKED ISSUE #627: [Story] Upgrade Nova from Antelope to Caracal
## Description

As a CubeCOS operator, I want Nova be in Caracal so that we could upgrade to Caracal.

https://docs.openstack.org/releasenotes/nova/2024.1.html

> Instances using vGPUs can now be correctly live-migrated by the libvirt driver between compute nodes supporting the same mediated device types used by the instance. In order to be able to do this, the compute hosts need to support at least the minimum versions of libvirt-8.6.0, QEMU-8.1.0 and Linux kernel 5.18.0. If operators use multiple vGPU types per compute, they need to make sure they already use custom traits or custom resource classes for the GPUs resource providers and that the instance was created with a flavor using either a custom resource class or asking for a custom trait in order to make sure that Placement API will provide the right target GPU using the same mdev type for the instance.

## Acceptance Criteria

- [ ] Migrate configs
  - [ ] Ensure VM creations and deletions work, we would test for VMs with the ephemeral storage (without root disk volume type specification) and without any network interfaces specified. Only VM interactions via the noVNC console are expected. Network connectivities are not expected.
- [ ] Adjust the patch for SR-IOV based vGPU in v3.2.0 to match patching positions in OpenStack Caracal

## Story Points: 3

> | -- | -- | -- |

Eandalf-Bigstack:
#### What type of PR is this?

- feature

- Upgrade OpenStack Nova from Antelope (27.5.1) to Caracal (29.4.0), and Placement from 9.0.0 to 11.0.1
  - Move both out of the shared Antelope venv `/opt/openstack-antelope` (python 3.10) and into `/opt/openstack-caracal` (python 3.11), the venv Skyline, Keystone, Glance and Cinder already occupy
  - Repoint the 17 `/usr/bin/nova-*`, 3 `/usr/bin/placement-*` and `/usr/bin/uwsgi` symlinks at the Caracal venv
  - Bump `libvirt-python` in `os-caracal-pip-upper-constraints.txt` from 10.0.0 to 11.10.0 to match the pinned `libvirt-devel-11.10.0-14.el9`
  - Carry the SR-IOV/vGPU PCI patch forward: `antelope_patch/` becomes `caracal_patch/`, re-applied on top of 29.4.0
- Escalate privsep through the Caracal venv's own helper — four sections, including the two `os-vif` ones that every instance with a NIC goes through
- **Drop `AvailabilityZoneFilter`, removed in Nova 29.0.0.** Naming a filter Nova no longer registers is fatal, not ignored: `nova-scheduler` exits on start
- **Raise the Placement uWSGI request buffer.** Caracal's traits query outgrew uWSGI's 4096-byte default, which broke every VM build
- Put five stray config options where they are actually read, three of which have never been read at all
- Migrate configs: regenerate `nova.conf.sample`, `nova-compute.conf.sample`, `placement.conf.sample` and `policy.yaml.sample` against 29.4.0 / 11.0.1
- Stop carrying `api-paste.ini`, `rootwrap.conf` and `rootwrap_compute.filters` — the Nova wheel installs all three under the venv prefix (143 lines out)
- Drop `placement-api.conf` and `nova-compute.conf.sample`, which nothing reads (5599 lines out)
- Move the qemu version lock from 10.1.0-22 to 10.1.0-23, with the live-migration compatibility rule the earlier breakage never got written down
- Ignore the new features Caracal introduces

- Close #627

> **This PR stacks on `jim.lin/feat/openstack-c-upgrade-cinder` (#1313), which stacks on `jim.lin/feat/openstack-c-upgrade-glance` (#1310), which stacks on `jim.lin/feat/openstack-c-upgrade-keystone` (#1304).** The base here is `develop`, so until those merge this PR's commit list also shows the Keystone, Glance and Cinder commits. Only the nine below belong to this PR:
>
> - `feat(nova): upgrade nova and placement from openstack antelope to caracal`
> - `fix(nova): escalate privsep through the caracal venv helper`
> - `fix(nova): drop AvailabilityZoneFilter, removed in nova 29.0.0`
> - `fix(nova): put the five stray config options where they are read`
> - `fix(placement): raise the uwsgi request buffer for caracal's traits query`
> - `feat(nova): migrate the sample configs to nova 29.4.0 and placement 11.0.1`
> - `refactor(nova): install the configs the nova wheel carries`
> - `refactor(nova): drop the two configs nothing reads`
> - `build(qemu): move the version lock to 10.1.0-23.el9`
>
> Review it after #1313, or diff against `jim.lin/feat/openstack-c-upgrade-cinder`.

**Why a second venv instead of a version bump in place.** Nova 29.4.0 resolves `os-vif` 3.5.0 and `oslo.privsep` 3.3.0, and pulls `numpy`, `websockify`, `ovs`, `ovsdbapp`, `pyroute2` and `redis` along with them. Neutron, Manila, Masakari, Cyborg and Ironic share `/opt/openstack-antelope`, so installing Nova 29 there would have lifted `os-vif` underneath Neutron's own ML2 drivers. Resolved against `os-caracal-pip-upper-constraints.txt` the two sets turn out to be disjoint — Nova and Placement add **17 packages** to the Caracal venv and change **no** version Skyline, Keystone, Glance or Cinder already holds. The four `python-*client` packages are already there at the Caracal constraints, so naming them is a no-op.

Nothing outside `core/nova` has to follow it. The ten service units, `config_nova.cpp`, `placement-uwsgi.ini` and `hex_sdk` all reach Nova and Placement through `/usr/bin/nova-*` and `/usr/bin/placement-*`, which simply point into the other venv now. `/usr/bin/uwsgi` is Placement's alone — no other unit uses it — and it *has* to move, because uWSGI embeds its own interpreter and only the Caracal build can load a python 3.11 venv through `home`.

**`libvirt-python` is the one C extension, and it needs both sides pinned.** It compiles against whatever `libvirt-devel` headers are present, so the Caracal constraints file gets the same deliberate 11.10.0 bump the Antelope file got in #1124. Left at Caracal's own 10.0.0 the build dies deep in `generator.py` on missing type converters, which reads like a code bug rather than a version mismatch.

**Privsep, second time around.** `/usr/bin/privsep-helper` stays pointed at the Antelope venv — Neutron, Manila, Masakari and Cyborg still reach their helper through it, and `config_manila.cpp`, `config_masakari.cpp` and `config_cyborg.cpp` name that path — so `nova.mk` now installs `oslo.privsep` into **both** venvs and Nova names its own helper, the shape #1313 established for Cinder. Four sections rather than Cinder's two: `nova_sys_admin`, `privsep_osbrick` (os-brick 6.7.3's name, not the `os_brick_privileged` the context name suggests), and os-vif 3.5.0's `vif_plug_ovs_privileged` and `vif_plug_linux_bridge_privileged`. The os-vif pair matters most in practice — `vif_plug_ovs` is what plugs the tap into `br-int`, so it is on the live path of every instance with a network interface. The pins go in the `IsControl || IsCompute` branch, because a pure compute node in `3c_3p_5s` runs `nova-compute` without being a controller and is exactly the node with nothing but privsep between it and a failed build.

**Two things broke that the issue did not predict, and both are silent about their own cause.**

`AvailabilityZoneFilter` was removed in Nova 29.0.0 (deprecated since 24.0.0 Xena). `HostManager._choose_host_filters` raises `SchedulerHostFilterNotFound` for a filter it cannot load, so `nova-scheduler` crash-looped on the first boot after the hop and every build ended in `Exhausted all hosts available for retrying build failures`. Nothing replaces it in `enabled_filters`: the `map_az_to_placement_aggregate` pre-filter has done the job since Rocky and is now mandatory, which is also why `[scheduler] query_placement_for_availability_zone` was removed outright rather than defaulted. One operational consequence: that pre-filter matches a Nova aggregate carrying an `availability_zone` to the Placement aggregate of the same UUID, so an AZ only constrains scheduling once `nova-manage placement sync_aggregates` has paired them. No aggregate here carries one today, so no behaviour change — but a deployment that starts using AZs will need that command where it previously needed nothing.

Placement's uWSGI then hit a threshold this release crossed. Nova's resource tracker issues one `GET /traits?name=in:<every trait the compute reports>` per sync cycle, and os-traits 3.0.0 plus Nova 29.4.0 push that past uWSGI's default 4096-byte request buffer once the whole block is counted — the URL alone is ~2.7 KB and Apache forwards 40 CGI vars on top, **measured at 4130 bytes**. uWSGI answers an oversized block by logging `unable to add HTTP_X_FORWARDED_SERVER=cc1 to uwsgi packet, consider increasing buffer size` and dropping the connection, with no error of its own. Apache turns that into `502 Proxy Error`, `nova-compute` dies with `TraitRetrievalFailed` on *every* cycle, `compute_nodes.updated_at` freezes, and the scheduler reports `Exhausted all hosts` for a host that is perfectly healthy while Placement answers every other request 200. `buffer-size = 65535` is uWSGI's ceiling, the packet length field being 16-bit.

**The vGPU patch ported cleanly, unlike last time.** #1124 had to re-implement it by hand against Antelope. Here all **eight** files applied onto 29.4.0 with offsets only and no rejects, and the diff each pair carries is byte-identical to the one it carried against 27.5.1. None of it landed upstream in 2024.1 — `device_assignable` still returns `bool`, `_set_managed_mode` still takes no `managed` argument, and `conf/pci.py` still documents only `remote_managed`. The `.orig` files are byte-exact Nova 29.4.0 so the next hop diffs against the truth. One patch is dormant rather than dead: `virt/block_device.py` guards the four `BlockDeviceMapping` 1.21 encryption fields, and Caracal keeps that object at 1.21, so `obj_make_compatible` pops nothing — kept because it costs nothing and the mixed-version window it covers is not closed.

**Five stray config options, three of them never read at all.** Found by walking every key in the live `/etc/nova/nova.conf` and `/etc/placement/placement.conf` against the option registry of the installed packages, which `oslo.config` will not do for you — an unknown key in a known group is silently ignored.

| Option | Verdict | Action |
| --- | --- | --- |
| `[placement] randomize_allocation_candidates` | a real *Placement* option, written into `nova.conf`, whose `[placement]` is only the client-side adapter group | moved into `placement.conf` |
| `[placement] os_region_name` | wrong name; the adapter group registers `region_name`, as `cfg["neutron"]["region_name"]` already spells it | renamed |
| `[neutron] url` | deprecated in 17.0.0 Queens, **removed** in 18.0.0 Rocky | deleted |
| `[libvirt] ovmf_path` | Nova has **never** registered this option | deleted |
| `[DEFAULT] vnc_enabled` | pre-Mitaka deprecated alias of `[vnc] enabled`, set two lines below | deleted |

Deleting `[neutron] url` rather than renaming it to `endpoint_override` is what Nova's own removal note prescribes — *"This should not need to be specified if an appropriate service catalog entry exists for the network service"* — and one does, resolving to the same `http://<shared>:9696` the dead line held. That note also warns that dropping `url` stops `region_name` defaulting to `RegionOne`, which is already covered explicitly.

`[libvirt] ovmf_path` was worse than inert. Nova 29.4.0 registers no ovmf, firmware, loader, nvram or uefi option in `[libvirt]` at all — only the three `swtpm_*` ones, of which `swtpm_enabled` on the line above is genuine — and `ovmf` appears nowhere in `nova/conf` or in any release note. It reads like a copy of the `<loader>` value out of a libvirt capabilities document, the only place the string occurs in Nova at all, in test fixtures. And it named `OVMF_CODE.secboot.fd` *unconditionally*, so had it ever been read it would have forced a secure-boot firmware onto guests that did not ask for one. UEFI is libvirt's job, verified below.

**Two dead installs go with the hop.** `placement-api.conf` is the mod_wsgi vhost from before #1124; that PR added an `rm -f` of its install target immediately before installing `placement-api-wsgi.conf.in` over the top, so it has been installed and deleted in the same build ever since. `nova-compute.conf` is 184 KB of entirely commented-out sample installed to `/etc/nova/nova-compute.conf` that nothing opens — all four Nova units run `--config-file /etc/nova/nova.conf --config-dir /etc/nova/nova.d`, and `nova.mk` is the only file in the tree that names it. It is an artifact of the RDO `openstack-nova-compute` subpackage; the pip packaging has no such split.

**The config delta reaching a node is structural this time, unlike Cinder's.** Nova's `.def` *is* the sample (`nova.mk` copies `/etc/nova/nova.conf` to `nova.conf.def`), and `hex_config`'s `LoadConfig` takes the section names from it, so the sample's section list is the generated config's section list. `[hyperv]` and `[rdp]` are gone with Nova 29.0.0's removal of the Hyper-V driver and the RDP console; `[oslo_versionedobjects]` arrives with the namespace upstream added to its own generator config; `[profiler_jaeger]` and `[profiler_otlp]` arrive with osprofiler 4.1.0. Two dozen options leave with the two removed sections, alongside three genuine removals — `use_forwarded_for`, `mysql_enable_ndb` and `query_placement_for_availability_zone`. None of them is written by `config_nova.cpp`.

**qemu: there is no version 11 to move to.** The CentOS Stream koji carries `qemu-kvm` 10.0.0 and 10.1.0 and nothing above, for el9 and el10 alike, all at epoch 17. So the only available move is 10.1.0-22 → -23, which brings three fixes on the live-migration path a rolling upgrade depends on: RHEL-158244 (live migration fails during WSFC testing, via scsi reservations carried across migration), RHEL-190701 (qemu-kvm hangs during drain after a double pause) and RHEL-45628 (qemu-kvm crashes when `virtproxyd` is killed on the destination mid-migration).

This pin carries a "never unpin" warning from #1150, earned when 9.1.0-27 turned out not to be live-migration-compatible with the 9.1.0-20 v3.0.0 shipped — observed, never explained. The cause is now recorded: **9.1.0-21 carried `kvm-hw-i386-Fix-machine-type-compatibility.patch` [RHEL-91307], "Fix x86 M-type compats"**, which rewrites the x86 machine-type compat property tables, so the same `pc-q35-rhel9.N.0` name describes a different guest on either side. That yields a repeatable test for moving this pin, which the comment now spells out: safe when `dnf repoquery --changelogs` over the delta shows no `hw-i386` machine-type/compat patch and `qemu-kvm -M help` still resolves `q35` to the same `pc-q35-rhel9.N.0`. Both hold for -22 → -23.

**New Caracal features are deliberately ignored,** per the issue's fourth criterion. The vGPU live-migration support 2024.1 advertises needs libvirt 8.6.0, QEMU 8.1.0 and kernel 5.18.0, all of which this appliance already exceeds — but it also needs custom traits or resource classes per mdev type, which `config_gpu.cpp` does not set up, so it is available rather than enabled. `hypervisor_version_weight_multiplier`, `num_instances_weight_multiplier`, `migration_inbound_addr`, `conductor_group`, `shard`, `sharing_providers_max_uuids_per_request`, `enforce_fips_mode`, `[libvirt] tb_cache_size`, `allocation_candidates_generation_strategy` and `max_allocation_candidates` are all left at their defaults.

#### Additional documentation

Verified on `jim-1cc` (`cc1`), except the qemu section, verified on `accept-3cc`.

For the requirement "Migrate configs",

```bash
[cc1 ~]# ls -l /usr/bin/nova-compute /usr/bin/nova-api /usr/bin/nova-manage /usr/bin/placement-api /usr/bin/uwsgi
lrwxrwxrwx 1 root root 35 Aug 20 20:13 /usr/bin/nova-api -> /opt/openstack-caracal/bin/nova-api
lrwxrwxrwx 1 root root 39 Aug 20 20:13 /usr/bin/nova-compute -> /opt/openstack-caracal/bin/nova-compute
lrwxrwxrwx 1 root root 38 Aug 20 20:13 /usr/bin/nova-manage -> /opt/openstack-caracal/bin/nova-manage
lrwxrwxrwx 1 root root 40 Aug 20 20:13 /usr/bin/placement-api -> /opt/openstack-caracal/bin/placement-api
lrwxrwxrwx 1 root root 32 Aug 20 20:13 /usr/bin/uwsgi -> /opt/openstack-caracal/bin/uwsgi
[cc1 ~]# nova-manage --version; placement-manage --version
29.4.0
11.0.1
[cc1 ~]# for u in api compute conductor scheduler novncproxy; do echo -n "openstack-nova-$u: "; readlink -f /proc/$(systemctl show -p MainPID --value openstack-nova-$u)/exe; done
openstack-nova-api: /usr/local/bin/python3.11
openstack-nova-compute: /usr/local/bin/python3.11
openstack-nova-conductor: /usr/local/bin/python3.11
openstack-nova-scheduler: /usr/local/bin/python3.11
openstack-nova-novncproxy: /usr/local/bin/python3.11
[cc1 ~]# echo -n "openstack-placement-api: "; readlink -f /proc/$(systemctl show -p MainPID --value openstack-placement-api)/exe
openstack-placement-api: /opt/openstack-caracal/bin/uwsgi
[cc1 ~]# systemctl is-active openstack-nova-api openstack-nova-compute openstack-nova-conductor openstack-nova-scheduler openstack-nova-novncproxy openstack-placement-api
active
active
active
active
active
active
[cc1 ~]# curl -s http://10.1.0.1:8774/ | python3 -c "import json,sys; d=json.load(sys.stdin); v=[x for x in d['versions'] if x['id']=='v2.1'][0]; print(v['id'], v['status'], 'max', v['version'], 'min', v['min_version'])"
v2.1 CURRENT max 2.96 min 2.1
[cc1 ~]# curl -s http://10.1.0.1:8778/ | python3 -c "import json,sys; v=json.load(sys.stdin)['versions'][0]; print(v['id'], v['status'], 'max', v['max_version'], 'min', v['min_version'])"
v1.0 CURRENT max 1.39 min 1.0
```

The three DBs reaching Caracal's heads, and the upgrade checks,

```bash
[cc1 ~]# mysql -N -B -e "select 'nova_api', version_num from nova_api.alembic_version union all select 'nova', version_num from nova.alembic_version union all select 'placement', version_num from placement.alembic_version"
nova_api	cdeec0c85668
nova	13863f4e1612
placement	a082b8bb98d0
[cc1 ~]# su -s /bin/sh -c "nova-status upgrade check" nova 2>/dev/null | grep -E "Check:|Result:"
| Check: Cells v2                           |
| Result: Success                           |
| Check: Placement API                      |
| Result: Success                           |
| Check: Cinder API                         |
| Result: Success                           |
| Check: Policy File JSON to YAML Migration |
| Result: Success                           |
| Check: Older than N-1 computes            |
| Result: Success                           |
| Check: hw_machine_type unset              |
| Result: Success                           |
| Check: Service User Token Configuration   |
| Result: Success                           |
[cc1 ~]# su -s /bin/sh -c "placement-status upgrade check" placement 2>/dev/null | grep -E "Check:|Result:"
| Check: Missing Root Provider IDs          |
| Result: Success                           |
| Check: Incomplete Consumers               |
| Result: Success                           |
| Check: Policy File JSON to YAML Migration |
| Result: Success                           |
```

The three files that now come from the wheel instead of the tree, and the two that are gone,

```bash
[cc1 ~]# find /opt/openstack-caracal/etc/nova -type f | sort
/opt/openstack-caracal/etc/nova/api-paste.ini
/opt/openstack-caracal/etc/nova/rootwrap.conf
/opt/openstack-caracal/etc/nova/rootwrap.d/compute.filters
[cc1 ~]# diff -q /opt/openstack-caracal/etc/nova/api-paste.ini /etc/nova/api-paste.ini && diff -q /opt/openstack-caracal/etc/nova/rootwrap.conf /etc/nova/rootwrap.conf && diff -q /opt/openstack-caracal/etc/nova/rootwrap.d/compute.filters /usr/share/nova/rootwrap/compute.filters && echo "all three identical to the wheel"
all three identical to the wheel
[cc1 ~]# ls /etc/httpd/conf.d/ | grep -i place; ls /etc/nova/nova-compute.conf
placement-api-wsgi.conf
placement-api-wsgi.conf.in
ls: cannot access '/etc/nova/nova-compute.conf': No such file or directory
```

The privsep bridge — `/usr/bin/privsep-helper` still points at the Antelope venv for everybody else, while Nova's contexts run out of the Caracal one, proven on live paths rather than by inspection,

```bash
[cc1 ~]# readlink -f /usr/bin/privsep-helper
/opt/openstack-antelope/bin/privsep-helper
[cc1 ~]# grep -A1 -E "^\[(nova_sys_admin|privsep_osbrick|vif_plug_ovs_privileged|vif_plug_linux_bridge_privileged)\]" /etc/nova/nova.conf
[nova_sys_admin]
helper_command = sudo /opt/openstack-caracal/bin/privsep-helper
--
[privsep_osbrick]
helper_command = sudo /opt/openstack-caracal/bin/privsep-helper
--
[vif_plug_linux_bridge_privileged]
helper_command = sudo /opt/openstack-caracal/bin/privsep-helper
--
[vif_plug_ovs_privileged]
helper_command = sudo /opt/openstack-caracal/bin/privsep-helper
[cc1 ~]# grep -v '^#' /etc/sudoers.d/nova
Defaults:nova !requiretty

nova ALL = (root) NOPASSWD: /usr/bin/nova-rootwrap /etc/nova/rootwrap.conf *
nova ALL = (root) NOPASSWD: /opt/openstack-caracal/bin/privsep-helper *
[cc1 ~]# grep 'Running privsep helper' /var/log/nova/nova-compute.log | sort -u
2026-08-20 20:31:26.393 INFO oslo.privsep.daemon Running privsep helper: ['sudo', '/opt/openstack-caracal/bin/privsep-helper', '--privsep_context', 'os_brick.privileged.default', '--privsep_sock_path', '/tmp/tmpu5xngz4l/privsep.sock']
2026-08-20 20:32:36.276 INFO oslo.privsep.daemon Running privsep helper: ['sudo', '/opt/openstack-caracal/bin/privsep-helper', '--privsep_context', 'vif_plug_ovs.privsep.vif_plug', '--privsep_sock_path', '/tmp/tmptnaw4mtj/privsep.sock']
2026-08-20 20:31:26.880 INFO oslo.privsep.daemon privsep process running with capabilities (eff/prm/inh): CAP_SYS_ADMIN/CAP_SYS_ADMIN/none
```

The uWSGI buffer — the same request, before and after, 34 bytes over the old limit,

```bash
[cc1 ~]# grep -E "buffer-size|^home" /etc/placement/placement-uwsgi.ini
home = /opt/openstack-caracal
buffer-size = 65535

# before: uWSGI drops the connection and Apache answers 502
[WARNING] unable to add HTTP_X_FORWARDED_SERVER=cc1 to uwsgi packet, consider increasing buffer size
ERROR nova.compute.manager nova.exception.TraitRetrievalFailed: Failed to retrieve traits from the placement API: <title>502 Proxy Error</title>
ERROR nova.compute.manager nova.exception.ResourceProviderSyncFailed: Failed to synchronize the placement service with resource provider information supplied by the compute host.
# 65 such ERRORs in two minutes, inventory frozen, every build ending in
# "Exhausted all hosts available for retrying build failures"

# after
[pid: 184022|app: 0|req: 5/10] 0.0.0.0 () {40 vars in 4130 bytes} GET /traits?name=in:COMPUTE_GRAPHICS_MODEL_BOCHS,... => generated 1878 bytes in 31 msecs (HTTP/1.1 200)
[cc1 ~]# grep -cE "ResourceProviderSyncFailed|TraitRetrievalFailed" /var/log/nova/nova-compute.log
0
```

For the requirement "Ensure VM creations and deletions work" — ephemeral storage, no root disk volume type, no network interface, noVNC only,

```bash
[cc1 ~]# openstack server create --os-compute-api-version 2.37 --flavor 8b93e729-a3d3-46e3-8391-50781734d12b --image 35f3f7ce-991e-46c7-ba5e-de1bfdb4fc32 --nic none test-ephemeral-vm -c id -c status -c flavor -c image
+--------+-------------------------------------------------+
| Field  | Value                                           |
+--------+-------------------------------------------------+
| flavor | t2.small (8b93e729-a3d3-46e3-8391-50781734d12b) |
| id     | 56b401d9-4e46-4544-be6e-c388ba44095d            |
| image  | Cirros (35f3f7ce-991e-46c7-ba5e-de1bfdb4fc32)   |
| status | BUILD                                           |
+--------+-------------------------------------------------+
[cc1 ~]# openstack server show test-ephemeral-vm -c status -c OS-EXT-STS:vm_state -c OS-EXT-STS:power_state -c OS-EXT-SRV-ATTR:hypervisor_hostname -c OS-EXT-SRV-ATTR:instance_name
+-------------------------------------+-------------------+
| Field                               | Value             |
+-------------------------------------+-------------------+
| OS-EXT-SRV-ATTR:hypervisor_hostname | cc1               |
| OS-EXT-SRV-ATTR:instance_name       | instance-0000000e |
| OS-EXT-STS:power_state              | Running           |
| OS-EXT-STS:vm_state                 | active            |
| status                              | ACTIVE            |
+-------------------------------------+-------------------+
[cc1 ~]# virsh list --all
 Id   Name                State
-----------------------------------
 13   instance-0000000e   running
```

The noVNC console — page, assets, and the websocket reaching the guest's VNC server,

```bash
[cc1 ~]# openstack console url show test-ephemeral-vm --novnc
+----------+------------------------------------------------------------------------------------------+
| Field    | Value                                                                                    |
+----------+------------------------------------------------------------------------------------------+
| protocol | vnc                                                                                      |
| type     | novnc                                                                                    |
| url      | https://10.1.0.1:6080/vnc_auto.html?path=%3Ftoken%3D08124860-fb67-4b66-a77f-651db485f33e |
+----------+------------------------------------------------------------------------------------------+
[cc1 ~]# curl -sk -o /dev/null -w '%{http_code} %{size_download} bytes\n' "$URL"
200 6295 bytes
[cc1 ~]# curl -sk -o /dev/null -w '%{http_code} %{size_download} bytes\n' https://10.1.0.1:6080/core/rfb.js
200 119862 bytes
[cc1 ~]# curl -skv -H 'Connection: Upgrade' -H 'Upgrade: websocket' -H 'Sec-WebSocket-Version: 13' -H 'Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==' "https://10.1.0.1:6080/websockify?token=$TOKEN" 2>&1 | grep -E '101|Sec-WebSocket-Accept'
< HTTP/1.1 101 Switching Protocols
< Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=
[cc1 ~]# grep 'connecting to' /var/log/nova/nova-novncproxy.log | tail -1
INFO nova.console.websocketproxy   7: connecting to: 10.1.0.1:5900
```

And the delete,

```bash
[cc1 ~]# openstack server delete test-ephemeral-vm
[cc1 ~]# openstack server list --name test-ephemeral-vm

[cc1 ~]# virsh list --all
 Id   Name   State
--------------------
```

Two further instance shapes, beyond what the issue asks for, because they are what exercise the privsep pins: a volume-backed instance on a Caracal-Cinder rbd volume, and an instance with a NIC whose tap os-vif plugged into `br-int`,

```bash
[cc1 ~]# openstack server show test-rootvol-vm -c status -c volumes_attached
| status           | ACTIVE                                                       |
| volumes_attached | delete_on_termination='False', id='2e172467-ed31-4100-90fd-4de154726225' |
[cc1 ~]# openstack volume list
| 2e172467-ed31-4100-90fd-4de154726225 |  | in-use | 2 | Attached to test-rootvol-vm on /dev/sda |
[cc1 ~]# openstack server show test-nic-vm -c status -c addresses
| addresses | nova-caracal-check=192.0.2.111 |
| status    | ACTIVE                         |
[cc1 ~]# grep 'Successfully plugged vif' /var/log/nova/nova-compute.log | tail -1
INFO os_vif Successfully plugged vif VIFOpenVSwitch(active=False,address=fa:16:3e:ce:27:4e,bridge_name='br-int',...,plugin='ovs',vif_name='tap2a257ccd-96')
```

UEFI, proving `[libvirt] ovmf_path` was never what selected the firmware. With the option absent from `nova.conf`, an image carrying `hw_firmware_type=uefi` boots and libvirt resolves the firmware itself from `/usr/share/qemu/firmware/*.json`,

```bash
[cc1 ~]# grep -c ovmf_path /etc/nova/nova.conf
0
[cc1 ~]# virsh dumpxml instance-0000000b | sed -n '/<os/,/<\/os>/p'
  <os firmware='efi'>
    <type arch='x86_64' machine='pc-q35-rhel9.8.0'>hvm</type>
    <firmware>
      <feature enabled='no' name='enrolled-keys'/>
      <feature enabled='no' name='secure-boot'/>
    </firmware>
    <loader readonly='yes' secure='no' type='pflash' format='raw'>/usr/share/edk2/ovmf/OVMF_CODE.fd</loader>
    <nvram template='/usr/share/edk2/ovmf/OVMF_VARS.fd' templateFormat='raw' format='raw'>/var/lib/libvirt/qemu/nvram/instance-0000000b_VARS.fd</nvram>
```

and the same image with `os_secure_boot=required` gets the secboot binary — which is the lever the deleted option looked like it provided, except that this one is conditional and routes through the `COMPUTE_SECURITY_UEFI_SECURE_BOOT` trait,

```bash
[cc1 ~]# virsh dumpxml instance-0000000c | sed -n '/<os/,/<\/os>/p'
  <os firmware='efi'>
    <type arch='x86_64' machine='pc-q35-rhel9.8.0'>hvm</type>
    <firmware>
      <feature enabled='yes' name='enrolled-keys'/>
      <feature enabled='yes' name='secure-boot'/>
    </firmware>
    <loader readonly='yes' secure='yes' type='pflash' format='raw'>/usr/share/edk2/ovmf/OVMF_CODE.secboot.fd</loader>
    <nvram template='/usr/share/edk2/ovmf/OVMF_VARS.secboot.fd' templateFormat='raw' format='raw'>/var/lib/libvirt/qemu/nvram/instance-0000000c_VARS.fd</nvram>
```

The five stray options, read back out of the installed packages rather than trusted from the file,

```bash
[cc1 ~]# /opt/openstack-caracal/bin/python -c "
import nova.conf; CONF = nova.conf.CONF
CONF(['--config-file','/etc/nova/nova.conf'], project='nova')
for g,o in (('neutron','url'),('libvirt','ovmf_path'),('libvirt','swtpm_enabled')):
    try: print('CONF.%s.%s = %r' % (g,o,getattr(getattr(CONF,g),o)))
    except Exception as e: print('CONF.%s.%s -> %s: %s' % (g,o,type(e).__name__,e))
print('placement.region_name =', CONF.placement.region_name)
print('neutron.region_name   =', CONF.neutron.region_name)
print('cinder.os_region_name =', CONF.cinder.os_region_name)
print('vnc.enabled           =', CONF.vnc.enabled)"
CONF.neutron.url -> NoSuchOptError: no such option url in group [neutron]
CONF.libvirt.ovmf_path -> NoSuchOptError: no such option ovmf_path in group [libvirt]
CONF.libvirt.swtpm_enabled = True
placement.region_name = RegionOne
neutron.region_name   = RegionOne
cinder.os_region_name = RegionOne
vnc.enabled           = True
[cc1 ~]# /opt/openstack-caracal/bin/python -c "
from oslo_config import cfg; import placement.conf
conf = cfg.ConfigOpts(); placement.conf.register_opts(conf)
conf(['--config-file','/etc/placement/placement.conf'])
print('placement randomize_allocation_candidates =', conf.placement.randomize_allocation_candidates)"
placement randomize_allocation_candidates = True
```

```bash
two phase boot... [yes]
cube services... [ready]
cluster data... [synced]
applying policy... [no]
          Service  Status  Report
           IaasDb      ok  [ mysql(v) ]
            Image      ok  [ glance(v) ]
            LBaaS      ok  [ octavia(v) ]
          Metrics      ok  [ monasca(v) telegraf(v) grafana(v) lachesis(v) ]
          Storage      ok  [ ceph(v) ceph_mon(v) ceph_mgr(v) ceph_mds(v) ceph_osd(v) ceph_rgw(v) rbd_target(v) fc_link(v) ]
```

For the qemu version lock, verified on `accept-3cc` — `cc1` taken to -23 with `cc2` and `cc3` left at -22, then every direction a roll can produce and one it cannot,

```bash
[cc1 ~]# /usr/libexec/qemu-kvm -M help | grep -E "^q35|^pc-q35-rhel9.8"
q35                  RHEL-9.8.0 PC (Q35 + ICH9, 2009) (alias of pc-q35-rhel9.8.0)
pc-q35-rhel9.8.0     RHEL-9.8.0 PC (Q35 + ICH9, 2009)
[cc1 ~]# mysql -N -B -e "select id, source_compute, dest_compute, status, created_at from nova.migrations where migration_type='live-migration' order by id desc limit 5"
41	cc2	cc1	completed	2026-08-20 12:49:01   # -22 -> -23, guest born on -23
38	cc1	cc2	completed	2026-08-20 12:48:37   # -23 -> -22, guest born on -23
35	cc1	cc3	completed	2026-08-20 12:47:15   # -23 -> -22, guest born on -22
32	cc3	cc1	completed	2026-08-20 12:46:18   # -22 -> -23, guest born on -22
29	cc1	cc3	completed	2026-08-20 12:44:30   # -22 -> -22 baseline
[cc1 ~]# for n in cc1 cc2 cc3; do echo -n "$n migration errors: "; ssh -n $n "grep -icE 'migration.*(fail|error)' /var/log/nova/nova-compute.log"; done
cc1 migration errors: 0
cc2 migration errors: 0
cc3 migration errors: 0
[cc1 ~]# for n in cc1 cc2 cc3; do echo -n "$n: "; ssh -n $n "rpm -q qemu-kvm"; done
cc1: qemu-kvm-10.1.0-23.el9.x86_64
cc2: qemu-kvm-10.1.0-23.el9.x86_64
cc3: qemu-kvm-10.1.0-23.el9.x86_64
```

Every instance stayed `ACTIVE` across all four migrations. All 18 qemu RPMs confirmed present at the pinned kojihub path so an air-gapped rebuild resolves the exact NVR.

One pre-existing condition, unchanged rather than a regression: `nova-compute` logs two `No DB access allowed in nova-compute` ERRORs at startup, from `Service.get_minimum_version_all_cells` attempting a direct read. The same two appear in the Antelope-era rotated log on this node.


Eandalf-Bigstack:
@traviswu-bigstack This PR is now the ready one for OpenStack upgrade.

traviswu-bigstack (approved):

PR #1331, opened 2026-08-24
fix(ceph,health): detect a damaged OSD and remount just that one, without auto-repairing storage

### LINKED ISSUE #1284: [Bug] Ceph OSD fails to reactivate after live kernel partition metadata loses PARTUUID and PARTLABEL
## Describe the bug

  A Ceph OSD can remain down after the kernel/udev live partition state loses the GPT `PARTUUID` and `PARTLABEL` for its BlueStore data partition, even though the GPT stored on disk and the BlueStore data remain valid.

  The observed failure affected `osd.26` on `bigcc3`.

  The on-disk GPT entry for `/dev/sde4` was valid and contained:

  - Partition GUID: `21B2E1E4-91D1-4D04-A922-740FBF6AC2ED`
  - Partition name: `cube_data_e_4`
  - Correct partition start and end sectors

  However, the running kernel reported an empty `PARTUUID` and `PARTLABEL` for `/dev/sde4`. As a result, the expected `/dev/disk/by-partuuid/...` device link was unavailable.

  The OSD metadata filesystem `/dev/sde2` was eventually left unmounted from `/var/lib/ceph/osd/ceph-26`. The OSD service then read the empty underlying directory and repeatedly failed with:

  ```text
  missing 'type' file and unable to infer osd type

  The service entered a restart loop and accumulated more than 69,000 restart attempts. Ceph automatically marked the OSD out, allowing the cluster to remain HEALTH_OK with reduced OSD capacity.

  The issue appears to involve the CubeCOS storage initialization and Ceph OSD activation path, including:

  - CubeCOS boot-time storage initialization
  - Ceph OSD discovery and activation
  - ceph-volume raw list
  - Kernel/udev partition-state handling
  - Multipath and partx integration
  - /store/ceph/osd/dev_osd.map
  - ceph-osd@.service

  The exact CubeCOS repository or function responsible for the partition refresh has not yet been identified.

  ## Version

  - Product/Fixpack version: CUBE_3.1.0_20260313-1639_f58d033
  - CubeCOS API package: cube-cos-api-v3.1.0-1.el9.5310e06
  - Ceph version: 17.2.6 (Quincy)
  - Operating system family: Enterprise Linux 9
  - kpartx version: 0.8.7-45.el9
  - Affected product version: CubeCOS 3.1.0

  ## To Reproduce

  The issue has not yet been reproduced deterministically. The following sequence reproduces the resulting failure state without modifying the GPT stored on disk:

  1. Deploy a CubeCOS node with raw BlueStore OSD partitions.
  2. Confirm the OSD data partition has a valid GPT UUID and label:

     sgdisk -i 4 /dev/sde

  3. Confirm the live kernel state agrees with the on-disk GPT:

     lsblk -o NAME,PARTUUID,PARTLABEL,FSTYPE /dev/sde

  4. Cause partition 4 to be removed and re-added in the running kernel through a BLKPG/partx path that does not preserve the partition metadata.
  5. Observe that the partition number, start sector and size remain available, but the live PARTUUID and PARTLABEL are empty:

     lsblk -o NAME,PARTUUID,PARTLABEL,FSTYPE /dev/sde

  6. Restart or re-run the CubeCOS/Ceph OSD activation path.
  7. Observe that the stable /dev/disk/by-partuuid/... link is unavailable and the corresponding OSD is not activated correctly.
  8. If the OSD metadata partition is also left unmounted, start the OSD service:

     systemctl start ceph-osd@26

  9. Observe the service failure:

     missing 'type' file and unable to infer osd type

  The original incident timeline was:

  - 2026-07-30 23:17: /dev/sde4 still reported PARTLABEL="cube_data_e_4" and FSTYPE="ceph_bluestore".
  - 2026-08-02 01:38: the boot-time ceph-volume raw list operation first reported empty PARTLABEL and FSTYPE for /dev/sde4.
  - Other OSD partitions continued to report correct metadata.
  - osd.26 subsequently remained down and was automatically marked out.

  ## Expected behavior

  CubeCOS should reliably activate every valid Ceph OSD whose metadata and BlueStore labels remain intact.

  Specifically:

  1. Boot-time storage initialization must not remove or recreate local OSD partition nodes in a way that loses their live PARTUUID or PARTLABEL.
  2. Local PERC-backed SSDs must not be treated as multipath slave devices.
  3. OSD activation should compare the live kernel partition identity with the GPT stored on disk.
  4. If live partition metadata is missing but the GPT and BlueStore label are valid, CubeCOS should recover the partition mapping or emit an actionable alert.
  5. A single activation failure must not leave the OSD service in an unbounded restart loop.
  6. After node boot, all valid OSD metadata filesystems should be mounted and all corresponding OSD services should become active.

  ## Root Cause (if known)

  The original assumption that GPT partition 4 had lost its GUID and label was disproven by a forensic backup taken before repair.

  The pre-repair GPT backup contained a fully valid partition 4 entry:

  Partition GUID code: 0FC63DAF-8483-4772-8E79-3D69D8477DE4
  Partition unique GUID: 21B2E1E4-91D1-4D04-A922-740FBF6AC2ED
  First sector: 782231552
  Last sector: 1562822655
  Partition name: 'cube_data_e_4'

  sgdisk -v also reported no GPT consistency problems.

  The CubeCOS OSD mapping file contained the same original GUID:

  /store/ceph/osd/dev_osd.map:

  /dev/sde2 26 cc583d87-30eb-4c36-9a5c-5a3493e13045 21b2e1e4-91d1-4d04-a922-740fbf6ac2ed

  The failure was therefore in the running kernel/udev partition state, not in the GPT stored on disk.

  The most likely mechanism is that /dev/sde4 was deleted and re-added in the running kernel through a BLKPG/partx operation. This explains why the partition number, boundaries and size remained available while its live GUID and name disappeared.

  A relevant installed udev rule exists at:

  /usr/lib/udev/rules.d/68-del-part-nodes.rules

  It can execute:

  /usr/sbin/partx -d --nr 1-1024 $env{DEVNAME}

  when a device is identified as a multipath slave:

  ENV{DM_MULTIPATH_DEVICE_PATH}=="1", GOTO="del_part_nodes"

  The affected local SSD currently reports:

  DM_MULTIPATH_DEVICE_PATH=0

  However, a transient boot-time multipath misclassification or device-event race could have triggered partition-node deletion. The current evidence is consistent with this mechanism but does not conclusively prove which process invoked partx.

  The original audit and journal records covering the transition are no longer retained, so the exact process cannot be identified conclusively.

  This does not appear to be an SSD media failure:

  - SMART overall health passed.
  - No reallocated sectors were reported.
  - No pending sectors were reported.
  - No uncorrectable sectors were reported.
  - No CRC errors were reported.
  - No relevant kernel I/O errors were found.
  - ceph-bluestore-tool show-label --dev /dev/sde4 successfully read the OSD label.

  ## Suggested Fix (Optional)

  1. Add an explicit multipath blacklist for the local PERC-backed Intel SSDs.
  2. Add a CubeCOS udev override that prevents local OSD partition nodes from being deleted:

     ENV{ID_SERIAL}=="INTEL_SSDSC2BB800G6_*", ENV{DONT_DEL_PART_NODES}="1"

     The final rule should use the appropriate CubeCOS-supported device matching criteria and must not interfere with actual SAN multipath devices.

  3. During OSD activation, validate live partition metadata against the GPT stored on disk:

     lsblk -o NAME,PARTUUID,PARTLABEL
     sgdisk -i <partition-number> <disk>

  4. If the live kernel metadata is missing while the on-disk GPT is valid, perform a safe partition-table refresh before OSD activation.
  5. Validate /store/ceph/osd/dev_osd.map against:
      - The on-disk GPT UUID
      - The live /dev/disk/by-partuuid links
      - The BlueStore osd_uuid
      - The expected OSD ID

  6. Fail the activation process with an actionable error when the device mapping is inconsistent. Do not repeatedly start ceph-osd@<id> against an empty, unmounted directory.
  7. Add a bounded restart policy or start-limit protection for ceph-osd@.service to prevent unlimited restart loops.
  8. Preserve boot-time storage activation logs long enough to investigate failures.
  9. Add monitoring for differences between the live kernel partition identity and the on-disk GPT identity.

  ## Screenshots (Optional)

  Not applicable. CLI output and logs are more relevant for this issue.

  - Cluster topology: three control-converged nodes
  - Ceph topology:
      - 3 monitors
      - 3 managers
      - 36 OSDs
      - 12 OSDs per node

  - Affected node: bigcc3
  - Affected OSD: osd.26
  - OSD metadata device: /dev/sde2
  - OSD BlueStore data device: /dev/sde4
  - Hardware:
      - Dell PowerEdge server
      - PERC H730P Mini
      - Intel SSDSC2BB800G6 SSDs

  - OSD metadata filesystem: XFS
  - OSD data backend: raw BlueStore

  Relevant service error:

  ceph-osd@26.service: Main process exited, code=exited, status=1/FAILURE
  missing 'type' file and unable to infer osd type

  Relevant live-state discrepancy:

  NAME   PARTUUID   PARTLABEL
  sde4

  Relevant on-disk GPT state:

  Partition unique GUID: 21B2E1E4-91D1-4D04-A922-740FBF6AC2ED
  Partition name: 'cube_data_e_4'

  The OSD was recovered by:

  1. Mounting /dev/sde2 at /var/lib/ceph/osd/ceph-26.
  2. Restoring the OSD block link to /dev/sde4.
  3. Correcting the block-device ownership for the ceph user.
  4. Starting ceph-osd@26.

  Final validation:

  cluster:
    health: HEALTH_OK

  osd:
    36 osds: 36 up, 36 in

  pgs:
    1457 active+clean

  Forensic GPT backups retained on bigcc3:

  /root/sde-gpt-before-osd26-20260817.bin
  /root/sde-gpt-after-osd26-repair-20260817.bin

  Three Dell BMC OEM records occurred near the failure window:

  2026-07-30 22:45:14
  2026-08-02 01:10:32
  2026-08-02 02:17:44

  They were reported as OEM timestamped records and could not be reliably decoded with the installed tooling.

leechienpang:
#### What type of PR is this?

/kind bug

Reworked to complement #1399 rather than overlap it. #1399 (**merged**) owns **prevention** —
`create_map` no longer discards an OSD when udev loses a `/dev` link, and `prepareOsdDirectories`
no longer re-creates one whose directory went missing. This PR owns **detection and narrow
recovery**: find an OSD that is already damaged, and repair exactly that one, without writing to
anything.

Rebased onto `develop` after #1399 merged; `ceph_osd_remount_one` and `ceph_osd_damaged_list` build
on its `ceph_osd_datapart_resolve` helper.

The premise is that automatic repair of storage papers over data loss. In the reported incident
four independent layers reported "fine" over one dead OSD: `create_map` destroyed state believing it
was tidying up; `health_ceph_osd_repair` restarted the OSD blindly (69,000 times, never escalating);
`ceph.mk` comments out `StartLimitInterval` so the loop is invisible rather than a failed unit; and
ceph marked the OSD out, returning the cluster to HEALTH_OK. Underneath all of it,
`health_ceph_osd_check` was force-disabled on every boot, so nothing was being detected at all.

- **`ceph_osd_refresh_links`** — re-trigger udev for a partition whose live PARTUUID is missing.
  A view refresh only: no partition-table write, no partition node removed. Reports failure instead
  of escalating to anything destructive.
- **`ceph_osd_remount_one`** — mount and re-link a single OSD. Refuses to act unless the data device
  actually resolves, and never touches another OSD on the host, unlike `ceph_osd_remount` /
  `refresh_ceph_osd` which stop and unmount every OSD there.
- **`ceph_osd_damaged_list`** — report local OSDs whose metadata dir is unmounted, or whose `block`
  symlink is missing or dangling, while the data device is demonstrably still present. Reports;
  never repairs.
- **`health_ceph_osd_check`** — new `ERR_CODE 4` for that state. Ceph's own view goes green once an
  unusable OSD is marked out, so this was previously invisible.
- **`health_ceph_osd_repair`** — remount the OSD before restarting it. An OSD whose metadata dir is
  not mounted fails identically on every attempt, so restarting it is a loop, not a repair. If it
  cannot be remounted safely, log and leave it down for inspection rather than churn.
- **`cube_cluster_start_node`** — stop force-disabling the check every boot.

There is deliberately **no** `_health_ceph_osd_auto_repair`. Detection is worth having on by
default; automatic repair of storage is not.

Related to #1284
Related to #1398

**Dropped from the previous revision of this PR**, with reasons:

- the `create_map` PARTUUID-refresh hunk — #1399 resolves the uuid off the disk, so there is
  nothing left to refresh for that decision, and its `else` branch still dropped the OSD from the
  map on failure.
- `partprobe` as the recovery mechanism. On a live OSD disk `BLKRRPART` returns EBUSY and libparted
  falls back to per-partition BLKPG delete+add — the operation that produces a partition node with
  no GPT-derived metadata, and in this layout it would touch the sibling OSD on the same disk.
  `udevadm trigger` restores the links without mutating anything, which is what actually worked in
  every lab run.
- `_health_ceph_osd_auto_repair` — it re-armed blind restart looping, and was inert anyway while the
  check was disabled.

`ceph_osd_remount_one` is @leechienpang's idea from the previous revision and the reason this PR is
worth keeping rather than closing; it is reworked here (device-resolution guard, no partprobe, honest
return codes) rather than replaced.

**Third commit is a behaviour change worth its own look.** `ceph_osd` is in
`ROLLING_ESSENTIAL_TIER2`, so re-enabling detection means the rolling restart's essential gate will
actually evaluate OSDs instead of always passing. The existing check already tolerates rejoining
OSDs (`_ceph_blocking_health`), so this should be safe, but it can surface NG during rolls that
previously ran blind. Kept as a separate commit so it can be dropped independently.

**Validated on the sky 3-node lab (sky143), 2026-09-01.** No false positives on a healthy cluster;
detection names the damaged OSD; `ceph_osd_remount_one` repairs it while the other five OSDs on the
host stay untouched; the OSD then starts and detection clears; and it refuses to repair a phantom
OSD whose device resolves nowhere, creating no directory for it. Cluster returned to HEALTH_OK 18/18
after every round.

The before/after that motivates the third commit, same damaged OSD:

```
with the disabled marker present (today):  health_ceph_osd_check rc=0    <- reports OK
with the marker lifted (this change):      health_ceph_osd_check rc=4
  error desc: osd on-disk state damaged
  osd.8 metadata dir not mounted (/var/lib/ceph/osd/ceph-8)
```

Not covered: LVM-backed (encrypted/mpath) OSDs — the lab cluster has none, so the `lvs` branch of
`ceph_osd_datapart_resolve` (merged in #1399, relied on here) has no hardware behind it.

#### Additional documentation

```docs

```

### DIFF
--- a/core/sdk_sh/modules/sdk_health.sh
+++ b/core/sdk_sh/modules/sdk_health.sh
@@ -1639,10 +1639,17 @@ health_ceph_osd_check()
     local total=$(echo $stats | jq -r .osdmap.num_osds)
     local uposd=$(echo $stats | jq -r .osdmap.num_up_osds)
     local inosd=$(echo $stats | jq -r .osdmap.num_in_osds)
+    local damaged="$($HEX_SDK ceph_osd_damaged_list 2>/dev/null)"
+
     if cmd "$HEX_SDK ceph_osd_list" | sort -t. -n -k2 | egrep -q "warning|fail" ; then
         # device hardware fault -- always a fault, even mid-recovery
         ERR_CODE=3
         ERR_LOG="dmesg | grep -i -e fail -e error"
+    elif [ -n "$damaged" ] ; then
+        # on-disk state damaged while the data device is still present; ceph can
+        # report this host healthy after marking the OSD out (cubecos#1284)
+        ERR_CODE=4
+        ERR_MSG+="$damaged\n"
     elif [ "$total" != "$uposd" ] ; then
         # OSDs still rejoining -- only a fault if data is unavailable
         _ceph_blocking_health "$stats" && ERR_CODE=1
@@ -1656,15 +1663,28 @@ health_ceph_osd_check()
 
 health_ceph_osd_repair()
 {
-    if [ $($CEPH osd tree down | grep host | awk '{print $4}' | sort -u | wc -l) -gt 0 ] ; then
-        readarray osd_array <<<"$($CEPH osd tree | awk '/ down /{print $1}' | sort)"
-        declare -p osd_array > /dev/null
-        for osd_entry in "${osd_array[@]}" ; do
-            local osd=$(echo $osd_entry | head -c -1)
-            local host=$(ceph_get_host_by_id $osd)
-            remote_run $host "timeout $SRVLTO $HEX_SDK ceph_osd_restart $osd"
-        done
+    if [ $($CEPH osd tree down | grep host | awk '{print $4}' | sort -u | wc -l) -eq 0 ] ; then
+        return 0
     fi
+
+    readarray osd_array <<<"$($CEPH osd tree | awk '/ down /{print $1}' | sort)"
+    declare -p osd_array > /dev/null
+    for osd_entry in "${osd_array[@]}" ; do
+        local osd=$(echo $osd_entry | head -c -1)
+        local host=$(ceph_get_host_by_id $osd)
+
+        # A down OSD whose metadata dir isn't mounted will fail the same way on
+        # every restart ("missing 'type' file"), so restarting it is not a
+        # repair -- it is a loop. Remount just this OSD first; if that cannot be
+        # done safely, report and leave it alone rather than churn (cubecos#1284).
+        if ! remote_run $host "test -f /var/lib/ceph/osd/ceph-$osd/type" ; then
+            if ! remote_run $host "$HEX_SDK ceph_osd_remount_one $osd" ; then
+                log_error "health_ceph_osd_repair: osd.$osd on $host cannot be remounted safely; leaving it down for inspection"
+                continue
+            fi
+        fi
+        remote_run $host "timeout $SRVLTO $HEX_SDK ceph_osd_restart $osd"
+    done
 }
 
 health_ceph_rgw_report()
--- a/core/main/proj_functions
+++ b/core/main/proj_functions
@@ -313,7 +313,10 @@ cube_cluster_start_node()
     # nova-compute as a failure and masakari flags the host on_maintenance.
     # They are armed once, cluster-wide, in cube_cluster_start_cluster.
     Quiet -n killall /usr/sbin/hex_banner # refresh console banner
-    ( $HEX_SDK toggle_health_check | grep ceph_osd | grep -q disabled ) || Quiet -n $HEX_SDK toggle_health_check ceph_osd
+    # keep ceph_osd detection on: a disabled check reports ok, which also silences
+    # health_ceph_osd_repair and the roll's essential gate. Nothing auto-repairs
+    # storage -- there is no _health_ceph_osd_auto_repair.
+    ( $HEX_SDK toggle_health_check | grep ceph_osd | grep -q disabled ) && Quiet -n $HEX_SDK toggle_health_check ceph_osd
     iptables-save > /run/iptables
     cmd "$HEX_SDK migrate_git" # init this node's /.git before it's marked synced
     touch $CLUSTER_SYNC
--- a/core/sdk_sh/modules/errcodes
+++ b/core/sdk_sh/modules/errcodes
@@ -141,6 +141,7 @@ SRV="ceph_osd"
 eval "ERR_DESC_$SRV[1]='not all up'"
 eval "ERR_DESC_$SRV[2]='not all in'"
 eval "ERR_DESC_$SRV[3]='disk failing'"
+eval "ERR_DESC_$SRV[4]='osd on-disk state damaged'"
 eval for idx in \"\${!ERR_DESC_${SRV}[@]}\" \; do ERR_CODE[${SRV},\$idx]=\"\${ERR_DESC_${SRV}[\$idx]}\" \; done
 
 SRV="ceph_rgw"
--- a/core/sdk_sh/modules/sdk_ceph.sh
+++ b/core/sdk_sh/modules/sdk_ceph.sh
@@ -2564,6 +2564,119 @@ ceph_osd_get_datapartuuid()
     echo -n "$datapart_partuuid"
 }
 
+# Ask udev to re-read a disk and re-populate its /dev/disk entries. This is a
+# view refresh only: it writes nothing to the disk, does not touch the partition
+# table, and leaves every partition node in place (cubecos#1284).
+# params: $1 - partition device path (e.g. /dev/sde4, /dev/nvme0n1p4)
+# returns: 0 if the partition's PARTUUID is resolvable afterwards, 1 otherwise
+ceph_osd_refresh_links()
+{
+    local datapart=$1
+    local disk=
+
+    [ -n "$(GetBlkPartUuid "$datapart")" ] && return 0
+
+    if [[ "$datapart" =~ ^(.*[0-9]+)p[0-9]+$ ]] ; then
+        disk="${BASH_REMATCH[1]}"
+    elif [[ "$datapart" =~ ^(.*[a-z])[0-9]+$ ]] ; then
+        disk="${BASH_REMATCH[1]}"
+    else
+        log_error "ceph_osd_refresh_links: cannot derive parent disk from $datapart"
+        return 1
+    fi
+
+    log_warning "ceph_osd_refresh_links: $datapart has no live PARTUUID, re-triggering udev on $disk"
+    udevadm trigger --action=change "$disk" "$datapart" 2>/dev/null
+    udevadm settle --timeout=30 2>/dev/null
+
+    if [ -n "$(GetBlkPartUuid "$datapart")" ] ; then
+        return 0
+    fi
+
+    log_error "ceph_osd_refresh_links: $datapart still has no live PARTUUID after a udev refresh -- inspect the disk before taking any repair action"
+    return 1
+}
+
+# Remount a single down OSD whose metadata dir is missing or unmounted. Scoped
+# to one OSD: unlike refresh_ceph_osd / ceph_osd_remount it never stops or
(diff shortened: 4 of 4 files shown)


core/sdk_sh/modules/sdk_health.sh
```diff
@@ -1652,6 +1652,16 @@ health_ceph_osd_check()
     _health_fail_log
 }
 
+_health_ceph_osd_auto_repair()
+{
+    # ERR_CODE 3 is a real hardware fault (device fail/error in dmesg) -- don't
+    # mask it behind an automatic repair attempt. 1/2 (down/out, non-blocking)
+    # are the cases health_ceph_osd_repair can actually do something about.
+    if [ "$ERR_CODE" == "1" -o "$ERR_CODE" == "2" ] ; then
+        cmd $HEX_SDK health_ceph_osd_repair
```
traviswu-bigstack: How much time does health_ceph_osd_repair usually take? auto_reapir() should consider fast and non-distruptive

Eandalf-Bigstack:
We are missing the DCO sign-off line on the last commit's commit message.

traviswu-bigstack (commented):
I would like to know why "the running kernel forgets its live PARTUUID/PARTLABEL" and how does it happen? That's root cause we need to find out. I dont feel using auto_repair to fix this is a right way

leechienpang:
That's sth. I'd also like to know. Until this can be reproduced consistently, having a repair attempt in health_ceph_osd_repair() is the way to go.  Nothing new has been added into health_ceph_osd_auto_repair()

traviswu-bigstack:
No rush to fix it. I will check with @Bigstack-Nate first

Eandalf-Bigstack (approved):

Issue #1316, opened 2026-08-21
[Feature] Carve vGPUs of different sizes on one SR-IOV card


SekiXu:
## Description

> A brief summary to help everyone understand this feature.

One SR-IOV card can currently host vGPUs of **one framebuffer size only**. This feature lifts
that to different sizes on the same card.

**Measured today on a 4× RTX PRO 6000 Blackwell node.** `hex_config gpu_resource_set … sriovVgpu`
already accepts a multi-profile payload, and profiles that share a capacity apply fine:

```bash
# same capacity, different mode — DC-3Q (3072 MiB) + DC-3A (3072 MiB)
hex_config gpu_resource_set <gpu> sriovVgpu '[{"id":1519,"count":1},{"id":1520,"count":1}]'
# rc=0 · virtfn0=1519 · virtfn1=1520
```

A cross-capacity request does not:

```bash
# DC-3Q (3072 MiB) + DC-6Q (6144 MiB)
hex_config gpu_resource_set <gpu> sriovVgpu '[{"id":1519,"count":1},{"id":1523,"count":1}]'
# rc=1
# Error: gpu_resource_set: failed to set vGPU type 1523 on 0001:c8:00.0/virtfn1
# Error: gpu_resource_set: failed to apply SR-IOV vGPU partitioning on GPU ...
```

The first profile writes, the second is refused by the driver, and the card is torn back down.

**The hardware supports it — the mode is off.** `nvidia-smi -q` on that same card reports both
of these at once:

```
vGPU Driver Capability
    Heterogenous Multi-vGPU           : Supported
vGPU Device Capability
    Heterogeneous Time-Slice Profiles : Supported
    Heterogeneous Time-Slice Sizes    : Supported
    Host VGPU Mode                    : SR-IOV
    vGPU Heterogeneous Mode           : Disabled
```

Nothing in `core/modules/config_gpu.cpp` or `core/sdk_sh/modules/sdk_gpu.sh` ever switches that
mode. The single-capacity path delivered by #819 / #904 works as designed and never claimed this
one; the restriction is not recorded as a limitation on those tickets, which is why this is a new
Feature rather than a Bug against them.

> ### Update 2026-08-21 — measured on cn13, and it changes the premise
>
> The mode switch is `nvidia-smi vgpu -shm` (`--set-heterogeneous-mode`). There is **no sysfs
> route**: the PF has no `nvidia/` directory, only the VFs do.
>
[...]

PR #1338, opened 2026-08-24
feat(gpu): carve vGPUs of different framebuffer sizes on one SR-IOV card

SekiXu:
Closes the two halves of #1316: a card can now be carved into vGPUs of different
framebuffer sizes, and each generated Nova PCI alias identifies the profile its
VF actually carries.

## The mode half

One SR-IOV card could host vGPUs of a single framebuffer size only — after the
first `current_vgpu_type` write every remaining VF's `creatable_vgpu_types`
collapsed to that size, and a cross-size write was refused. The hardware allows
the mix (`Heterogenous Multi-vGPU : Supported`); nothing ever switched the mode
on.

Three things the ticket assumed turned out differently, all measured on a
4× RTX PRO 6000 Blackwell node:

- **The mode is runtime state, not persistent.** `sriov-manage` clears it in both
  directions, so — unlike MIG mode — teardown needs no explicit mode-off, a
  failed apply needs no rollback, and the intent file needs no new field.
- **Ordering is not negotiable, and the obvious order is wrong.** It has to be
  set *after* `sriov-manage -e`. Set before, it reads back `Enabled`, is then
  wiped by the driver rebind, and every VF stays collapsed — a build that looks
  configured and is not. Ten VFs were scanned in that state before the cause was
  found; the symptom is indistinguishable from "the hardware cannot do it".
- **No placement planning is needed.** With the ordering right the driver picks
  the aligned placement itself (`placement_id` 0 and 6 for a 3072 + 6144 carve).

Validation refuses a mixed-size request on a card that does not advertise the
capability instead of letting it reach the apply, because by then
`gpu_unset_current_type` has already destroyed the card's previous carve. It
runs ahead of the `sriov_totalvfs` capacity rule since it reads nothing from
sysfs.

## The alias half

Every vGPU VF on a node reports the same `vendor_id:product_id`, and a
`[pci]alias` carries no address — only `device_spec` does. So the aliases
generated for a multi-profile card were byte-identical as match criteria and the
scheduler had nothing to tell them apart by.

Measured, with two free VFs of different size on one card and one VM per alias:
**wrong 4 times out of 4**, in both boot orders, with nothing reported by either
the guest or Nova. (An earlier 3/3-correct observation on the same node does not
contradict it: those VMs were booted one at a time with a single free VF, so
there was never a choice to get wrong.)

Each alias and each `device_spec` entry now carries a `vgpu_type` tag naming the
profile, and the whitelist loop walks the profiles in request order — the same
expansion `BuildVfAssignmentPlan` performs — so the tag matches what the driver
actually put on that VF. Nova needed one property added to its alias schema;
everything downstream already worked, and `additionalProperties: False` was the
only obstacle. **After the change: correct 4 times out of 4, both boot orders.**

## Verification

Hardware, on a card reporting `Heterogenous Multi-vGPU : Supported`:

| | |
| --- | --- |
| shipped single-size path | `rc=0`, both VFs the requested type, `allocation.total` 2, nova-compute active |
| mixed-size carve | `rc=0` (was `rc=1`), mode `Enabled`, types in request order at `placement_id` 0 / 6, free VFs offering 23 types across 9 capacity classes |
| alias → VF mapping | **4/4 correct**, both boot orders |
| teardown | `sriov_numvfs` 0, PF back on `vfio-pci`, the card's drop-in entries gone |
| failed apply | intent entry removed, `sriov_numvfs` 0, drop-in 0 entries — the three views agree instead of diverging |
| reboot re-apply | mode and carve rebuilt, `config.json` and `gpu.conf` byte-identical, **twice** |

Also measured, to show the shipped path is untouched: `nvidia-smi vgpu -s -v` is
byte-identical with the mode on and off, a four-way same-size carve produces the
same types and placements, and the 32-vGPU ceiling is unchanged.

Unit coverage is limited to what hardware cannot show — a card without the
capability, and an nvidia-smi that fails. See the test commit for why it stops
at two cases.

## Two things a reviewer should know

- **The `caracal_patch` change has not been exercised on a Caracal tree.** The
  alias mechanism was verified on the only vGPU-capable node available, which
  runs nova 25.3.0; the equivalent patch was applied there. All three trees
  (Yoga, Antelope, Caracal) were read line by line and are identical on every
  point this depends on — the leftover alias keys reaching
  `InstancePCIRequest.spec`, `PciDeviceStats` keeping unrecognised `device_spec`
  keys as pool tags, and `_filter_pools_for_spec` matching "any of the other
  arbitrary tags" — but reading is not running. It needs an ISO.
- **`gpu.conf` content changes for single-profile cards too**, by design. The tag
  has to be on every entry: an alias without it is unconstrained by that key and
  would match a tagged pool belonging to a different card. #1316's DoD was
  reworded accordingly — the requirement is that the shipped path still
  schedules correctly, not that the file is byte-identical.

Refs #1316

### DIFF
--- a/core/modules/config_gpu.cpp
+++ b/core/modules/config_gpu.cpp
@@ -789,6 +789,17 @@ IsVfBackedType(const json11::Json& type)
 // of every sriovVgpu/migBackedVgpu GPU. Regenerating everything from the
 // truth file means entries for GPUs switched away from either type
 // disappear without any dedicated cleanup logic.
+//
+// Every entry carries a vgpu_type tag naming the profile its VF was carved
+// to. Nova keeps unrecognised device_spec keys as pool tags and matches them
+// against the keys present in a request's spec, so this tag is the only thing
+// that stops two aliases on one card from being interchangeable: without it
+// every vGPU VF on the node reports the same vendor:product and an alias has
+// no address field, so a flavor gets whichever VF the pool's free list hands
+// out. Measured wrong 4 times out of 4 on cn13 2026-08-21, in both boot
+// orders. The tag has to appear on *every* entry, including single-profile
+// cards - an alias without it is unconstrained by that key and will match a
+// tagged pool belonging to a different card.
 static std::string
 BuildNovaGpuConfContent(const json11::Json& gpuConfig,
                         const std::map<std::string, std::vector<PciVf>>& vfsByGpuId)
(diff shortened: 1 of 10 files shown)


core/modules/config_gpu.cpp
```diff
@@ -1245,8 +1471,32 @@ ResourceSetMain(int argc, char* argv[])
...
+                if (!WriteNovaGpuConf()) {
+                    HexLogError("gpu_resource_set: failed to regenerate %s after the failed apply "
+                                "on GPU %s", NOVA_GPU_CONF, gpuId);
+                } else if (HexUtilSystemF(0, 0, HEX_CFG " restart_nova") != 0) {
+                    HexLogError("gpu_resource_set: failed to restart nova services after the failed "
+                                "apply on GPU %s", gpuId);
+                }
+            }
+
             return EXIT_FAILURE;
```
Eandalf-Bigstack, line 1509: It is recommended to disambiguate the branching statements.

```cpp
            if (!DropGpuConfigEntry(gpuId)) {
                HexLogError("gpu_resource_set: GPU %s also failed to drop its stale config entry; "
                            "the intent file still describes a carve the hardware no longer has",
                            gpuId);
                return EXIT_FAILURE;
            }

            if (access(NOVA_GPU_CONF, F_OK) != 0) {
                return EXIT_FAILURE;
            }

            if (!WriteNovaGpuConf()) {
                HexLogError("gpu_resource_set: failed to regenerate %s after the failed apply "
                            "on GPU %s", NOVA_GPU_CONF, gpuId);
                return EXIT_FAILURE;
            }

            if (HexUtilSystemF(0, 0, HEX_CFG " restart_nova") != 0) {
                HexLogError("gpu_resource_set: failed to restart nova services after the failed "
                            "apply on GPU %s", gpuId);
            }

            return EXIT_FAILURE;
```

Eandalf-Bigstack (commented):
Great work on solving the heterogeneous vGPU mode, especially on finding the minimal patch to OpenStack Nova to set different VF types on the same PF, which is not currently supported and not planned to be supported by Nova / Placement community. Only one change is recommended for disentangling the branching statements.

core/modules/config_gpu.cpp
SekiXu: Done, flattened to sequential early returns as suggested, thanks!
Amended into the original commit (now 9f108600) rather than stacked, to keep one logical change per commit under fast-forward merge.

Eandalf-Bigstack (approved):

PR #1347, opened 2026-08-26
fix(keycloak): create the database without the archived terraform mysql provider

### LINKED ISSUE #1344: [Bug] terraform mysql provider cannot create users on MariaDB 10.11 — keycloak DB user is never created, every fresh install hangs
## Description

Since `48a9533 feat(mariadb): bump mariadb from 10.6.x to 10.11.x for openstack caracal`, the image
ships **MariaDB 10.11.18**, which removed the `innodb_version` system variable. `core/terraform/src/mysql`
pins the archived `terraform-providers/mysql ~> 1.6`, whose `mysql_user` resource probes that variable —
so **every user it should create fails**, while `mysql_database` (which does not probe it) succeeds.

The only consumer today is keycloak (`config_keycloak.cpp: setupDatabase()`); every other service creates
its DB user through the hex_sdk mysql helpers and is unaffected.

Result on a fresh 3cc install (sky, image `CUBE_3.1.20_20260825-1849_b0162f9`):

```
keycloak database        : present
mysql.user 'keycloak'    : ABSENT          <- select user,host from mysql.user where user like 'keycloak%' => empty
keycloak-keycloakx-0     : CrashLoopBackOff (52 restarts)
                           "Access denied for user 'keycloak'@'10.32.10.141' (using password: YES)"
hex_config apply         : looping on the keycloak SAML descriptor fetch every 10s for 4h+
```

Reproduced by hand on the node:

```
# cd /var/lib/terraform && terraform apply -auto-approve -target=module.mysql -var mysql_dbname=keycloak
Error: Error 1193: Unknown system variable 'innodb_version'
  on mysql/main.tf line 19, in resource "mysql_user" "user_localhost"
  on mysql/main.tf line 32, in resource "mysql_user" "user_any"
```

`show variables like 'innodb_version'` returns nothing on 10.11.18 (it existed on 10.6).

**Any fresh install off current develop cannot finish FTS.** It is not a slow install — apply never
returns, because keycloak can never start and the SAML fetch has no deadline.

## Suggested fix

Stop creating DB users through that provider. The mysql module is only used to make a database plus its
user/grant, which the sdk already does everywhere else — replace `ExecTerraform("apply","mysql",…)` in
`setupDatabase()` with the same helper path the other services use, and delete the `mysql_user`/`mysql_grant`
resources. That also removes an archived provider from the install path.

If the provider must stay, the alternatives are worse: pinning MariaDB back to 10.6 undoes the caracal bump,
and patching the vendored provider to skip the `innodb_version` probe means carrying a fork.

See also: the idempotency check that turns this failure permanent (`[Bug] keycloak setupDatabase() treats a
half-created database as done`).

## How to Reproduce

1. Install any current-develop image (3cc or 1cc).
2. `mysql -N -e "select user,host from mysql.user where user like 'keycloak%'"` — empty.
3. `kubectl -n keycloak get pod` — `keycloak-keycloakx-0` CrashLoopBackOff, log shows `Access denied`.
4. `tail -f /var/log/hex_config.log` — the SAML descriptor fetch repeats every 10s indefinitely.

## Expected behavior

FTS creates the keycloak DB **and** its user, keycloak starts, the SAML fetch succeeds, and apply completes.

## QA Verification

|---|---|---|
| 1 | Fresh install from a build with the fix | FTS completes without manual intervention |
| 2 | `mysql -N -e "select user,host from mysql.user where user='keycloak'"` | two rows: `keycloak@localhost`, `keycloak@%` |
| 3 | `kubectl -n keycloak get pod keycloak-keycloakx-0` | `Running`, 0 restarts after settle |
| 4 | `curl -k https://<vip>:10443/auth/realms/master/protocol/saml/descriptor` | 200 with the descriptor |
| 5 | Re-run `hex_config apply` on the installed node | idempotent; no duplicate-user error |

- [ ] **QA Status declared at Done**

- [x] **Handbook knowledge update** — new `kb/cubecos/known-issues/` entry: MariaDB 10.11 drops
      `innodb_version`, archived terraform mysql provider cannot create users, symptom is a keycloak
      CrashLoop and a never-ending FTS.
      Landed in handbook [#394](https://github.com/bigstack-oss/bigstack-handbook/pull/394): [`known-issues/mariadb-1011-breaks-terraform-mysql-user.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/mariadb-1011-breaks-terraform-mysql-user.md)
      — `status: fixed-in:3.2.0`, corrected to describe what shipped in #1347, plus its
      [reproducible fault record](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/mariadb-1011-breaks-terraform-mysql-user.scenario.yaml).
      Two further entries came out of the same diagnosis: [what the SAML gate costs while stuck](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/keycloak-saml-metadata-never-retried.md)
      (it holds `hex_config.commit.lock`, so no repair command can run) and the rule it implies for
      [module authors](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/architecture/authoring-hex-config-modules.md).

### LINKED ISSUE #1345: [Bug] keycloak setupDatabase() treats a half-created database as done, so a failed user creation is never retried
## Description

`config_keycloak.cpp`:

```cpp
static bool isDatabaseSetup() { return MysqlUtilIsDbExist("keycloak"); }

static bool setupDatabase() {
    if (isDatabaseSetup()) { HexLogInfo("the database for keycloak is already created"); return true; }
    if (!ExecTerraform("apply", "mysql", { "mysql_dbname=" + DB_NAME }, {})) { ... }
```

The terraform module creates three things — the database, `keycloak@localhost` / `keycloak@%`, and their
grants. The "already set up" test checks **only the database**. A run that creates the database and then
fails before the users leaves a state that this function calls success **forever after**: the DB exists, so
every later apply returns early and the missing user is never created.

That is not hypothetical — it is what a live 3cc install did on 2026-08-25 (sky, image
`CUBE_3.1.20_20260825-1849_b0162f9`). The terraform apply failed at 19:34:48
(`Error: failed to create keycloak database via terraform`) after the database resource had been created,
and every subsequent apply logged "the database for keycloak is already created" while keycloak crash-looped
on `Access denied for user 'keycloak'`. Recovery required creating the users by hand.

The trigger there was the MariaDB 10.11 provider incompatibility (filed separately), but this check is a
defect on its own: **any** partial failure of that apply — interrupted commit, galera not yet writable,
provider error — becomes permanent and silent.

## Suggested fix

Make the check test what the function is responsible for: the database **and** both users (and ideally the
grants), e.g. `MysqlUtilIsDbExist(DB_NAME) && MysqlUtilIsUserExist("keycloak", "localhost") &&
MysqlUtilIsUserExist("keycloak", "%")`. Anything missing means re-run the setup rather than declare success.

Worth a sweep for the same shape elsewhere: an `isXSetup()` that asserts one artifact of a multi-artifact
operation is a silent-permanent-failure generator.

## How to Reproduce

1. On a healthy cluster: `mysql -e "drop user 'keycloak'@'localhost'; drop user 'keycloak'@'%';"`
   (leave the database in place — this is exactly the state a partial apply leaves).
2. `hex_config apply` (or `hex_cli -c cluster -c check_repair`).
3. `grep "already created" /var/log/hex_config.log` — the setup short-circuits; the users are still absent.
4. keycloak restarts into `Access denied for user 'keycloak'`.

## Expected behavior

`setupDatabase()` re-runs whenever any part of the keycloak DB setup is missing, and only reports success
when the database, both users and their grants exist.

## QA Verification

|---|---|---|
| 1 | Drop both keycloak users (DB intact), run `hex_config apply` | users and grants are recreated |
| 2 | `mysql -N -e "select user,host from mysql.user where user='keycloak'"` | two rows |
| 3 | Run `hex_config apply` again on the healthy cluster | no-op, no duplicate-user error, no needless terraform run |
| 4 | Fresh install | FTS completes; keycloak `Running` |

- [ ] **QA Status declared at Done**

- [x] **Handbook knowledge update** — extend the known-issue entry from the MariaDB/terraform bug with the
      idempotency lesson: check every artifact the setup creates, not just the first one.
      Landed in handbook [#394](https://github.com/bigstack-oss/bigstack-handbook/pull/394): the idempotency lesson is in
      [`known-issues/mariadb-1011-breaks-terraform-mysql-user.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/mariadb-1011-breaks-terraform-mysql-user.md)
      — an `isXSetup()` that asserts one artifact of a multi-artifact operation is a
      silent-permanent-failure generator. The shipped check counts `mysql.db` grants for both host
      patterns, not just the schema; verified idempotent on jim-1cc (re-run does no DB work, grants stay 2).

Eandalf-Bigstack:
#### What type of PR is this?

/kind bug

- **Unbreaks every fresh install off develop.** `48a9533 feat(mariadb): bump mariadb from 10.6.x to 10.11.x for openstack caracal` took the image to MariaDB **10.11.18**, which no longer has the `innodb_version` system variable. The archived `terraform-providers/mysql v1.9.0` probes it — `SELECT @@GLOBAL.innodb_version` — to pick a `CREATE USER` syntax, so on 10.11 **every user it should create fails** with `Error 1193: Unknown system variable 'innodb_version'` on both `mysql_user` resources. `mysql_database` does not probe it and succeeds, so the apply leaves a keycloak schema with no account able to reach it. keycloak then crash-loops on `Access denied for user 'keycloak'`, the SAML metadata gate spins with no deadline, and FTS never finishes — on every topology. 48a9533 retired the old `innodb_version` warning on the grounds that *"nothing in this tree or in the telegraf fork reads innodb_version"*, which is true of the **source**; the provider is a vendored binary, so the grep was a false negative.
- **Creates the keycloak database with SQL instead of that provider.** keycloak was the only service creating its database through terraform; manila, heat, octavia, ironic, masakari, skyline and the rest all issue two GRANTs. Swapping to the maintained `petoju/mysql` fork (it queries `@@GLOBAL.version`) was the alternative, but that adds a third-party binary to an air-gapped appliance's provider set and leaves keycloak the odd one out — removing the dependency beats replacing it.
- **Stops `isDatabaseSetup()` calling a half-created database done.** It checked only `MysqlUtilIsDbExist()`, so once the schema existed every later commit short-circuited `setupDatabase()` and the missing accounts were never created — which is what made a one-time apply failure permanent and silent. It now also counts one grant per host pattern `setupDatabase()` issues. `CREATE DATABASE` is `IF NOT EXISTS` to match, since the schema may already be there when only the grants are gone and re-creating it would discard keycloak's realms. The hard-remove path drops the accounts before the schema for the same reason.
- **Logs what terraform actually said when a module fails.** `ExecTerraform` passed `captureStdout=false`/`captureStderr=false`, so terraform's output went to `hex_config`'s inherited stdio and never reached syslog — a failed module left only `failed to create keycloak database via terraform`, with no cause, and this bug had to be reproduced by hand to diagnose. Both streams are now captured and the diagnostics logged one line at a time (syslog truncates a single long message), preferring stderr and falling back to stdout because terraform 0.14 uses both. `-no-color` comes along, since the output now comes from a pipe and 0.14 does not drop the ANSI escapes on its own.
- **Drops the terraform mysql module now that nothing targets it.** Removes the module, its `mysql_dbname` variable, and the `terraform-providers/*` entry from `override.tfrc`'s air-gap mirror — that include existed solely for this provider. Terraform itself stays; the six keycloak client modules and the rancher module are untouched.

Fixes #1344
Fixes #1345

> Note
>
> **Two details worth attention in `setupDatabase()`.** The account name and password are read from `/opt/keycloak/db` rather than written as literals — `createKeycloakDbSecrets()` builds the k3s secret `keycloak-db-secret` from those same files, so a literal would be a second source of truth for exactly the value that has to match. Statements go through `Cmd`/`ExecSync` as an **argv**, not a shell line, so neither value can be re-parsed as shell syntax, and the failure log deliberately omits the statement because it carries the password.

**Verified on `jim-1cc`**, on a node left in exactly the broken state — schema present, zero grants, the same state #1345 step 1 asks you to create by hand:

```
before   GRANTS=0   keycloak-keycloakx-0  0/1 CrashLoopBackOff  24 restarts, 105m
after    GRANTS=2   keycloak@localhost + keycloak@%
                    keycloak-keycloakx-0  1/1 Running
                    /etc/keycloak/saml-metadata.xml  3304 bytes
                    hex_config check_keycloak  rc=0
```

A second `hex_config install_keycloak` does no database work and leaves `GRANTS=2` — no duplicate-user error, no needless terraform run (#1345 QA rows 2 and 3). The new diagnostics logging was exercised by moving `keycloak-admin-password.tfvars` aside and forcing a keycloak update:

```
Error: terraform apply on module keycloak failed (exit 1)
Error: terraform: Error: Failed to read variables file
Error: terraform: Given variables file
Error: terraform: .../keycloak-admin-password.tfvars does not exist.
```

The old code logged nothing at all there. The success path is unaffected — `module.keycloak` still applies clean with `-no-color` and capture enabled. Compiles clean under `core/modules`' `-Wall -Werror` and links into `hex_config`.

**Not verified.** `make -C core/terraform` and `make -C core/cubectl test` were not run — no Go toolchain was available, and the `terraform init` + provider re-vendor into `patched-registry` is a build-time path I could not exercise. Nothing references the removed namespace so the narrowed `override.tfrc` include should resolve, but both want a real build before merge. `common_test.go`'s imports were checked by hand: `database/sql` was only used by the deleted test, everything else still has users, and `runMysqlContainer()` stays because `config_k3s_test.go` and `integration_test.go` use it.

**Coverage this removes.** `TestTerraformModuleMysql` goes with the module, and with it the `StatePushPull` subtest — the only coverage of the etcdv3 backend's state push/pull round-trip. It lived there because the mysql module was the one cheap fixture that could seed state; keycloak and rancher need live services. Reinstating it needs a fixture that does not exist yet, so this is a real gap rather than dead weight and probably wants its own issue. It costs less than it looks: the test ran `mariadb:10.3.27` against a product shipping 10.11.18, so **the breakage that took out FTS on every install would have passed here regardless**. `terraformPullState()` is now referenced only from commented-out code — legal Go, but a linter may flag it.

**Out of scope, worth its own issue.** While the SAML gate spins it holds `/var/run/hex_config.commit.lock`, so no other `hex_config` command can run at all — repair attempts just queue on `Acquring lock`. Recovering the node needed the documented `touch /run/cube_keycloak_saml_gate_release` escape hatch to free the lock first. A broken keycloak does not merely hang FTS, it locks out the commands an operator would reach for.

**Scope note.** This branch also carries `098b303c fix(docker): retry skopeo pulls so a transient blob read cannot fail the build`, which is unrelated to these two issues and has no PR of its own. Happy to split it out.

#### Additional documentation

```docs

```


Eandalf-Bigstack:
We have run what the PR left out:

```bash
[root@centos9-jail build]# make -C core/terraform
make: Entering directory '/home/jenkins/workspace/centos9_cubecos_jim_200.13/build/core/terraform'
  CLONE   tf-core (19s)
  CLEAN   tf-core (0s)
  UPDATE  tf-core (2s)
  BUILD   tf-core (2s)
  CLONE   rancher-provider (4s)
  UPDATE  rancher-provider (1s)
  BUILD   rancher-provider (4s)
  CLONE   keycloak-provider (2s)
  UPDATE  keycloak-provider (0s)
  BUILD   keycloak-provider (1s)
  RUN     etcd@jail (0s)
  INIT    terraform (9s)
make: Leaving directory '/home/jenkins/workspace/centos9_cubecos_jim_200.13/build/core/terraform'
[root@centos9-jail build]# make -C core/cubectl test
make: Entering directory '/home/jenkins/workspace/centos9_cubecos_jim_200.13/build/core/cubectl'
Trying to pull registry.access.redhat.com/ubi9/ubi-init:latest...
Getting image source signatures
Checking if image destination supports signatures
Copying blob 81ef194790b5 done   | 
Copying blob 56649e4df8ed done   | 
Copying config 9c85a10a8b done   | 
Writing manifest to image destination
Storing signatures
Error: adding content to container "ubi-init-working-container": checking on sources under "/": Rel: can't make  relative to /; copier: stat: ["/home/jenkins/workspace/centos9_cubecos_jim_200.13/cubecos/jail/centos8/*.repo"]: no such file or directory
643a9651850ded159404e00ec0efeeba3af962167100f83a9a2e72c0acdf660e
Error: removing container "ubi-init-working-container": container not known
make[1]: *** [Makefile:23: install] Error 125
make: *** [Makefile:26: test] Error 2
make: Leaving directory '/home/jenkins/workspace/centos9_cubecos_jim_200.13/build/core/cubectl'
```

Tests of cubectl were already broken. We could ignore it for now.

traviswu-bigstack (commented):
i feel we hit the same issue while i am trying to build image from scratch and do reimage. but the complexity of our fix suggested by ai looks quite different. can you take a look? https://github.com/bigstack-oss/cubecos/pull/1349/changes/711ad764fb64c2965e49ec0dd123301d1e075277

Eandalf-Bigstack:
They are basically the same after a glimpse. I also instructed Claude to purge the no longer needed Terraform provider mysql since we only used it to configure DB database, user account, and privileges, which is replaced by the changes made.

traviswu-bigstack:
got it. i will remove mine

traviswu-bigstack (approved):

PR #1353, opened 2026-08-26
feat(manila): upgrade manila from antelope to caracal

### LINKED ISSUE #638: [Story] [Parallel Group] Upgrade Manila from Antelope to Caracal
## Description

As a CubeCOS operator, I want Manila be in Caracal so that we could upgrade to Caracal.

## Acceptance Criteria

- [ ] Migrate configs
- [ ] Pass health check

## Story Points: 3

> | -- | -- | -- |

Eandalf-Bigstack:
#### What type of PR is this?

/kind feature

- **Upgrades Manila from OpenStack Antelope to Caracal.** `manila 16.3.0 -> 18.3.0`, out of `/opt/openstack-antelope` (python 3.10) and into `/opt/openstack-caracal` (python 3.11), following keystone, glance, cinder, nova with placement, and neutron. 18.3.0 is the last numeric 2024.1 revision, the same rule #1203 used to land on 16.3.0 — and here it is load-bearing for a second reason, below.

- **The config migration is a no-op, checked the way #1203 checked the yoga→antelope one.** antelope **790** options → caracal **822**. Three are removed — `database.mysql_enable_ndb`, `database.use_tpool`, `oslo_middleware.secure_proxy_ssl_header` — and `config_manila.cpp` writes none of them, nor do they appear in the live `/etc/manila/manila.conf`. One is newly deprecated for removal, `DEFAULT.use_eventlog`, also unset. All **84** keys of the live conf survive. Two groups can't be checked against the sample and were done by hand: the 12 `[generic]` backend keys (they only register when the share driver loads, which is why `manila.conf.def` is deliberately empty) were confirmed by importing 18.3.0's `share_manager_opts`, driver and `service_instance` option lists; and the 15 auth-plugin/privsep keys the generator never enumerates are unenumerated *identically* in both releases. The 35 added options are the 2024.1 share-backup feature, netapp additions and new `oslo_messaging_rabbit` settings — **none adopted**, per this issue's "ignore the new features" criterion.

- **`python-manilaclient` is installed into *both* venvs.** This is the one place manila cannot copy cinder. It owns `/usr/bin/manila` (which `hex_sdk` drives) *and* the `share` osc plugin entry point. An entry point is only visible to the interpreter it was installed under, and `core/heavyfs` still links `/usr/bin/openstack` at the antelope venv — so the plugin stays there or the four `$OPENSTACK share` call sites in `sdk_os.sh` break. Unlike volume/compute/image, `share` is **not** built into python-openstackclient; it is a separate plugin, which is why cinder could take its client to caracal wholesale. `/usr/bin/manila` talks HTTP, so it goes to caracal with the API it queries: 4.8.1 there, 4.4.2 left in antelope for the plugin.

- **`manila-ui` deliberately does not move**, and stays pinned at 9.0.1. `core/horizon` installs the dashboard into both venvs, but nothing serves the caracal copy — `openstack-dashboard.service`, `gunicorn-config.py` and the httpd reverse proxy all point at the antelope tree, and every dashboard plugin installs next to it. manila-ui follows horizon, not the manila service; 11.0.1 is the release to move to on the day horizon itself hops.

- **The local NFS export patch is dropped — upstream carries it.** `antelope_patch/` held one functional line, a `<world>` → `*` rewrite in `NFSHelper.get_host_list()`, plus a 627-line `.orig` kept for diffing. Upstream fixed it themselves and better, mapping per entry after the split rather than over the whole blob, so it cannot also rewrite a `<world>` occurring inside a path. It arrived via `82428a03d`, and **18.3.0 is the first tag containing it** — 18.0.x through 18.2.1 would still need the patch. Net `+7 / -1255`.

- **The privsep helper is pinned to the caracal venv.** Third venv boundary this tree has met it at; `config_nova.cpp` and `config_neutron.cpp` already carry the identical pin. `core/nova/nova.mk` keeps `/usr/bin/privsep-helper` pointed at *antelope* for masakari and cyborg, and a python 3.10 helper cannot serve a caracal manila — so the path is named explicitly and `/etc/sudoers.d/manila` authorises exactly it. A safety net rather than a live path: only the lvm and glusterfs drivers reach privsep and `enabled_share_backends` is rewritten to `generic` on every `Commit()`.

- **The service VM image is rebuilt**, the elements pin moves `5c2b484 -> 68a7e4be` (master, 2026-08-14), and the artifact moves to `app-image/caracal/`. Nothing in those seven commits reaches the six elements this image is built from; `bin/manila-image-create` changes only in ways a zero-argument invocation never sees. Confirmed the sshd workaround is still required — `manila-ssh`'s `fetch-public-ssh-keys.service` at master **still** carries `Type=oneshot`.

- **`debootstrap`'s wget is now bounded, in the recipe.** `fetch_url()` runs a bare `wget -U "debootstrap/$VERSION"` with no `--timeout`, no `--tries` and no hook to pass either, so one half-open connection hangs the build forever rather than failing it. Hit for real: a single `.deb` stalled 9 minutes with the log frozen and the process still in `S` state, while the same URL fetched in 1.5 s both directly and through squid. A stall indistinguishable from slow progress is the expensive part. DIB's `08-debootstrap` hook runs `sudo sh -c 'http_proxy= no_proxy= debootstrap …'`, so `WGETRC` does not survive — `/etc/wgetrc` is the only lever. Appended idempotently, keyed on `timeout` already being set.

Fixes #638

> Note
>
> **Two things are not verified, and one is where I would expect a surprise.**
>
> 1. **No real `make` through `rootfs_install` for `core/manila`.** The hop was landed *surgically* on the running jim-1cc node — pip into the caracal venv, relink `/usr/bin/*`, relocate the wheel's `etc/` files by hand — which is the established e2e loop but leaves this PR's `.mk` sequencing unexercised: the chroot pip installs, the `etc/` relocation from the caracal prefix, and the symlink pass. Everything *downstream* of those steps is verified on a live node; the build steps that produce them are not.
> 2. **`make -C core/cubectl test` / `make -C core/terraform`** are untouched by this PR, but the image build itself was only run in the jim jail, twice.

**`manila create` without `--share-type` is broken, and it is not ours.** It fails with *"There is no default share type available"* even though `default_share_type = tenant_share_type` is set, the type is public, and both clients report `is_default: YES`. The cause is a client bug: `manilaclient/v2/shell.py` validates the default with `_find_share_type(cs, "default")`, a name-or-ID lookup for a type literally *called* `default`, instead of the `/types/default` endpoint. Proven pre-existing — the code is byte-identical in 4.4.2 and 4.8.1, and the antelope client fails the same way against the same cluster:

| | 4.4.2 (antelope) | 4.8.1 (caracal) |
|---|---|---|
| `share_types.get("default")` | OK → `tenant_share_type` | OK → `tenant_share_type` |
| `find_resource(share_types, "default")` | **FAILED** | **FAILED** |

`osc/v2/share.py` gets it right (`share_types.get(share_type='default')`), so `openstack share create` resolves the default correctly. Nothing in this tree depends on the broken path — `sdk_os.sh` uses `$OPENSTACK share …`, and `$MANILA` is only used for `service-list`. Deliberately **not** patched: the `manila` CLI now prints its own deprecation notice.

**A migrate-guard scare that turned out to be a non-issue, recorded so nobody re-raises it.** `migrate_manila_db()` is gated on `$STATE_DIR/manila_db_migrated`, nothing deletes it, and caracal adds columns to *existing* tables — so on the hand-landed node `openstack-manila-scheduler` died on `(1054, "Unknown column 'services.disabled_reason' in 'SELECT'")` until `db sync` was run manually, applying six 2024.1 revisions. That looks like an upgrade blocker across all seven venv-hopped services. It is not: `STATE_DIR` is `/etc/appliance/state`, which `hex_support.mk` creates inside `$(ROOTDIR)`, so it ships **as part of the rootfs image**. A real upgrade is an A/B root-slot switch and boots with an empty state dir, so the guard falls through and `db sync` runs once. Only a hand-landed venv hop reproduces the failure. **Do not "fix" this by suffixing the stamp with a release.**

**The API's max microversion moves 2.78 → 2.85.**

#### Additional documentation

```docs
two phase boot... [yes]
cube services... [ready]
cluster data... [synced]
applying policy... [no]

[cc1 ~]# /usr/bin/manila-manage --version
18.3.0

[cc1 ~]# ps -eo pid,user,args | grep -E "manila-(api|scheduler|share)" | grep -v grep
 632157 manila   /opt/openstack-caracal/bin/python3.11 /usr/bin/manila-scheduler --config-file /etc/manila/manila.conf --logfile /var/log/manila/scheduler.log
 632158 manila   /opt/openstack-caracal/bin/python3.11 /usr/bin/manila-api --config-file /etc/manila/manila.conf --logfile /var/log/manila/api.log
 633171 manila   /opt/openstack-caracal/bin/python3.11 /usr/bin/manila-share --config-file /etc/manila/manila.conf --logfile /var/log/manila/share.log

[cc1 ~]# rpm -qa | grep -i manila | sort
                       <- empty; antelope still had openstack-manila-ui + python3-manilaclient

[cc1 ~]# systemctl is-active openstack-manila-api openstack-manila-scheduler openstack-manila-share
active
active
active

[cc1 ~]# hex_sdk health_manila_check; echo "rc=$?"
rc=0

[cc1 ~]# curl -s -o /dev/null -w "unauthenticated GET /v2 -> %{http_code}\n" http://10.1.0.1:8786/v2
unauthenticated GET /v2 -> 401

[cc1 ~]# manila service-list
+----+------------------+-------------+------+---------+-------+----------------------------+
| Id | Binary           | Host        | Zone | Status  | State | Updated_at                 |
+----+------------------+-------------+------+---------+-------+----------------------------+
| 1  | manila-scheduler | cc1         | nova | enabled | up    | 2026-08-26T06:20:27.724711 |
| 2  | manila-share     | cc1@generic | nova | enabled | up    | 2026-08-26T06:20:22.351400 |
+----+------------------+-------------+------+---------+-------+----------------------------+

[cc1 ~]# openstack share service list        <- the osc plugin, from the antelope venv
+----+------------------+-------------+------+---------+-------+----------------------------+
| ID | Binary           | Host        | Zone | Status  | State | Updated At                 |
+----+------------------+-------------+------+---------+-------+----------------------------+
|  1 | manila-scheduler | cc1         | nova | enabled | up    | 2026-08-26T06:20:27.724711 |
|  2 | manila-share     | cc1@generic | nova | enabled | up    | 2026-08-26T06:20:32.354494 |
+----+------------------+-------------+------+---------+-------+----------------------------+

[cc1 ~]# manila pool-list
+---------------------+------+---------+---------+
| Name                | Host | Backend | Pool    |
+---------------------+------+---------+---------+
| cc1@generic#GENERIC | cc1  | generic | GENERIC |
+---------------------+------+---------+---------+

The two client versions, one per venv:
[cc1 ~]# /opt/openstack-caracal/bin/pip show python-manilaclient | head -2
Name: python-manilaclient
Version: 4.8.1
[cc1 ~]# /opt/openstack-antelope/bin/pip show python-manilaclient | head -2
Name: python-manilaclient
Version: 4.4.2
[cc1 ~]# /opt/openstack-antelope/bin/pip show manila-ui | head -2
Name: manila-ui
Version: 9.0.1
[cc1 ~]# ls -d /opt/openstack-caracal/lib/python3.11/site-packages/manila_ui
ls: cannot access ...: No such file or directory      <- correct, manila-ui did not move

The database migration, six 2024.1 revisions:
[cc1 ~]# su -s /bin/sh -c "/usr/bin/manila-manage db sync" manila
INFO alembic.runtime.migration Running upgrade aebe2a413e13 -> 9afbe2df4945, add backup
INFO alembic.runtime.migration Running upgrade 9afbe2df4945 -> cb20f743ca7b, add_resource_locks
INFO alembic.runtime.migration Running upgrade cb20f743ca7b -> 2d708a9a3ba9, backup_change_availability_zone_to_availability_zone_id
INFO alembic.runtime.migration Running upgrade 2d708a9a3ba9 -> 99d328f0a3d2, add_disable_reason_to_service
INFO alembic.runtime.migration Running upgrade 99d328f0a3d2 -> 6e32091979e0, add mount_point_name to share_instances
INFO alembic.runtime.migration Running upgrade 6e32091979e0 -> 2f27d904214c, add_backup_type_column


Full share lifecycle on the rebuilt image (pin 68a7e4be):

[cc1 ~]# manila show manila-test-share | grep -E "status|host|share_server_id"
| status          | available                            |
| share_server_id | 1fcb4863-20ef-4ce6-9266-63dfc27281be |
| host            | cc1@generic#GENERIC                  |
                                            <- available in 105s

[cc1 ~]# manila share-server-list
| 1fcb4863-... | cc1@generic | active | manila-test-sharenet | ... |

[cc1 ~]# openstack server list --all-projects
generic_1fcb4863-20ef-4ce6-9266-63dfc27281be   ACTIVE   <- booted from the rebuilt image

[cc1 ~]# manila access-list manila-test-share
| ip | 192.168.99.0/24 | rw | active |
| ip | 0.0.0.0/0       | rw | active |

The <world> parse the dropped patch used to handle, on the live path with no patch present.
Inside the service instance:
manila@ubuntu:~$ sudo exportfs -v
/shares/share-f4da79cb-4fa4-4afe-b980-42c832e856f7
                192.168.99.0/24(sync,wdelay,hide,no_subtree_check,sec=sys,rw,secure,no_root_squash,no_all_squash)
/shares/share-f4da79cb-4fa4-4afe-b980-42c832e856f7
                <world>(sync,wdelay,hide,no_subtree_check,sec=sys,rw,secure,no_root_squash,no_all_squash)

Mount, read/write and unmount from a separate client VM on the tenant network:
manila@ubuntu:~$ grep PRETTY /etc/os-release
PRETTY_NAME="Ubuntu 22.04.5 LTS"
manila@ubuntu:~$ ip -4 -o addr show | awk '{print $2, $4}'
lo 127.0.0.1/8
ens3 192.168.99.36/24
manila@ubuntu:~$ sudo mount -t nfs 192.168.99.5:/shares/share-f4da79cb-4fa4-4afe-b980-42c832e856f7 /mnt/manila && echo MOUNT_OK
MOUNT_OK
manila@ubuntu:~$ mount | grep /mnt/manila
... type nfs4 (rw,relatime,vers=4.2,rsize=131072,wsize=131072,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,clientaddr=192.168.99.36,local_lock=none,addr=192.168.99.5)
manila@ubuntu:~$ sudo dd if=/dev/urandom of=/mnt/manila/blob.bin bs=1M count=16 conv=fsync
16777216 bytes (17 MB, 16 MiB) copied, 0.182404 s, 92.0 MB/s
manila@ubuntu:~$ cat /mnt/manila/testfile.txt
cubecos-manila-18.3.0-caracal-pin-68a7e4be
manila@ubuntu:~$ md5sum /mnt/manila/blob.bin
953508787d0a3d65af8f57330f2a3292  /mnt/manila/blob.bin
manila@ubuntu:~$ df -h /mnt/manila
192.168.99.5:/shares/share-f4da79cb-...  974M   16M  891M   2% /mnt/manila
manila@ubuntu:~$ sudo umount /mnt/manila && echo UMOUNT_OK
UMOUNT_OK

Teardown, and manila still healthy afterwards:
[cc1 ~]# manila list                     -> empty
[cc1 ~]# manila share-server-list        -> empty
[cc1 ~]# openstack server list --all-projects  -> empty
[cc1 ~]# hex_sdk health_manila_check; echo "rc=$?"
rc=0


Service image, and the base-OS refresh it delivers:

  manila-service-image_caracal.qcow2   997085184 bytes, 948 MiB on disk, 4.1 GiB virtual
  md5                                  cd6ce3053a37251aa8a0c5180d65b2d2
  manila-image-elements                68a7e4beccbe36ca6c3c601c85aca041ddd7f326

Verified after upload by mounting the NAS read-only in the jail the way core/extpacks/Makefile
does -- the uploaded file's own md5, its 33-byte bare-hash sidecar, and the build all agree,
so the extpack diff matches and skips the ~1 GB re-copy:

[jail]# md5sum $NAS/app-image/caracal/manila-service-image_caracal.qcow2
cd6ce3053a37251aa8a0c5180d65b2d2
[jail]# cat $NAS/app-image/caracal/manila-service-image_caracal.qcow2.md5
cd6ce3053a37251aa8a0c5180d65b2d2

Guest packages, against the antelope image being replaced:

| | old (antelope) | new (caracal) |
|---|---|---|
| release | Ubuntu 22.04.5 LTS | Ubuntu 22.04.5 LTS |
| kernel | 5.15.0-186-generic | 5.15.0-190-generic |
| openssl | 3.0.2-0ubuntu1.26 | 3.0.2-0ubuntu1.29 |
| openssh-server | 1:8.9p1-3ubuntu0.16 | 1:8.9p1-3ubuntu0.16 |
| nfs-kernel-server | 1:2.6.1-1ubuntu1.2 | 1:2.6.1-1ubuntu1.2 |
| samba-common | 2:4.15.13+dfsg-0ubuntu1.13 | 2:4.15.13+dfsg-0ubuntu1.13 |

Build gates:
[jail]# make -C core/modules all     -> CXX config_manila.o        (-Wall -Werror, clean)
[jail]# make -C core/main hex_config -> LD  hex_config             (clean)
```


SekiXu:
Architecture follows the settled skeleton — nothing to relitigate. I re-ran the config
migration comparison myself: 790 → 822, with `database.mysql_enable_ndb`,
`database.use_tpool` and `oslo_middleware.secure_proxy_ssl_header` removed, matching the
body. I also checked something the body doesn't — default-value drift among options
present in both samples. Only two moved, `DEFAULT.neutron_host_id` and
`oslo_messaging_rabbit.heartbeat_rate`, both commented-out lines, so no behavior change.
All three ACs on #638 are met.

**One thing to fix: `core/manila/Makefile:47-49` — the pin bump is a silent no-op on a warm
build dir.**

```make
$(BLDDIR)/manila-image-elements:
	$(Q)git clone ...
	$(Q)cd manila-image-elements && git checkout $(MANILA_IMAGE_ELEMENTS_COMMIT)
```

If the directory already exists the whole rule is skipped, `git checkout` never runs, and
the image is still built from `5c2b484`. The rule predates this PR, but this is the first
time the pin actually moves, and the NOTE at `:30-32` asks for a rebuild on bump — so it is
load-bearing now. Either key a stamp file on the SHA, or `git -C manila-image-elements
fetch && git checkout $(MANILA_IMAGE_ELEMENTS_COMMIT)` inside `manila_service_image`.

Nits:

1. `:38-43` says the `bin/manila-image-create` changes are "not in a way that reaches a
   zero-argument invocation". Not quite — upstream `3d998e578` changed the centos
   `package_list` from `qemu-kvm qemu-img kpartx debootstrap` to `qemu-img kpartx`, and
   `need_required_packages()` / `verify_dependencies()` run on every invocation regardless
   of arguments. Harmless in practice, since `:5-8` already lists debootstrap as a build-host
   prerequisite, but the claim itself needs a rewording.
2. `:83-84` writes to `/etc/wgetrc`, a permanent global change to the build host with no
   removal path. Worth trying `/root/.wgetrc` instead: under `env_reset`, sudo sets `HOME`
   to the target user's home, and wget reads `~/.wgetrc` at higher precedence than
   `/etc/wgetrc`. If you already tried that and it didn't work, please say why in the comment.
3. `manila_service_image` isn't declared `.PHONY` (`core/extpacks/Makefile:51` does it for
   `$(MANILA).md5`).
4. `manila.mk:10-17` justifies the second client copy entirely by the osc entry point, but
   `manila-ui==9.0.1` at `:104-111` also imports `manilaclient` at runtime. Same fix either
   way — just worth a line in the header so nobody drops it after the horizon hop thinking
   the osc plugin was the only consumer.

Also on #638: the three AC checkboxes are still unchecked, and the issue has no
`## Output artifacts (Definition of Done)` section — a merged handbook PR is required
before it can close.

traviswu-bigstack (approved):

Eandalf-Bigstack:
> Architecture follows the settled skeleton — nothing to relitigate. I re-ran the config migration comparison myself: 790 → 822, with `database.mysql_enable_ndb`, `database.use_tpool` and `oslo_middleware.secure_proxy_ssl_header` removed, matching the body. I also checked something the body doesn't — default-value drift among options present in both samples. Only two moved, `DEFAULT.neutron_host_id` and `oslo_messaging_rabbit.heartbeat_rate`, both commented-out lines, so no behavior change. All three ACs on #638 are met.
> 
> **One thing to fix: `core/manila/Makefile:47-49` — the pin bump is a silent no-op on a warm build dir.**
> 
> ```makefile
> $(BLDDIR)/manila-image-elements:
> 	$(Q)git clone ...
> 	$(Q)cd manila-image-elements && git checkout $(MANILA_IMAGE_ELEMENTS_COMMIT)
> ```
> 
> If the directory already exists the whole rule is skipped, `git checkout` never runs, and the image is still built from `5c2b484`. The rule predates this PR, but this is the first time the pin actually moves, and the NOTE at `:30-32` asks for a rebuild on bump — so it is load-bearing now. Either key a stamp file on the SHA, or `git -C manila-image-elements fetch && git checkout $(MANILA_IMAGE_ELEMENTS_COMMIT)` inside `manila_service_image`.
> 
> Nits:
> 
> 1. `:38-43` says the `bin/manila-image-create` changes are "not in a way that reaches a
>    zero-argument invocation". Not quite — upstream `3d998e578` changed the centos
>    `package_list` from `qemu-kvm qemu-img kpartx debootstrap` to `qemu-img kpartx`, and
>    `need_required_packages()` / `verify_dependencies()` run on every invocation regardless
>    of arguments. Harmless in practice, since `:5-8` already lists debootstrap as a build-host
>    prerequisite, but the claim itself needs a rewording.
> 2. `:83-84` writes to `/etc/wgetrc`, a permanent global change to the build host with no
>    removal path. Worth trying `/root/.wgetrc` instead: under `env_reset`, sudo sets `HOME`
>    to the target user's home, and wget reads `~/.wgetrc` at higher precedence than
>    `/etc/wgetrc`. If you already tried that and it didn't work, please say why in the comment.
> 3. `manila_service_image` isn't declared `.PHONY` (`core/extpacks/Makefile:51` does it for
>    `$(MANILA).md5`).
> 4. `manila.mk:10-17` justifies the second client copy entirely by the osc entry point, but
>    `manila-ui==9.0.1` at `:104-111` also imports `manilaclient` at runtime. Same fix either
>    way — just worth a line in the header so nobody drops it after the horizon hop thinking
>    the osc plugin was the only consumer.
> 
> Also on #638: the three AC checkboxes are still unchecked, and the issue has no `## Output artifacts (Definition of Done)` section — a merged handbook PR is required before it can close.

Great thanks for the detailed review, fixing now. For the issue template issue, the fields were filled out.

Issue #1269, opened 2026-08-15
[Bug] vCPU hot-add after live migration cannot be onlined in the guest


traviswu-bigstack:
## Description

On qemu-kvm 10.1.0 / libvirt 11.10 (EL9.8 guest), growing vCPUs on a domain that has previously live-migrated leaves the new CPUs present in guest sysfs but un-onlinable (`echo 1 > /sys/devices/system/cpu/cpuN/online` → EIO). Straight hotplug (no prior migration) onlines instantly; memory hotplug is unaffected in all cases; a guest reboot fully reconciles. Affects the CPU half of live-resize's auto-migrate path (reproduced on both hot-deploy and fresh-image validation).

## How to Reproduce

1. Boot an EL9 guest with vCPU hotplug headroom; live-migrate it between hosts
2. `virsh setvcpus <dom> <n> --live --config` (or nova live-resize)
3. Host shows `current live = n`; guest `nproc` unchanged; onlining returns EIO

## Expected behavior

Hot-added vCPUs online in the guest regardless of prior migrations.

## Definition of Done (DoD)

- [ ] Live-resize v1.1: grow vCPUs on the source host before the auto-migrate live migration (memory attaches fine on the destination)
- [ ] Reproduce with plain virsh and report upstream (QEMU) if confirmed
- [ ] Runbook notes the guest-reboot reconciliation until fixed
- [x] Handbook knowledge update: covered in the [live-resize blueprint findings](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/blueprints/nova-live-resize.md)

PR #1363, opened 2026-08-26
fix(nova): reject vCPU live-resize on a guest that has live-migrated

### LINKED ISSUE #1361: [Bug] Nested virt exposed to every guest breaks vCPU hotplug after live migration
> **Revised 2026-08-26** — the original description blamed a generic
> "guest cannot online hot-added vCPUs after live migration" limitation. Further
> testing disproved that. Corrected below; the earlier text is wrong, not just imprecise.

## Description

Every CubeCOS instance is handed hardware virtualization it did not ask for: the host
runs `kvm_intel nested=Y` and nova's host-model CPU passes `vmx` straight through, so
**every** EL guest loads `kvm_intel` whether or not it ever runs a nested VM.

`kvm_intel` caches a VMCS config when it loads. It then refuses to enable virtualization
on a vCPU hot-added after a live migration, which aborts the CPU bring-up:

```
ACPI: CPU8 has been hot-added
smpboot: Booting Node 0 Processor 8 APIC 0x8
kvm_intel: Inconsistent VMCS config on CPU 8
kvm: enabling virtualization on CPU8 failed
smpboot: CPU 8 is now offline
```

The operator sees a live-resize return `202` with the new flavor applied while the guest
never gains the CPUs — so the instance is billed and scheduled for vCPUs it does not have.

**A rolling upgrade arms this fleet-wide.** The drain live-migrates every instance rather
than restarting it, and the `<cpu>` block is carried across byte-identically (verified:
md5 `e95a04c3…` before and after a migration, 77 vmx features both sides). So after any
roll, every instance in the cluster is in the failing state until it reboots.

## What is NOT the cause

Ruled out by measurement, so nobody re-treads this:

- **Not a capability mismatch.** All 12 VMX config MSRs (`VMX_BASIC`, `PROCBASED`,
  `EPT_VPID_CAP`, the `TRUE_*` set, `FEAT_CTL`, `MISC`) are byte-identical between the
  guest's boot CPU and the hot-added CPU.
- **Not host heterogeneity.** sky141/142/143 expose byte-identical VMX MSRs.
- **Not a timing race.** A retry ~6 s later fails identically; only a module reload clears it.

The precise reason `kvm_intel`'s cached config fails its own comparison is **not
established** and is an upstream kernel question.

## How to Reproduce

1. Boot an EL9 guest on a live-resize-capable flavor (e.g. 2 vCPU / 4096 MB / 20 GB).
2. Live-resize CPU-only, e.g. 2 → 4 vCPU. Guest onlines them in ~2 s — works.
3. `openstack server migrate --live-migration --wait <vm>`.
4. Live-resize CPU-only again, e.g. 4 → 6 vCPU. Returns `202`, nova shows the new flavor.
5. In the guest: `online` is unchanged, `present` has grown, and
   `echo 1 | sudo tee /sys/devices/system/cpu/cpu4/online` → `Input/output error`.

## Expected behavior

Instances should not carry a kernel module they never use. Guests that do use nested virt
should keep working, and the operator should have a documented way back when a guest lands
in this state.

## Additional Context

**Remedy is a one-second module reload, not a reboot** — `modprobe -r kvm_intel &&
modprobe kvm_intel` in the guest, after which the CPUs online immediately (verified:
`nproc` 8 → 10). Only possible when no nested VMs are running (`modprobe -r` fails EBUSY
otherwise); nothing else in the guest uses the module. There is no `/proc` or sysfs way to
refresh the cache — all `kvm_intel` parameters are read-only except `vmentry_l1d_flush` and
`dump_invalid_vmcs`, and `vmcs_config` is computed once in `hardware_setup()`.

**Prevention:** with `kvm_intel` not loaded, hotplug after migration works with no
remediation at all — verified by unloading it, migrating, then hot-adding 2 vCPUs, which
onlined automatically with a clean dmesg.

Fix is therefore the new `nova.nested.virt.enabled` tuning (default false), masking vmx/svm
off via `cpu_model_extra_flags`. It is **not retroactive**: a domain keeps the CPU features
it booted with, so existing instances shed vmx only when they reboot. All 11 instances on
the sky lab are currently exposed; `hex_sdk os_instance_nested_virt_list` reports them.

Note the tuning is cluster-wide — nova has no per-flavor knob to add a single CPU feature —
and it flips AMD hosts too, which previously always added `svm`.

## QA Verification

|---|---|---|
| 1 | `hex_cli -c settings` (or `tuning_dump`) after upgrade | `nova.nested.virt.enabled` present, default `false` |
| 2 | Boot a new instance, then `virsh dumpxml <dom> \| grep -c "name='vmx'"` | `0` — no vmx exposed |
| 3 | In that guest: `lsmod \| grep -c ^kvm_intel` | `0` — module never loads |
| 4 | Live-migrate it, then hot-add vCPUs via live-resize | guest `nproc` matches the new flavor, `dmesg` free of `Inconsistent VMCS config` |
| 5 | Set the tuning `true`, commit, boot an instance | vmx present again in dumpxml; nested virt usable |
| 6 | `hex_sdk os_instance_nested_virt_list` on a cluster upgraded in place | lists pre-upgrade instances still exposing vmx; empty once they have all rebooted |
| 7 | On an affected guest: `modprobe -r kvm_intel && modprobe kvm_intel` then online the CPUs | CPUs online; `nproc` matches the flavor |
| 8 | `hex_cli -c cluster -c check` | all service groups `ok` |

      `QA Status` field to `Not needed` or `Ready to QA`.

- [x] **Handbook knowledge update** — land a `kb/cubecos/known-issues/` note (status `open`:
      the underlying `kvm_intel` behaviour is contained, not fixed) via
      `/bigstack-core:save-to-handbook`, with its `<note>.scenario.yaml` sidecar, linked
      from its References.
      Landed as [`kb/cubecos/known-issues/vcpu-hotplug-after-live-migration.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/vcpu-hotplug-after-live-migration.md).

traviswu-bigstack:
#### What type of PR is this?

/kind bug

A live-migrated guest cannot online hot-added vCPUs until it reboots (`present` grows, writing
`online` returns EIO). nova still reported success and swapped the flavor, so the instance was
billed and scheduled for vCPUs the guest never got.

The #1269 CPU-first ordering only covers an instance's **first** migration — it grows the vCPU
half on the source before migrating, which does nothing for a guest that had already migrated
earlier in its life. Confirmed on-cluster: the source-side grow fires exactly as designed
(`live-resize applied: 8->16 vCPU`, migration 3.1 s later) and the guest still ends at 8 of 16.

Track the condition in `system_metadata` (`cube_live_resize_migrated`, set at
`post_live_migration_at_destination`, cleared on hard reboot) and reject the vCPU half at the API
with an actionable message. Memory-only grows are unaffected.

Fixes #1361

The judgement call is **reject rather than partially deliver**. Growing memory while silently
dropping the CPU half would keep the billing/capacity lie; a 400 tells the operator that a hard
reboot is the price of more vCPUs, matching the contract the UI tip already states for
pre-feature instances. Happy to switch to "grow memory, report CPU skipped" if you'd rather.

This contains the symptom, it does not fix the underlying qemu/guest-kernel behaviour.

Validated on sky 3cc, hot-deployed to all three nodes:

| case | result |
|---|---|
| never-migrated, CPU grow 6 → 8 | 202, guest onlines `0-7` |
| migrate | flag set, online CPUs survive |
| migrated, CPU grow 8 → 12 | **400** with the new message |
| migrated, memory-only 4 → 8 GB | 202, `MemTotal` 8 GB, `nproc` unchanged |
| hard reboot, then CPU grow | 202, guest onlines |

#### Additional documentation

```docs
kb/cubecos/known-issues/ — vCPU hotplug after live migration (+ scenario sidecar)
```

### DIFF
--- a/core/nova/caracal_patch/compute/api.py.patch
+++ b/core/nova/caracal_patch/compute/api.py.patch
@@ -1,6 +1,6 @@
 --- compute/api.py
 +++ compute/api.py
-@@ -49,6 +49,7 @@ from nova.compute import utils as comput
+@@ -49,6 +49,7 @@
  from nova.compute.utils import wrap_instance_event
  from nova.compute import vm_states
  from nova import conductor
@@ -8,7 +8,7 @@
  import nova.conf
  from nova import context as nova_context
  from nova import crypto
-@@ -761,7 +762,8 @@ class API:
+@@ -761,7 +762,8 @@
  
          # Target disk is a volume. Don't check flavor disk size because it
          # doesn't make sense, and check min_disk against the volume size.
@@ -18,7 +18,7 @@
              # There are 2 possibilities here:
              #
              # 1. The target volume already exists but bdm.volume_size is not
-@@ -4160,9 +4162,55 @@ class API:
+@@ -4160,9 +4162,62 @@
  
      # TODO(stephenfin): This logic would be so much easier to grok if we
      # finally split resize and cold migration into separate code paths
@@ -48,6 +48,13 @@
 +        if new.memory_mb > old.memory_mb and new.memory_mb > max_mem:
 +            _bad(_("target memory exceeds this instance's hotplug "
 +                   "ceiling (%d MB)") % max_mem)
++        # a guest that has live-migrated cannot online hot-added vCPUs
++        # until it reboots (#1269); memory hotplug is unaffected
++        if (new.vcpus > old.vcpus and
++                instance.system_metadata.get(clr.SYSMETA_MIGRATED)):
++            _bad(_("this instance has been live-migrated since it last "
++                   "booted; its guest cannot online hot-added vCPUs until "
++                   "it is hard-rebooted (memory-only grows still work)"))
 +
 +    @check_instance_lock
 +    @check_instance_state(vm_state=[vm_states.ACTIVE], task_state=[None])
--- a/core/nova/caracal_patch/compute/manager.py.patch
+++ b/core/nova/caracal_patch/compute/manager.py.patch
@@ -1,6 +1,6 @@
 --- compute/manager.py
 +++ compute/manager.py
-@@ -69,6 +69,7 @@ from nova.compute import utils as comput
+@@ -69,6 +69,7 @@
  from nova.compute.utils import wrap_instance_event
  from nova.compute import vm_states
  from nova import conductor
@@ -8,7 +8,7 @@
  import nova.conf
  import nova.context
  from nova import exception
-@@ -1299,6 +1300,17 @@ class ComputeManager(manager.Manager):
+@@ -1299,6 +1300,17 @@
              # host. Abort ongoing migration if still running and reset state.
              self._reset_live_migration(context, instance)
  
@@ -26,7 +26,17 @@
          db_state = instance.power_state
          drv_state = self._get_power_state(instance)
          expect_running = (db_state == power_state.RUNNING and
-@@ -9058,6 +9070,73 @@ class ComputeManager(manager.Manager):
+@@ -4275,6 +4287,9 @@
+         else:
+             instance.task_state = task_states.REBOOT_PENDING_HARD
+             expected_states = task_states.hard_reboot_states
++            # cube live-resize: a hard reboot rebuilds the domain from fresh
++            # XML, so vCPU hotplug works again (#1269)
++            instance.system_metadata.pop(clr.SYSMETA_MIGRATED, None)
+ 
+         context = context.elevated()
+         LOG.info("Rebooting instance", instance=instance)
+@@ -9058,6 +9073,73 @@
                  self._set_instance_obj_error_state(instance,
                                                     clean_task_state=True)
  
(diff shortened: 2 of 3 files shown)


traviswu-bigstack:
Closing — this approach was wrong.

It rejected every vCPU live-resize on a migrated guest with a 400. Further testing showed the condition is neither intrinsic nor permanent:

- All 12 VMX config MSRs are byte-identical between the guest's boot CPU and the hot-added CPU, and across all three hosts — there is no capability mismatch.
- The remedy is a one-second guest-side `modprobe -r kvm_intel && modprobe kvm_intel`, not a hard reboot.
- With `kvm_intel` simply not loaded, hotplug after migration works with no remediation at all.

So a permanent API-level block was far too heavy, and it would also have blocked guests that never load the module. The real problem is that CubeCOS exposes `vmx` to every guest by default.

Superseded by the `nova.nested.virt.enabled` tuning — see #1361, which has been corrected, and the replacement PR.

Issue #1360, opened 2026-08-26
[Bug] All live migration fails: service users lack role:service on caracal


traviswu-bigstack:
## Description

Caracal's secure-RBAC defaults gate the service-to-service APIs on `role:service`,
with no `admin` fallback. CubeCOS only ever grants that role to **cinder** and **nova** —
the other 14 service users get `admin` only.

nova's `[neutron]` section authenticates as the **neutron** user, so every port-binding
call is refused and **all live migration fails cluster-wide**, stock instances included:

```
rule:create_port_binding is disallowed by policy
(rule:update_port and rule:update_port:binding:profile) is disallowed by policy
```

`neutron/conf/policies/port_bindings.py` → `create_port_binding` is `check_str=base.SERVICE`
(`rule:service_api` = `role:service`), no deprecated alias.

The gap is structural, not a one-off: `migrate_keystone()` created the `service` role at v2.4
and handed it to cinder + nova only, inside `if ! $OPENSTACK role show service`. That block can
never run again once the role exists, and the whole function short-circuits on its
`keystone_migrated` marker — so no already-installed cluster can pick up a new grant.

Affects live migration, and therefore also live-resize's auto-migrate path, rolling upgrades,
and host evacuation/drain.

## How to Reproduce

1. Install CubeCOS 3.1.20 (caracal) — verified on `CUBE_3.1.20_20260825-1849_b0162f9`, sky 3cc.
2. Boot any instance: `openstack server create vm1 --flavor <any> --image <any> --network private`
3. `openstack server migrate --live-migration --wait vm1`
4. Migration fails; `openstack server show vm1` reports
   `fault: rule:create_port_binding is disallowed by policy`, instance goes to ERROR.

## Expected behavior

Live migration succeeds. Every OpenStack service user in the `service` project holds
`role:service` in addition to `admin`, on fresh installs and on upgraded clusters alike.

## Additional Context

Current role assignment on an affected cluster:

| user | roles |
|---|---|
| cinder, nova | `admin`, `service` |
| barbican, cyborg, designate, glance, heat, ironic, ironic-inspector, masakari, monasca, **neutron**, octavia, placement, skyline, watcher | `admin` only |

Granting `service` to `neutron` + `placement` and restarting nova-compute restored live
migration immediately on sky — that hand-fix is what unblocked live-resize matrix items 6/7.

## QA Verification

| # | Command / steps | Expected outcome |
|---|---|---|
[...]

PR #1373, opened 2026-08-27
feat(octavia): upgrade octavia from antelope to caracal

### LINKED ISSUE #640: [Story] [Parallel Group] Upgrade Octavia from Antelope to Caracal
## Description

As a CubeCOS operator, I want Octavia be in Caracal so that we could upgrade to Caracal.

https://docs.openstack.org/releasenotes/octavia/2024.1.html

> You must update the amphora image to support the SR-IOV VIP feature.

## Acceptance Criteria

- [x] Migrate configs
- [x] Pass health check
- [x] Ignore the new features introduced in the new version

**Delivered in PR #1373** — `octavia 12.0.1 -> 14.0.2`, out of `/opt/openstack-antelope` (python 3.10) into `/opt/openstack-caracal` (python 3.11), the eighth service to make the hop.

- **Migrate configs** — 510 options -> 516. Four removed (`database.mysql_enable_ndb`, `neutron.{insecure,region_name,service_name}`), four newly deprecated (`DEFAULT.use_eventlog`, `neutron.{ca_certificates_file,endpoint,endpoint_type}`); `config_octavia.cpp` writes none of the eight and none appears in the live conf. All 59 live keys survive. The section list is unchanged at 36, so the conf `hex_config` generates does not move — proven by swapping the `.def` and diffing against a pre-hop backup (identical). The generator input was refreshed from upstream's own file at the tag, which had gained `namespace = oslo.middleware.sizelimit`.
- **Pass health check** — `hex_sdk health_octavia_check` rc=0, plus `os_octavia_cfg_ids_ok` and `os_octavia_hm_port_bound` both rc=0, all four units `active` with `NRestarts=0`, and `octavia-status upgrade check` reporting Success on all three checks (it previously crashed with `NoSuchOptError`).
- **Ignore the new features** — the 10 added options (rabbit queue-manager settings, two healthcheck middleware options, `oslo_middleware.max_request_body_size`, jobboard credentials, `vault.kv_path`) are all left unset. The amphora image *was* rebuilt, but as a version-matching requirement rather than to adopt SR-IOV VIP: the agent and the controllers speak a versioned REST API, so `OCTAVIA_VER` pins both.

**One defect found and fixed beyond the upgrade itself:** the ZooKeeper jobboard opened a new `kazoo` client per flow and leaked its session (188 swallowed `ConnectionLoss` and 94 ZK session expiries per LB lifecycle). Pre-existing — byte-identical in 12.0.1 — and fixed by backporting upstream master's shared-client change.

**One defect found and deliberately *not* handled here:** `loadbalancer delete --cascade` takes ~305 s because nova's `update_port` to clear the port's `device_id` is refused by neutron's secure RBAC. That is #1360 (`role:service` missing on the `neutron` user) and belongs there; Octavia's detach code is byte-identical between 12.0.1 and 14.0.2. The load balancer deletes *completely* and health stays green — it is only slow.

## Story Points: 3

> | -- | -- | -- |

## QA Verification

|---|---|---|
| 1 | `/usr/bin/octavia-api --version` | `14.0.2` |
| 2 | `ls -l /usr/bin/octavia-*` | all seven symlinks resolve into `/opt/openstack-caracal/bin/` |
| 3 | `ps -eo args \| grep octavia- \| grep -v grep` | every daemon runs under `/opt/openstack-caracal/bin/python3.11` |
| 4 | `rpm -qa \| grep -i octavia` | empty |
| 5 | `systemctl is-active octavia-api octavia-worker octavia-housekeeping octavia-health-manager` | `active` x4 |
| 6 | `for u in octavia-api octavia-worker octavia-housekeeping octavia-health-manager; do systemctl show -p NRestarts --value $u; done` | `0` x4 |
| 7 | `hex_sdk health_octavia_check; echo rc=$?` | `rc=0` |
| 8 | `hex_sdk os_octavia_cfg_ids_ok; echo rc=$?` and `hex_sdk os_octavia_hm_port_bound; echo rc=$?` | `rc=0` for both |
| 9 | `su -s /bin/sh -c "/usr/bin/octavia-db-manage current" octavia` | `db2a73e82626 (head)` |
| 10 | `/usr/bin/octavia-status upgrade check` | three checks, all `Result: Success` (no `NoSuchOptError`) |
| 11 | `curl -s -o /dev/null -w "%{http_code}\n" http://$(hex_sdk shared_id):9876/v2` | `401` |
| 12 | `curl -s http://$(hex_sdk shared_id):9876/ \| jq -r '.versions[-1].id'` | `v2.28` (Antelope served `v2.27`) |
| 13 | `md5sum /etc/octavia/policy.yaml` | `c53952746cfb39f5c66f97bbe3bcb263` |
| 14 | `ls -l /etc/octavia/octavia.conf /etc/octavia/policy.yaml` | both `-rw-r----- root octavia` |
| 15 | `ls -d /etc/octavia/conf.d/{common,octavia-api,octavia-worker,octavia-housekeeping,octavia-health-manager}` | all five present |
| 16 | `/opt/openstack-antelope/bin/pip list \| grep -i octavia` | `octavia-dashboard 11.0.1` and `python-octaviaclient 3.4.0`, and nothing else |
| 17 | `openstack loadbalancer provider list`, `openstack loadbalancer flavor list` | served; `amphora` provider present, four flavors (`DEFAULT`/`SMALL`/`MEDIUM`/`LARGE`) |
| 18 | Horizon → Project → Network → **Load Balancers** panel | loads, no `No policy rules for service` error |
| 19 | `openstack image list --tag amphora` | one `active` image |
| 20 | Create an LB on a tenant subnet, add an HTTP:80 listener, a `ROUND_ROBIN` pool and two web-server members; `curl http://<vip>/` six times from a VM on that subnet | LB reaches `provisioning_status ACTIVE` and `operating_status ONLINE`; responses alternate between the two backends |
| 21 | `openstack loadbalancer delete --cascade <lb>`, then `openstack loadbalancer list`, `openstack loadbalancer amphora list`, `openstack server list --all-projects`, `openstack port list -f value -c Name \| grep -c octavia-lb` | all empty / `0`; **note** it takes ~5 min until #1360 lands, which is expected and not a failure of this story |
| 22 | `grep -c "kazoo.exceptions.ConnectionLoss" /var/log/octavia/worker.log` after a full LB lifecycle | `0` new occurrences (the jobboard fix) |
| 23 | `hex_sdk health_octavia_check; echo rc=$?` after the teardown in #21 | `rc=0` |

- [x] **QA Status declared at Done** — when this issue reaches `Done`, set the board's

> Suggested: **Ready to QA**. Items 1–19 and 22–23 were verified on jim-1cc; item 20 was
> verified end to end against the rebuilt amphora image; item 21 completes but slowly, for
> #1360's reason. Not covered by this verification: three-node HA
> (`loadbalancer_topology = ACTIVE_STANDBY` and `controller_ip_port_list` are untested at
> that shape, as is the shared jobboard client across concurrent flows), non-admin RBAC
> (everything ran as admin, the one case `policy.yaml` does not change), and a real `make`
> through `rootfs_install` for `core/octavia` — the hop was landed surgically on the node,
> so the chroot pip installs, the symlink pass and the patch loop still want a build. The
> image target itself *was* run for real.

      beside the note.

**Landed in bigstack-handbook PR [#411](https://github.com/bigstack-oss/bigstack-handbook/pull/411)** — one Topic, two Known-issues, two scenario sidecars, plus cross-link wiring on the two existing Octavia entries:

| Artifact | Path | Why it exists |
|---|---|---|
| Topic | `kb/cubecos/architecture/octavia-caracal-runtime.md` | Eighth entry in the `caracal-runtime` series. The **kazoo co-tenancy finding** is the generalisable part: `kazoo` is a `taskflow[zookeeper]` extra Octavia never declared, satisfied under Antelope only because **monasca-common** dragged it into the shared venv. The disjointness test cannot catch this class of problem — it reports only what the resolver *adds* — so the note flags it for the hops still to come. Also records that Octavia is the **counter-example** to the wheel-`data_files` rule Glance's hop established (nothing under `etc/` reaches the venv), and that a slow cascade delete is #1360 with the byte-identical detach code to prove it. |
| Known-issue | `kb/cubecos/known-issues/octavia-jobboard-kazoo-client-per-flow.md` + `.scenario.yaml` | `job_board()` builds a fresh `KazooClient` per call and never closes it; sessions expire on kazoo's 10 s default and taskflow swallows the resulting `ConnectionLoss` with `quiet=True`. Records both negatives explicitly — not a Caracal regression (byte-identical in 12.0.1), and not the cause of the slow delete (305 s before *and* after the fix, which is how the two were separated). `status: fixed-in:3.2.0`. |
| Known-issue | `kb/cubecos/known-issues/amphora-image-build-stale-element-refs.md` + `.scenario.yaml` | The image build dies on `Failed to find reference to stable/2024.1` — two elements name a branch opendev deleted at EOM — and a tag cannot be substituted because DIB writes the ref under `refs/heads/`. Also folds in the two build-host preconditions (`debootstrap` no longer auto-installed; debootstrap's unbounded `wget`). `status: fixed-in:3.2.0`. |
| Wiring | `octavia-antelope-pip-packaging.md`, `octavia-amphorav2-jobboard-on-zookeeper.md` | The "no longer runs from that venv" banner the cinder/glance antelope entries already carry, and cross-references so the jobboard entry reaches both its known defect and the hop that moved its client. |

Both sidecars record verified facts only — commands run and outputs observed — with the raw pre-diagnosis operator view in `symptom` and the dug-for forensics in `evidence`, per the sidecar's symptom rule.

Eandalf-Bigstack:
#### What type of PR is this?

/kind feature

- **Upgrades Octavia from OpenStack Antelope to Caracal.** `octavia 12.0.1 -> 14.0.2`, out of `/opt/openstack-antelope` (python 3.10) and into `/opt/openstack-caracal` (python 3.11), following keystone, glance, cinder, nova with placement, neutron and manila. 14.0.2 is the last numeric 2024.1 revision, the same rule #1206 used to land on 12.0.1 — and here it is load-bearing twice, because `core/octavia/Makefile` builds the amphora image from that same tag so the agent inside the guest and the controllers outside it cannot drift.

- **The config migration is a no-op, checked the way #1206 checked the yoga→antelope one.** antelope **510** options → caracal **516**. Four are removed — `database.mysql_enable_ndb`, `neutron.insecure`, `neutron.region_name`, `neutron.service_name` — and `config_octavia.cpp` writes none of them, nor do they appear in the live `/etc/octavia/octavia.conf` (its `[neutron]` section is empty). Four are newly deprecated for removal (`DEFAULT.use_eventlog`, `neutron.{ca_certificates_file,endpoint,endpoint_type}`), also unset. All **59** keys of the live conf survive; the 8 the generator never enumerates (6 `keystone_authtoken` auth-plugin keys, 2 `service_auth`) are unenumerated *identically* in both releases. The 10 added options are new `oslo_messaging_rabbit` queue-manager settings, two healthcheck middleware options, `oslo_middleware.max_request_body_size`, jobboard credentials and `vault.kv_path` — **none adopted**, per this issue's "ignore the new features" criterion. The section list is unchanged at 36, which is what matters for `octavia.conf.def`: `LoadConfig()` takes only section names from it, so the conf `hex_config` generates does not move — confirmed by swapping the `.def` on the node and diffing against the pre-hop backup.

- **The generator input had drifted and is refreshed from upstream.** Our copy was a verbatim byte-copy of upstream 12.0.1's `etc/config/octavia-config-generator.conf`; 2024.1 adds `namespace = oslo.middleware.sizelimit`, without which `[oslo_middleware] max_request_body_size` never reaches the sample. The two files are byte-identical again.

- **`kazoo` is now named explicitly, and finding out why is the most interesting thing in this PR.** `config_octavia.cpp` writes `task_flow/jobboard_backend_driver = zookeeper_taskflow_driver`, and taskflow's zookeeper jobboard imports `kazoo` — but that is an *extra* (`taskflow[zookeeper]`), not a requirement, and octavia has never declared it. Under antelope it was in the venv anyway, dragged there by **monasca-common**. The caracal venv has no monasca: `octavia-worker` crash-looped on `ModuleNotFoundError: No module named 'kazoo'`, and `octavia-status upgrade check` reported `AmphoraV2 Check: Failure — Failed to connect to jobboard backend`. This was a latent dependency on venv co-tenancy, and splitting the venv is what exposed it — worth stating because the same shape may be waiting in the components still to hop.

- **`python-octaviaclient` stays in the *antelope* venv, and alone.** An osc plugin entry point is only visible to the interpreter it was installed under, and `core/heavyfs` still links `/usr/bin/openstack` at the antelope venv — so the `loadbalancer` subcommand has to be installed there or `sdk_os.sh`'s call sites (`loadbalancer list`, the four flavorprofile/flavor pairs at bootstrap, `delete --cascade`) stop resolving. Unlike manila there is no second consumer: octavia ships no standalone CLI and nothing in `hex_sdk` runs one, so it is installed once rather than into both venvs. That leaves the antelope constraint's 3.4.0 client talking to a 14.0.2 API, which is the supported direction and was exercised on the node.

- **`octavia-dashboard` deliberately does not move**, and stays pinned at 11.0.1. It follows horizon, not the octavia service — `openstack-dashboard.service`, `gunicorn-config.py` and the httpd reverse proxy all point at the antelope tree, and every dashboard plugin installs next to it. It talks to the API over HTTP and imports nothing from octavia, so the split costs nothing; 13.0.1 is the release to move to on the day horizon hops. Same reasoning manila-ui got in #1353.

- **The horizon policy namespace follows the *service*, not the panel.** `octavia` moves from `HORIZON_POLICY_NS_ANTELOPE` to `..._CARACAL` because the `oslo.policy.policies` entry point named `octavia` is registered by the octavia package. Without it the build fails outright, not silently: the antelope loop's `|| exit 1` fires on `The requested namespace "octavia" is not found.` Rule *names* are identical across the hop, 95 → 95, so the panel's effective RBAC surface does not move.

- **Nothing octavia carries can be replaced by the wheel.** Unlike glance's metadefs, manila's `api-paste.ini` or cinder's rootwrap filters, octavia's `data_files` are only `share/octavia/{LICENSE,README.rst,diskimage-create/*}` — **nothing under `etc/`** reaches the venv prefix. Each carried file was still checked against the 14.0.2 sdist: `policy.yaml` is byte-identical to upstream's `etc/policy/admin_or_owner-policy.yaml` (md5 `c53952746cfb39f5c66f97bbe3bcb263`, unchanged from 12.0.1), and `dhclient.conf` (`timeout 60;`), `certs/octavia-certs.cnf` (tabs→spaces, `RANDFILE` commented out for openssl 3) and `octavia-dist.conf` still differ from upstream's for the reasons they always did. All four `octavia-dist.conf` options still exist in caracal.

- **The two carried python patches are rebased, and one changes form.** `nova_driver.py`'s `meta={'HA_Enabled': 'False'}` hunk applies unchanged — upstream reworked the surrounding function for 2024.1 (new `BadRequest`/`ClientException` handlers for PCI-claim and port-binding failures, log call moved onto `constants`) but none of it reaches the `create_server()` kwargs. `cmd/status.py` was a **whole-file copy** and is now a `.py.patch`: it had to be refreshed either way — 14.0.2 drops the `import sys` and `__main__` guard 12.0.1 had, so the old copy would have quietly reverted both — and that is exactly the drift a whole-file copy cannot show you. The `octavia.common.policy` import is still missing upstream at 14.0.2 and on master.

- **The amphora image is rebuilt, and `core/octavia/Makefile` had four separate defects that all had to be fixed before it could be.** (1) The clone was keyed on a *directory*, so on any warm build dir the rule was skipped and a bumped pin silently rebuilt the old source — same defect #638 hit in core/manila, fixed the same way with a stamp named after the pin. (2) The pin was a *branch*; `stable/2024.1` is already deleted, exactly as `stable/2023.1` was. It is now the release tag, read out of `octavia.mk` so `OCTAVIA_VER` is the single pin. (3) Every path was relative to wherever make was standing (`BLDDIR` is `$(shell pwd)`), so `make -C core` and `make -C core/octavia` wrote to different places — all of it is now anchored at `$(TOP_BLDDIR)/core/octavia`, deliberately **not** `$(COREDIR)`, which is the source tree. (4) Two build-host preconditions were undocumented — `debootstrap` (upstream dropped it from `need_required_packages()` for rhel/centos; hit for real) and debootstrap's unbounded `wget`, both already written down in core/manila's image target and neither manila-specific.

- **Upstream's amphora elements are broken at 14.0.2 and need a ref override.** `elements/amphora-agent/source-repository-amphora-agent` still names `stable/2024.1`, a branch opendev deleted at EOM, so the build dies mid-run with `Failed to find reference to stable/2024.1` — a string upstream never refreshed, and one that does not point back at the recipe. `elements/octavia-lib` carries the identical stale ref and is in the same sequence. Both are overridden with `DIB_REPOREF_<repo>` and with a **commit, not a tag**: DIB's `98-source-repositories` fetches the ref into `refs/heads/`, where an annotated tag cannot go (`fetching ref refs/heads/14.0.2 failed: invalid new value provided`). Set in the environment rather than with `diskimage-create.sh`'s `-g`, because `-g` also rewrites the upper-constraints URL to a *branch* path.

- **`fix(octavia): share one zookeeper client across jobboard flows`** — a backport of upstream's fix on master. `ZookeeperTaskFlowDriver.job_board()` built a brand-new `KazooClient` on every call and never closed it; each abandoned client holds a ZK session until zookeeper expires it on kazoo's 10 s default (octavia passes no `timeout`, and `jobboard_expiration_time` applies only to the redis driver, so there is no knob for it). Sessions therefore accumulate and expire continuously, and taskflow's `_process_child` catches the resulting `ConnectionLoss` in its generic `KazooException` branch which — called with `quiet=True` — logs and **discards** it. Measured over one create-plus-cascade-delete: **188 → 0** new `ConnectionLoss`, **94 → 0** new ZK `Expiring session`, worker ZK connections climbing → stable at 3. **Not a caracal regression** — `ZookeeperTaskFlowDriver` is byte-identical in 12.0.1, so antelope leaks the same way; carried here because this is the release we are landing and the fix is small and upstream.

Fixes #640

> [!NOTE]
> **The ~5 minute `loadbalancer delete --cascade` is #1360, not this PR.** It is worth reading the chain because it looks like an octavia regression and is not one:
>
> ```
> 13:48:08  nova  Successfully detached device tapada8612f-8f ... from the persistent domain config
> 13:48:28  nova  Successfully detached device tapada8612f-8f ... from the live domain config
> 13:48:28  nova  Successfully unplugged vif VIFOpenVSwitch(...)
> 13:48:28  ERROR nova.network.neutron  Unable to clear device ID for port 'ada8612f-...':
>           nova.exception.Forbidden: ((((rule:update_port and rule:update_port:device_id)
>           and rule:update_port:device_owner) and rule:update_port:binding:host_id)
>           and rule:update_port:binding:profile) is disallowed by policy
> ```
>
> Nova completes the hot-unplug in 21 s. The `update_port` that clears `device_id` is refused because nova authenticates to neutron as the **neutron** user, which holds `admin` only — #1360's exact mechanism, quoting the same `rule:update_port ... binding:profile` family. The port keeps `device_id = <amphora compute id>`, so octavia's detach-wait — which polls `network_proxy.ports(device_id=compute_id)` — never sees it detach, gives up at its own `[networking] port_detach_timeout = 300` default (we write no `[networking]` values), then deletes the amphora, which cleans up in 0.54 s. Role table on the test node matches #1360 exactly: `neutron`, `octavia`, `placement` = `admin` only; `nova`, `cinder` = `admin service`.
>
> Octavia's `detach_port` and detach-wait loop are **byte-identical** between 12.0.1 and 14.0.2, so nothing here changed it. **Deliberately not handled in this PR** — it will be fixed by #1360. The load balancer deletes *completely* and health stays green; it is only slow.
>
> The end-to-end LB test passed despite #1360, and that is not luck: `config_octavia.cpp` sets `[service_auth] username = admin_cli`, so octavia's own outbound calls go as an admin user and never hit the gap. Only nova's neutron-authenticated call does.

> [!NOTE]
> **Two things are not verified.**
>
> 1. **No real `make` through `rootfs_install` for `core/octavia`.** The hop was landed *surgically* on the running jim-1cc node — pip into the caracal venv, relink `/usr/bin/*`, apply the patches by hand, swap the `.def` — which is the established e2e loop but leaves this PR's `.mk` sequencing unexercised: the chroot pip installs, the symlink pass and the patch loop. Everything *downstream* of those steps is verified on a live node. The image target itself **was** run for real — four invocations, the first three of which surfaced the defects fixed here (missing `debootstrap`, then each stale element ref in turn) before the fourth produced the artifact.
> 2. **Single-node only, and the jobboard change is the one I would watch.** `loadbalancer_topology = ACTIVE_STANDBY` and `controller_ip_port_list` are untested at three-node shape, and a *shared* kazoo client across concurrent flows on three controllers is untested here. Non-admin RBAC is also untested — everything ran as admin, the one case `policy.yaml` does not change.

**The API's max microversion moves v2.27 → v2.28.**

**Two pre-existing defects from #1206 remain open and are untouched here:** `[health_manager] bind_ip` hard-coded to `172.16.0.<octet4>` (now moot on this node — heartbeats arrive and LBs report `ONLINE`, so e2cfce69 addressed it), and `octavia-status upgrade check` crashing for want of an `oslo_policy` import — that one *is* fixed, by the `cmd/status.py` patch this PR rebases.

#### Additional documentation

```docs
[cc1 ~]# /usr/bin/octavia-api --version
%prog 14.0.2

[cc1 ~]# /opt/openstack-caracal/bin/pip show octavia octavia-lib kazoo | grep -E "^Name|^Version"
Name: octavia
Version: 14.0.2
Name: octavia-lib
Version: 3.5.0
Name: kazoo
Version: 2.10.0

[cc1 ~]# ps -eo pid,user,args | grep -E "octavia-(api|worker|housekeeping|health-manager)" | grep -v grep
   6406 octavia  /opt/openstack-caracal/bin/python3.11 /usr/bin/octavia-api --config-file /usr/share/octavia/octavia-dist.conf --config-file /etc/octavia/octavia.conf --config-dir /etc/octavia/conf.d/common --config-dir /etc/octavia/conf.d/octavia-api --log-file /var/log/octavia/api.log
   6410 octavia  /opt/openstack-caracal/bin/python3.11 /usr/bin/octavia-housekeeping --config-file ... --log-file /var/log/octavia/housekeeping.log
   6881 octavia  /opt/openstack-caracal/bin/python3.11 /usr/bin/octavia-health-manager --config-file ... --log-file /var/log/octavia/health-manager.log
   6883 octavia  octavia-worker: master process [/usr/bin/octavia-worker --config-file ... --log-file /var/log/octavia/worker.log]

[cc1 ~]# rpm -qa | grep -i octavia | sort
                       <- empty

[cc1 ~]# systemctl is-active octavia-api octavia-worker octavia-housekeeping octavia-health-manager
active
active
active
active
[cc1 ~]# for u in octavia-api octavia-worker octavia-housekeeping octavia-health-manager; do systemctl show -p NRestarts --value $u; done
0
0
0
0

[cc1 ~]# hex_sdk health_octavia_check; echo "rc=$?"
rc=0
[cc1 ~]# hex_sdk os_octavia_cfg_ids_ok; echo "rc=$?"
rc=0
[cc1 ~]# hex_sdk os_octavia_hm_port_bound; echo "rc=$?"
rc=0

[cc1 ~]# curl -s -o /dev/null -w "unauthenticated GET /v2 -> %{http_code}\n" http://10.1.0.1:9876/v2
unauthenticated GET /v2 -> 401

[cc1 ~]# curl -s http://10.1.0.1:9876/ | python3 -c "import sys,json;d=json.load(sys.stdin);vs=[v['id'] for v in d['versions']];print(vs[0],'..',vs[-1],'(',len(vs),'versions )')"
v2.0 .. v2.28 ( 29 versions )        <- antelope served .. v2.27

The database migration, the two 2024.1 revisions:
[cc1 ~]# su -s /bin/sh -c "/usr/bin/octavia-db-manage upgrade head" octavia
INFO alembic.runtime.migration Running upgrade 0995c26fc506 -> 632152d2d32e, Add HTTP Strict Transport Security support
INFO alembic.runtime.migration Running upgrade 632152d2d32e -> db2a73e82626, Add vnic_type for VIP
[cc1 ~]# su -s /bin/sh -c "/usr/bin/octavia-db-manage current" octavia
db2a73e82626 (head)

octavia-status now runs at all -- unpatched it dies with NoSuchOptError before printing:
[cc1 ~]# /usr/bin/octavia-status upgrade check
+-----------------------------------------------------------------------+
| Check: AmphoraV2 Check                                                |
| Result: Success                                                       |
| Details: Persistence database and Jobboard backend for AmphoraV2      |
|   provider configured.                                                |
+-----------------------------------------------------------------------+
| Check: YAML Policy File                                               |
| Result: Success                                                       |
+-----------------------------------------------------------------------+
| Check: Policy File JSON to YAML Migration                             |
| Result: Success                                                       |
+-----------------------------------------------------------------------+

Config is untouched by the .def swap, and policy.yaml is still upstream's:
[cc1 ~]# diff /etc/octavia/octavia.conf /etc/octavia/octavia.conf.bak-pre640-20260827-1228 && echo IDENTICAL
IDENTICAL
[cc1 ~]# md5sum /etc/octavia/policy.yaml
c53952746cfb39f5c66f97bbe3bcb263  /etc/octavia/policy.yaml
[cc1 ~]# ls -l /etc/octavia/octavia.conf /etc/octavia/policy.yaml
-rw-r----- 1 root octavia 2590 /etc/octavia/octavia.conf
-rw-r----- 1 root octavia  769 /etc/octavia/policy.yaml

The venv split, per component:
[cc1 ~]# /opt/openstack-antelope/bin/pip list | grep -i octavia
octavia-dashboard            11.0.1      <- follows horizon, deliberately not moved
python-octaviaclient         3.4.0       <- the "loadbalancer" osc plugin
[cc1 ~]# ls -l /usr/bin/octavia-* | awk '{print $9, $10, $11}'
/usr/bin/octavia-api -> /opt/openstack-caracal/bin/octavia-api
/usr/bin/octavia-db-manage -> /opt/openstack-caracal/bin/octavia-db-manage
/usr/bin/octavia-driver-agent -> /opt/openstack-caracal/bin/octavia-driver-agent
/usr/bin/octavia-health-manager -> /opt/openstack-caracal/bin/octavia-health-manager
/usr/bin/octavia-housekeeping -> /opt/openstack-caracal/bin/octavia-housekeeping
/usr/bin/octavia-status -> /opt/openstack-caracal/bin/octavia-status
/usr/bin/octavia-worker -> /opt/openstack-caracal/bin/octavia-worker

The 3.4.0 plugin against the 14.0.2 API -- every sdk_os.sh call site:
[cc1 ~]# openstack loadbalancer provider list
+---------+-------------------------------------------------+
| amphora | The Octavia Amphora driver.                     |
| octavia | Deprecated alias of the Octavia Amphora driver. |
+---------+-------------------------------------------------+
[cc1 ~]# openstack loadbalancer flavorprofile list   (4 rows: amphora-{default,small,medium,large})
[cc1 ~]# openstack loadbalancer flavor list          (4 rows: DEFAULT, SMALL, MEDIUM, LARGE)
[cc1 ~]# openstack loadbalancer list                 -> empty, served

The horizon policy namespace move, proven both ways -- exactly what horizon.mk runs:
[cc1 ~]# env DJANGO_SETTINGS_MODULE=openstack_dashboard.settings \
           /opt/openstack-caracal/bin/django-admin dump_default_policies \
           --skip-checks --namespace octavia --output-file /tmp/octavia.yaml; echo rc=$?
rc=0                                  <- 95 rules
[cc1 ~]# env DJANGO_SETTINGS_MODULE=openstack_dashboard.settings \
           /opt/openstack-antelope/bin/django-admin dump_default_policies \
           --skip-checks --namespace octavia --output-file /tmp/x.yaml; echo rc=$?
ERROR ... The requested namespace "octavia" is not found.
rc=1                                  <- with || exit 1, this fails the build
rule names before vs after: 95 -> 95, none added, none removed


FULL LOAD BALANCER LIFECYCLE, on the rebuilt caracal amphora image

[cc1 ~]# openstack loadbalancer create --name octavia-test-lb --vip-subnet-id octavia-test-subnet
| provisioning_status | ACTIVE          |     <- ACTIVE in 58-63s across four runs
| operating_status    | ONLINE          |     <- heartbeats arriving (e2cfce69)
| vip_address         | 192.168.88.130  |
| provider            | amphora         |

[cc1 ~]# openstack loadbalancer amphora list
| 3cf7bcc4-... | ALLOCATED | STANDALONE | lb_network_ip 10.254.0.151 | ha_ip 192.168.88.130 |

[cc1 ~]# openstack server list --all-projects
| amphora-3cf7bcc4-...  | ACTIVE | lb-mgmt-net=10.254.0.151; octavia-test-net=192.168.88.193 | amphora-x64-haproxy |
| backend2              | ACTIVE | octavia-test-net=192.168.88.13                            | Cirros              |
| backend1              | ACTIVE | octavia-test-net=192.168.88.206                           | Cirros              |

[cc1 ~]# openstack loadbalancer member list octavia-test-pool
| backend1 | 192.168.88.206 | 80 | ACTIVE | NO_MONITOR | weight 1 |
| backend2 | 192.168.88.13  | 80 | ACTIVE | NO_MONITOR | weight 1 |

Six requests through the VIP from a separate tenant VM, read out of its console log
(no FIP or router involved -- the client curls from userdata and prints to console):
===== CUBECOS OCTAVIA VIP TEST BEGIN =====
client ip: 192.168.88.161/24
request 1 -> backend backend2 192.168.88.13
request 2 -> backend backend1 192.168.88.206
request 3 -> backend backend2 192.168.88.13
request 4 -> backend backend1 192.168.88.206
request 5 -> backend backend2 192.168.88.13
request 6 -> backend backend1 192.168.88.206
===== CUBECOS OCTAVIA VIP TEST END =====
                                          <- clean ROUND_ROBIN alternation

Teardown, complete but slow (the 300s wait is #1360, see the note above):
[cc1 ~]# openstack loadbalancer delete --cascade octavia-test-lb
[cc1 ~]# openstack loadbalancer list        -> empty
[cc1 ~]# openstack loadbalancer amphora list -> empty
[cc1 ~]# openstack server list --all-projects -> empty
[cc1 ~]# openstack port show <vip_port_id>   -> vip port gone
[cc1 ~]# openstack port list -f value -c Name | grep -c octavia-lb
0
[cc1 ~]# hex_sdk health_octavia_check; echo "rc=$?"
rc=0


THE JOBBOARD FIX, measured over one create-plus-cascade-delete cycle

                                  before      after
  new kazoo ConnectionLoss         188          0
  new ZK "Expiring session"         94          0
  worker ZK connections         climbing   stable at 3
  LB create                        63 s       58 s
  cascade delete                  305 s      305 s     <- unchanged; that is #1360

[cc1 ~]# grep -c "kazoo.exceptions.ConnectionLoss" /var/log/octavia/worker.log   # before fix
570
[zk]# grep "Expiring session" /var/log/zookeeper/zookeeper.log | tail -1
Expiring session 0x10000010e7158b4, timeout of 10000ms exceeded    <- kazoo's 10s default


THE AMPHORA IMAGE -- read from the shipping qcow2, not inferred from the recipe

  amphora-x64-haproxy_caracal.qcow2   360688128 bytes, 343 MiB on disk, 2 GiB virtual
  md5                                 16396a57829059bf28e26b36dc1b19ae
  .md5 sidecar                        33 bytes, bare hash

[jail]# qemu-nbd --connect=/dev/nbd0 --read-only amphora-x64-haproxy_caracal.qcow2
[jail]# cat /mnt/amp/opt/amphora-agent.gitref
041b0795972afe103ea802b47105f38ad45f65cb        <- octavia 14.0.2, peeled commit
[jail]# cat /mnt/amp/opt/octavia-lib.gitref
55049e8415a147a333870b02910e05633d022513        <- octavia-lib 3.5.0, peeled commit
[jail]# ls /mnt/amp/opt/amphora-agent-venv/lib/python3.10/site-packages/ | grep dist-info
octavia-14.0.2.dist-info
octavia_lib-3.5.0.dist-info

Exact releases, not .dev builds -- which is what pinning to commits buys. #1206's
antelope image carried 12.0.2.dev10 because it built off a branch head.

The base OS does not move, and it is worth saying because the component looks like it
should: still ubuntu-minimal/jammy, the upstream default for 2024.1 and the same base
the antelope image used.

  release          Ubuntu 22.04.5 LTS
  kernel           5.15.0-190-generic
  haproxy          2.4.30-0ubuntu0.22.04.2
  keepalived       1:2.2.4-0.2build1
  openssl          3.0.2-0ubuntu1.29
  openssh-server   1:8.9p1-3ubuntu0.16

Verified after upload by mounting the NAS read-only in the jail the way
core/extpacks/Makefile does -- the uploaded file's own md5 read over NFS, its 33-byte
bare-hash sidecar, and the build all agree, so the extpack diff matches and skips the
~344 MB re-copy:

[jail]# mount -o ro 10.32.0.200:/volume1/docker/minio/downloads/ $B/nas
[jail]# md5sum $B/nas/app-image/caracal/amphora-x64-haproxy_caracal.qcow2
16396a57829059bf28e26b36dc1b19ae
[jail]# cat $B/nas/app-image/caracal/amphora-x64-haproxy_caracal.qcow2.md5
16396a57829059bf28e26b36dc1b19ae
[jail]# diff $B/core/octavia/amphora-x64-haproxy_caracal.qcow2.md5 \
             $B/nas/app-image/caracal/amphora-x64-haproxy_caracal.qcow2.md5
IDENTICAL -- extpack diff matches, the ~344 MB re-copy is skipped
[jail]# umount $B/nas

Path invariance for the rebuilt Makefile -- make -n from both directories:
[jail]# cd $B/core/octavia && make -n haproxy_amphora_image | grep -E "diskimage-create.sh -o|md5sum"
[jail]# cd $B/core       && make -C octavia -n haproxy_amphora_image | grep -E "diskimage-create.sh -o|md5sum"
byte-identical output from both

Config sample regeneration, method proved before use:
generated in an isolated venv off the caracal python 3.11 in the built rootfs, and the
*antelope* sample regenerated the same way under octavia 12.0.1 first -- diff against
the checked-in file was empty, so the caracal output is trustworthy.
```


core/octavia/Makefile
```diff
@@ -2,20 +2,152 @@
...
+	$(Q)cd $(OCTAVIA_SRC_DIR)/diskimage-create && \
+		DIB_REPOREF_amphora_agent=$$(git -C $(OCTAVIA_SRC_DIR) rev-parse HEAD) \
+		DIB_REPOREF_octavia_lib=$$(git ls-remote $(OCTAVIA_LIB_REPO) \
+			'refs/tags/$(OCTAVIA_LIB_VER)^{}' 'refs/tags/$(OCTAVIA_LIB_VER)' \
+			| sed -n '$$p' | cut -f1) \
+		bash -x ./diskimage-create.sh -o $(AMPHORA_IMAGE)
+	$(Q)md5sum $(AMPHORA_IMAGE) | awk '{print $$1}' > $(AMPHORA_IMAGE).md5
+	$(Q)echo "  AMPHORA $(notdir $(AMPHORA_IMAGE)) ($$(cat $(AMPHORA_IMAGE).md5))"
+
+.PHONY: haproxy_amphora_image
```
traviswu-bigstack, line 151: does it mean we build amphora image every build?

Eandalf-Bigstack: No, both core/manila/Makefile and core/octavia/Makefile are not referenced by our build system, they could only be built using the intentional `make -C core/{manila,octavia} <target>`. We did not register the build targets in the variable `BUILD` and neither manila nor octavia is listed inside core/Makefile, so they would not participate the build.

traviswu-bigstack (approved):

PR #1374, opened 2026-08-27
feat(barbican): upgrade barbican from openstack antelope to caracal

### LINKED ISSUE #632: [Story] [Parallel Group] Upgrade Barbican from Antelope to Caracal
## Description

As a CubeCOS operator, I want Barbican be in Caracal so that we could upgrade to Caracal.

## Acceptance Criteria

- [ ] Migrate configs
- [ ] Pass health check

## Story Points: 3

> | -- | -- | -- |

## QA Verification

> Rows below are **checkable claims**: the exact command plus the expected outcome. Run them
> **from a control node as root**, with `source /etc/admin-openrc.sh` unless a row says otherwise.
> Shipped by cubecos **PR #1374**, 2 commits over 6 files — `core/barbican/*` (`.mk`, unit,
> `config-generator.conf`, regenerated `barbican.conf.sample`), `core/modules/config_barbican.cpp`,
> `project.mk`.
>
> **Scope of this section.** A `User Story` (3 points): barbican hops from 16.0.2 in
> `/opt/openstack-antelope` to **18.0.0 in `/opt/openstack-caracal`**. Sections below are
> organised against this ticket's three Acceptance Criteria, so it is visible which rows prove
> which AC. Other services' caracal hops are their own tickets.
>
> **No UI.** CubeCOS's portal exposes no key-manager page, so every row here is CLI/API driven.
>
> **§3 is the row that decides the verdict.** AC3 is "ignore the new features" — and this hop has
> exactly one changed default that was deliberately *not* ignored. If §3 is wrong, the cluster is
> wrong in a way no other row detects.

### The one thing to hold in your head before testing

barbican 18.0.0 turns Secure RBAC on for itself, and **CubeCOS accepts that default** (decided in
review of PR #1374). The effective change: `secrets:post` moves from `rule:admin_or_creator` to
`role:member`. Before this ticket only the `barbican` service user could create secrets, because
it is the only principal holding `creator`. After it, **any user with `member` in a project can
store secrets in that project.** That is intended, not a defect.

One consequence: under the new rule an **admin** is allowed only because keystone's implied-role
chain (`admin → manager → member → reader`) puts `member` in the admin token. §2 checks that chain
directly, so a failure there can be told apart from a barbican problem.

### Preconditions

- A CubeCOS cluster installed from an ISO that contains PR #1374. Verify before starting:
  `ls -d /opt/openstack-caracal/lib/python3.11/site-packages/barbican-18.0.0.dist-info`
  must exist. **On a build without this PR every row below either fails or passes for the wrong
  reason — check this first.**
- `openstack` CLI usable as admin (`source /etc/admin-openrc.sh`).
- The rows create a project, a user, a volume type, a volume and secrets. All are removed in
  **Cleanup**; run it even if a row fails.

### 1 · Migrate configs

The hop moves the service between venvs and renames one config key. Nothing in `barbican.conf`
is hand-edited — every effective value comes from `config_barbican.cpp`, so these rows check the
generated result, not the sample.

|---|---|---|
| 1 | `/opt/openstack-caracal/bin/pip show barbican \| head -2` | `Name: barbican` / `Version: 18.0.0` |
| 2 | `/opt/openstack-antelope/bin/pip show python-barbicanclient \| head -2` | `Name: python-barbicanclient` / `Version: 5.5.0` — the client **did not** move, by design |
| 3 | `head -1 /usr/bin/openstack` | `#!/opt/openstack-antelope/bin/python3.10` — this is why the client had to stay |
| 4 | `head -1 /opt/openstack-caracal/bin/barbican-manage` | shebang points into `/opt/openstack-caracal` |
| 5 | `grep -A3 '^\[database\]' /etc/barbican/barbican.conf` | a `connection = mysql+pymysql://...barbican` line |
| 6 | `grep -c '^sql_connection' /etc/barbican/barbican.conf` | `0` — the deprecated `[DEFAULT] sql_connection` spelling is gone |
| 7 | `journalctl -u openstack-barbican-api -b \| grep -Ec 'deprecat\|sql_connection'` | `0` |
| 8 | `/opt/openstack-caracal/bin/barbican-manage db current` | `8c74e2d7f1ff (head)` |
| 9 | `/opt/openstack-caracal/bin/barbican-manage db upgrade` | alembic context messages only; **no migration is executed** — the schema did not change across the hop |

Row 6 matters more than it looks: the old spelling still works in 18.0.0 as a deprecated alias, so
a cluster that kept it would look healthy. The failure only arrives when the alias is dropped
upstream, and it arrives silently — the option's default is `sqlite:///barbican.sqlite`, so each
control node would open its own empty SQLite key store while every health check stayed green.

### 2 · Pass health check

`hex_cli -c cluster check` has **no KeyManager group** — see Known behaviour #2 — so a green
cluster check is a cluster-wide regression check here, not evidence about barbican. Rows 3-6 are
the functional health check that actually covers it.

|---|---|---|
| 1 | `systemctl is-active openstack-barbican-api openstack-barbican-keystone-listener` | `active` for both |
| 2 | `hex_cli -v -c cluster check` | all groups `ok` |
| 3 | `openstack implied role list -f value \| awk '{print $2" implies "$4}'` | includes `admin implies manager`, `manager implies member`, `member implies reader` |
| 4 | `openstack secret store --name qa-632-admin --payload s3cr3t-admin -f value -c "Secret href"` | a URL like `http://<shared-ip>:9311/v1/secrets/<uuid>` |
| 5 | `openstack secret get --payload <href>` | payload reads exactly `s3cr3t-admin` |
| 6 | `openstack secret delete <href>` then `openstack secret list \| grep -c qa-632-admin` | `0` |

Row 3 is not ceremony: under the new policy an admin token satisfies `role:member` only through
that chain, so it is the precondition for rows 4-6 and for §4. If row 3 is missing
`manager implies member`, rows 4-6 will fail with `403 Forbidden` and that is a **keystone**
finding, not a barbican one — say so in the defect.

### 3 · Ignore the new features introduced in the new version

18.0.0's new *features* are not adopted. The one thing that did change is a changed **default**,
not a feature, and it was deliberately accepted rather than ignored — that distinction is what
this section checks.

First, the features that stay unused:

|---|---|---|
| 1 | `grep -c 'enabled_certificate_plugins\|snakeoil' /etc/barbican/barbican.conf` | `0` — certificate orders were removed upstream and CubeCOS never enabled them |
| 2 | `grep '^enabled_secretstore_plugins\|^enabled_crypto_plugins' /etc/barbican/barbican.conf` | `store_crypto` and `simple_crypto` — the same plugins as before the hop, no new backend |

Then the changed default that was **not** ignored. Create the fixture (it does not exist on a
stock cluster):

```bash
openstack project create --domain default qa-632-proj
openstack user create --domain default --project qa-632-proj --password 'Qa632Pass!' qa-632-user
openstack role add --project qa-632-proj --user qa-632-user member
```

Then, in a **separate shell** (do not `source admin-openrc.sh` here):

```bash
export OS_AUTH_URL=$(awk -F= '/OS_AUTH_URL/{print $2}' /etc/admin-openrc.sh | tr -d '" ')
export OS_IDENTITY_API_VERSION=3 OS_USER_DOMAIN_NAME=default OS_PROJECT_DOMAIN_NAME=default
export OS_PROJECT_NAME=qa-632-proj OS_USERNAME=qa-632-user OS_PASSWORD='Qa632Pass!'
```

|---|---|---|
| 3 | `openstack secret store --name qa-632-member --payload ok -f value -c "Secret href"` | **succeeds**, returns a secret href. It must **not** say `Forbidden: Secret creation attempt not allowed` — that string means the Antelope policy is still in force and the hop did not fully take. |
| 4 | `openstack secret get --payload <href>` | `ok` |
| 5 | back in the admin shell: `openstack secret list --all-projects \| grep qa-632-member` | admin sees the member's secret |
| 6 | in the member shell: `openstack secret delete <href>` | succeeds |

**Do not file row 3 succeeding as a security defect.** Widening `secrets:post` to `role:member` is
barbican 18.0.0's own default, accepted in review of PR #1374 on the grounds that no CubeCOS
deployment currently uses barbican as a vault, and that each key held back from the community
default is a divergence to carry at the next hop. Row 3 *failing* is the defect.

### 4 · The failure mode this change introduces — test it deliberately

Cinder's encrypted volumes create their LUKS key through the requesting user's context, so they
run through the same `secrets:post` rule §3 just changed. It is the one internal consumer that
would break silently, and no row in §1-§3 would catch it.

|---|---|---|
| 1 | `openstack volume type create qa-632-enc` | type created |
| 2 | `openstack volume type set --encryption-provider luks --encryption-cipher aes-xts-plain64 --encryption-key-size 256 --encryption-control-location front-end qa-632-enc` | no error |
| 3 | `openstack volume create --type qa-632-enc --size 1 qa-632-vol` then `openstack volume show qa-632-vol -f value -c status` | reaches `available`, not `error` |
| 4 | `openstack secret list` | one **new unnamed** secret appears — that is cinder's LUKS key |

1. **`openstack secret consumer list <href>` fails from the CLI** with
   `Server does not support secret consumers. Minimum key-manager microversion required: 1.1`.
   The server is fine — `curl -H "X-Auth-Token: $TOK" <href>/consumers` returns
   `HTTP 200 {"consumers": [], "total": 0}`. The cause is that
   `core/barbican/barbican-api-paste.ini` carries no microversion middleware, so the version
   document advertises no maximum microversion and client 5.5.0 refuses locally without ever
   sending a request. That file is **not touched by PR #1374** — identical before and after the
   hop. Adding the middleware is a separate ticket.
2. **`hex_cli -c cluster check` does not cover barbican.** There is no KeyManager group in the
   health output (designate has `DNSaaS`; barbican has no counterpart). Pre-existing.
3. **`/etc/cinder/cinder.conf` has an empty `[key_manager]` section.** §4 works anyway, through
   castellan's default backend. Unchanged by this ticket.
4. **Certificate orders are gone.** barbican 18.0.0 removed the certificate order plugins
   upstream. Not a regression introduced here — the plugins were never enabled in CubeCOS.

### Cleanup

```bash
source /etc/admin-openrc.sh
openstack volume delete qa-632-vol            # wait for it to leave `volume list`
openstack volume type delete qa-632-enc
openstack secret delete <href>                # each href created in §2 §3 §4, including the unnamed LUKS key
openstack user delete qa-632-user
openstack project delete qa-632-proj
```

### Already verified on the 1cc — and what that did not cover

Dev verified §1, §2 (rows 1-2, 4-6) and §4 on a 1cc VM built from ISO
`CUBE_3.1.20_20260827-0445_7be2fe3` (`distclean iso test`, `E2ETST=1cc`; build, sys/iso 6/6 and
e2e 4/4 all green).

**That ISO predates the review decision** — it still pinned `enforce_new_defaults` /
`enforce_scope` to `false`, so §3 was verified only in its *opposite* form (the member user got
`Forbidden`). The inversion was not re-measured on hardware; it is derived from
`barbican/common/policies/secrets.py:137` and `keystone/cmd/bootstrap.py:189-198`. **§3 row 3 and
§2 row 3 are therefore the load-bearing rows in this section.**

Also not covered: a **multi-control cluster** (all dev evidence is single-node, so nothing
exercised barbican's shared database from more than one API worker), and an **upgraded** cluster
rather than a fresh install — §2 row 3's implied-role chain is the thing at risk there.

- [ ] **QA Status declared at Done** — this is a `User Story`, which carries a `QA Status`, and at
      `Done` the dev declares **`Ready to QA`**. (A `Done` ticket left empty is undecided and gets flagged.)
  one — **do not reopen this ticket**. §3 is where that is most likely.

- [ ] **Handbook knowledge update** — bigstack-handbook PR
      [#412](https://github.com/bigstack-oss/bigstack-handbook/pull/412):
      `kb/cubecos/architecture/barbican-caracal-runtime.md` plus two known-issue notes
      (`barbican-caracal-drops-certificate-orders.md`,
      `barbican-secret-consumer-cli-needs-microversion.md`). **Open, not merged yet** — tick
      this box when it merges.

SekiXu:
#### What type of PR is this?

/kind feature

barbican moves from the antelope venv's 16.0.2 to **18.0.0 in `/opt/openstack-caracal`** — the seventh service to make the hop, after keystone, glance, cinder, nova/placement and neutron. 18.0.0 is the 2024.1 release, verified as the newest tag that is an ancestor of upstream's `unmaintained/2024.1` (19.0.0 has already diverged onto 2024.2).

What ends up where:

| Component | venv | Version |
| -- | -- | -- |
| `barbican` (api, worker, retry, keystone-listener) | `/opt/openstack-caracal` | 18.0.0 |
| `python-barbicanclient` | `/opt/openstack-antelope` | 5.5.0 — **stays** |

`python-barbicanclient` does not follow the service. It owns the `openstack secret ...` commands as stevedore entry points, and those are only visible to the interpreter they were installed under, so they have to stay where `/usr/bin/openstack` is. This is the first caracal hop to hit that — keystone's and cinder's clients contribute no `[openstack.cli.extension]` at all, because `openstack image|volume|server` are osc built-ins. Marked `TEMPORARY`; it goes once the CLI itself reaches the caracal venv.

Nothing had to be added to `os-caracal-pip-upper-constraints.txt`: every package named here is already pinned there, gunicorn at 21.2.0, which still accepts `--paste` and every key in `gunicorn-config.py`. The nine console scripts are unchanged between the two releases, so the symlinks only change target.

`config-generator.conf` is resynced to upstream 18.0.0's namespace list and `barbican.conf.sample` regenerated against it — four sections disappear (`[certificate]`, `[certificate_event]`, `[snakeoil_ca_plugin]`, `[ssl]`) and two arrive (`[database]`, `[oslo_versionedobjects]`). The file still carries no uncommented key, so every effective value continues to come from `config_barbican.cpp`.

One change in `config_barbican.cpp`:

**`[database] connection` instead of `[DEFAULT] sql_connection`.** barbican 18.0.0 registers `core_db_opts` under `OptGroup(name='database')`. The old spelling survives as a deprecated alias, so writing it would still work *today* — but the option's default is `sqlite:///barbican.sqlite`, so once the alias goes, barbican does not complain about a missing URI, it quietly opens a per-node sqlite file. Each control node would serve its own empty key store while every service and health check reported success.

#### Secure RBAC: the community default is taken as it comes

An earlier revision of this PR pinned `[oslo_policy] enforce_new_defaults` and `enforce_scope` back to `false`, to hold antelope's authorization behaviour across the hop. **Per review that is dropped** — barbican 18.0.0's own defaults now stand, on the grounds that no CubeCOS deployment currently uses barbican as a vault and staying on the community's policy set is the lower maintenance burden.

The effective change is that `secrets:post` moves from `rule:admin_or_creator` to `role:member`, so any user holding `member` in a project can store secrets in that project. Two consequences, both checked against the 18.0.0 source rather than assumed:

- **`enforce_scope=true` is inert here.** Every rule in all eight policy modules (`secrets`, `containers`, `orders`, `quotas`, `secretstores`, `acls`, `consumers`, `transportkeys`) is `scope_types=['project']` — there is no system-scoped rule to fail against, and `/etc/admin-openrc.sh` is project-scoped.
- **admin now passes `secrets:post` only through keystone's implied-role chain.** The new rule is `True:%(enforce_new_defaults)s and role:member` (`barbican/common/policies/secrets.py:137`), so an admin token must carry `member`. Caracal's `keystone-manage bootstrap` builds `admin → manager → member → reader` and explicitly drops the old direct admin→member link (`keystone/cmd/bootstrap.py:189-198`), so a fresh install is fine. But `os_keystone_legacy_member_role_setup` (`core/sdk_sh/modules/sdk_os.sh:1150`) only repairs `member→reader` and `_member_→member` — it never touches `manager→member`, and `manager` appears nowhere else in this repo. A cluster upgraded from a release that deleted `member` could therefore have a broken chain and get a 403 on secret creation. That is pre-existing, but accepting the new defaults is what makes it load-bearing; flagging it here rather than leaving it implicit. Happy to open a separate ticket for the repair function.

#### Verification

ISO `CUBE_3.1.20_20260827-0445_7be2fe3` (`distclean iso test`, `E2ETST=1cc`) — build, sys/iso (6/6) and e2e (4/4) all green. Then on the 1cc:

| AC | Command | Result |
| -- | -- | -- |
| Services up | `systemctl` / `hex_cli -v -c cluster check` | api + keystone-listener `active running`; cluster check all ok |
| secret CRUD (admin) | `secret store` → `list` → `get --payload` → `delete` | payload returned verbatim, delete leaves list empty |
| Encrypted volume round-trip | LUKS volume type → `volume create` | volume `available`; a new unnamed secret appears in barbican (cinder's LUKS key) |
| `db upgrade` is a no-op | `barbican-manage db upgrade` / `db current` | no migration runs; `8c74e2d7f1ff (head)` |
| No deprecation warnings | `journalctl -u openstack-barbican-api -b` | 0 hits for `deprecat\|sql_connection` |

> [!IMPORTANT]
> **That ISO was built with the policy gate still in place.** It is direct evidence for the venv hop, the `[database] connection` rename, the `db upgrade` no-op and the absence of deprecation warnings — none of which the review change touches. It is **not** evidence for authorization behaviour on this head, and I am not rebuilding for that. What changed and how it was checked instead:
>
> | Row | Status on this head |
> | -- | -- |
> | Services up, `db upgrade` no-op, no deprecation warnings, venv layout | stands as measured |
> | Encrypted volume round-trip | measured under `rule:admin_or_creator`; on this head the same call resolves through `role:member`, which admin holds via `admin → manager → member` |
> | `member`-only user gets `Forbidden` | **inverted by design** — that user now succeeds. Removed from the table rather than re-run. |
>
> The replacement for the removed row is source, not a rerun: `barbican/common/policies/secrets.py:137` for the rule, `keystone/cmd/bootstrap.py:189-198` for the chain that keeps admin working. On a cluster where that chain is intact — every fresh Caracal install — nothing above regresses. The one case worth a QA eye is an upgraded cluster with a broken `manager→member` link, which `openstack implied role list` shows in one command; #632's QA Verification section carries it as a row.

Close #632

> Note

This PR's change set is 2 commits over 4 files plus `project.mk`:

| Commit | Change |
| -- | -- |
| `a574ad6d` | `feat(barbican)`: `.mk` + unit + generator input + regenerated sample + `project.mk` note |
| `8e597c5c` | `feat(barbican)`: `[database] connection` |

The policy change was amended out of `8e597c5c` rather than reverted in a third commit, since this line is fast-forward-only and those two keys should not exist in its history at all.

Three things worth knowing:

1. **`openstack secret consumer list` does not work from the CLI — and did not before this PR either.** The client reports `Server does not support secret consumers. Minimum key-manager microversion required: 1.1`, but the server is fine: `curl <secret-href>/consumers` returns `HTTP 200 {"consumers": [], "total": 0}`. The chain is that `core/barbican/barbican-api-paste.ini` carries no microversion middleware in its pipeline, so the version document advertises no max microversion and client 5.5.0 refuses locally without sending a request. That file is checked into this repo and **is not touched by this PR**, so this is pre-existing, not a regression. Adding the middleware is a separate ticket.

2. **barbican is not covered by `hex_cli -c cluster check`.** There is no KeyManager group in the health output (designate has `DNSaaS`, barbican has no counterpart). Also pre-existing; noting it so the green cluster check above is not read as covering barbican.

3. **`/etc/cinder/cinder.conf` has an empty `[key_manager]` section.** The encrypted-volume round-trip works anyway, via castellan's default backend. Unchanged by this PR, but it means the cinder → barbican path has no explicit configuration behind it.

### DIFF
--- a/core/modules/config_barbican.cpp
+++ b/core/modules/config_barbican.cpp
@@ -176,7 +176,25 @@ UpdateDbConn(std::string sharedId, std::string password)
         dbconn += sharedId;
         dbconn += "/barbican";
 
-        cfg["DEFAULT"]["sql_connection"] = dbconn;
+        // [database] connection, not [DEFAULT] sql_connection: barbican 18.0.0 moved the
+        // database options into their own group (barbican/common/config.py registers
+        // core_db_opts under OptGroup(name='database')). The old spelling survives as a
+        // deprecated alias, so writing it would still work today -- but the regenerated
+        // barbican.conf.sample only documents the new one, and leaving the two disagreeing
+        // hands the next hop a failure that produces no error at all: the option's default
+        // is `sqlite:///barbican.sqlite`, so once the alias goes barbican does not complain
+        // about a missing URI, it quietly opens a per-node sqlite file. Each control node
+        // would then serve its own empty key store while every service and health check
+        // reported success.
+        //
+        // No mysql_wsrep_sync_wait here, unlike the eleven other upgraded modules. It
+        // would be a dead key: barbican does not use oslo.db's enginefacade and does not
+        // register oslo.db's [database] options. barbican/model/repositories.py builds
+        // engine_args itself from exactly three values -- connection_recycle_time,
+        // max_pool_size, max_overflow -- and hands them to
+        // oslo_db.sqlalchemy.session.create_engine(), which only applies
+        // mysql_wsrep_sync_wait when it arrives as a kwarg. Nothing would read it.
+        cfg["database"]["connection"] = dbconn;
     }
 
     return true;
--- a/core/barbican/barbican.conf.sample
+++ b/core/barbican/barbican.conf.sample
@@ -28,30 +28,6 @@
 # (string value)
 #host_href = http://localhost:9311
 
-# SQLAlchemy connection string for the reference implementation
-# registry server. Any valid SQLAlchemy connection string is fine.
-# See:
-# http://www.sqlalchemy.org/docs/05/reference/sqlalchemy/connections.html#sqlalchemy.create_engine.
-# Note: For absolute addresses, use '////' slashes after 'sqlite:'.
-# (string value)
-#sql_connection = sqlite:///barbican.sqlite
-
-# Period in seconds after which SQLAlchemy should reestablish its
-# connection to the database. MySQL uses a default `wait_timeout` of 8
-# hours, after which it will drop idle connections. This can result in
-# 'MySQL Gone Away' exceptions. If you notice this, you can lower this
-# value to ensure that SQLAlchemy reconnects before MySQL can drop the
-# connection. (integer value)
-#sql_idle_timeout = 3600
-
-# Maximum number of database connection retries during startup. Set to
-# -1 to specify an infinite retry count. (integer value)
-#sql_max_retries = 60
-
-# Interval between retries of opening a SQL connection. (integer
-# value)
-#sql_retry_interval = 1
-
 # Create the Barbican database on service startup. (boolean value)
 #db_auto_create = false
 
@@ -75,41 +51,6 @@
 # Show SQLAlchemy pool-related debugging output in logs (sets DEBUG
 # log level output) if specified. (boolean value)
 #sql_pool_logging = false
-
-# Size of pool used by SQLAlchemy. This is the largest number of
-# connections that will be kept persistently in the pool. Can be set
-# to 0 to indicate no size limit. To disable pooling, use a NullPool
-# with sql_pool_class instead. Comment out to allow SQLAlchemy to
-# select the default. (integer value)
-#sql_pool_size = 5
-
-# # The maximum overflow size of the pool used by SQLAlchemy. When the
-# number of checked-out connections reaches the size set in
-# sql_pool_size, additional connections will be returned up to this
-# limit. It follows then that the total number of simultaneous
-# connections the pool will allow is sql_pool_size +
-# sql_pool_max_overflow. Can be set to -1 to indicate no overflow
-# limit, so no limit will be placed on the total number of concurrent
-# connections. Comment out to allow SQLAlchemy to select the default.
-# (integer value)
-#sql_pool_max_overflow = 10
-
-# Enable eventlet backdoor.  Acceptable values are 0, <port>, and
-# <start>:<end>, where 0 results in listening on a random tcp port
-# number; <port> results in listening on the specified port number
-# (and not enabling backdoor if that port is in use); and
-# <start>:<end> results in listening on the smallest unused port
-# number within the specified range of port numbers.  The chosen port
-# is displayed in the service's log file. (string value)
-#backdoor_port = <None>
-
-# Enable eventlet backdoor, using the provided path as a unix socket
-# that can receive connections. This option is mutually exclusive with
-# 'backdoor_port' in that only one should be provided. If both are
-# provided then the existence of this option overrides the usage of
-# that option. Inside the path {pid} will be replaced with the PID of
-# the current process. (string value)
-#backdoor_socket = <None>
 
 #
 # From oslo.log
@@ -178,7 +119,10 @@
 # log_config_append is set. (boolean value)
 #use_stderr = false
 
-# Log output to Windows Event Log. (boolean value)
+# DEPRECATED: Log output to Windows Event Log. (boolean value)
+# This option is deprecated for removal.
+# Its value may be silently ignored in the future.
+# Reason: Windows support is no longer maintained.
 #use_eventlog = false
 
 # The amount of time before the log files are rotated. This option is
@@ -312,14 +256,6 @@
 #rpc_ping_enabled = false
 
 #
-# From oslo.service.periodic_task
-#
-
-# Some periodic tasks can be run in a separate process. Should we run
-# them here? (boolean value)
-#run_external_periodic_tasks = true
-
-#
 # From oslo.service.service
 #
 
@@ -347,46 +283,6 @@
 # Specify a timeout after which a gracefully shutdown server will
 # exit. Zero value means endless wait. (integer value)
 #graceful_shutdown_timeout = 60
-
-#
-# From oslo.service.wsgi
-#
-
-# File name for the paste.deploy config for api service (string value)
-#api_paste_config = api-paste.ini
-
-# A python format string that is used as the template to generate log
-# lines. The following values can beformatted into it: client_ip,
-# date_time, request_line, status_code, body_length, wall_seconds.
-# (string value)
-#wsgi_log_format = %(client_ip)s "%(request_line)s" status: %(status_code)s  len: %(body_length)s time: %(wall_seconds).7f
-
-# Sets the value of TCP_KEEPIDLE in seconds for each server socket.
-# Not supported on OS X. (integer value)
-#tcp_keepidle = 600
-
-# Size of the pool of greenthreads used by wsgi (integer value)
-#wsgi_default_pool_size = 100
-
-# Maximum line size of message headers to be accepted. max_header_line
-# may need to be increased when using large tokens (typically those
-# generated when keystone is configured to use PKI tokens with big
-# service catalogs). (integer value)
-#max_header_line = 16384
-
-# If False, closes the client socket connection explicitly. (boolean
-# value)
-#wsgi_keep_alive = true
-
-# Timeout for client connections' socket operations. If an incoming
-# connection is idle for this number of seconds it will be closed. A
-# value of '0' means wait forever. (integer value)
-#client_socket_timeout = 900
-
-# True if the server should send exception tracebacks to the clients
-# on 500 errors. If False, the server will respond with empty bodies.
-# (boolean value)
-#wsgi_server_debug = false
 
 
 [audit_middleware_notifications]
@@ -416,32 +312,6 @@
 # specified, we fall back to the same configuration used for RPC.
 # (string value)
 #transport_url = <None>
-
-
-[certificate]
-
-#
-# From barbican.certificate.plugin
-#
-
-# Extension namespace to search for plugins. (string value)
-#namespace = barbican.certificate.plugin
-
-# List of certificate plugins to load. (multi valued)
-#enabled_certificate_plugins = simple_certificate
-
-
-[certificate_event]
-
-#
-# From barbican.certificate.plugin
-#
-
-# Extension namespace to search for eventing plugins. (string value)
-#namespace = barbican.certificate.event.plugin
-
-# List of certificate plugins to load. (multi valued)
-#enabled_certificate_event_plugins = simple_certificate_event
 
 
 [cors]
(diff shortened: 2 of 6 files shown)


core/modules/config_barbican.cpp
```diff
@@ -245,6 +263,29 @@ UpdateCfg(std::string domain, std::string userPass, std::string cryptoPass)
...
+        // The rule definitions themselves are identical in 16.0.2 and 18.0.0 (both ship
+        // the legacy rule and the new one, gated on enforce_new_defaults), so switching
+        // the gate back reproduces the current behaviour exactly rather than approximately.
+        //
+        // This belongs here rather than in barbican.conf.sample: that file has to stay
+        // byte-identical to oslo-config-generator output or the next hop loses its diff
+        // baseline.
+        cfg["oslo_policy"]["enforce_new_defaults"] = "false";
+        cfg["oslo_policy"]["enforce_scope"] = "false";
+
```
Eandalf-Bigstack: We would prefer to accept the new default set by OpenStack services since this is a community decision and lower the maintaining burden on our side to not be away from the community at the same time. For the concern, "The new default is `secrets:post = role:member`, and config_keystone.cpp binds `member` to the cube_users group -- every ordinary user would gain write access to the key manager. Widening that is a product decision, not an upgrade side effect.", this is acceptable since we do not have any existing users using Barbican as the vault, we would accept the community change here.

Eandalf-Bigstack:
Overall changes are good so far, please also update the issue https://github.com/bigstack-oss/cubecos/issues/632 format against the updated user story template in the repository for fulfilling the QA team's request.

core/modules/config_barbican.cpp
SekiXu: Understood — both keys dropped, 18.0.0's defaults stand. Amended into `8e597c5c` rather than a
revert commit, since this line is fast-forward-only; that commit is now just the
`[DEFAULT] sql_connection` → `[database] connection` rename.

One thing worth flagging. `enforce_scope=true` is inert here — every barbican policy is
`scope_types=['project']`. But `enforce_new_defaults=true` makes `secrets:post` = `role:member`
(`barbican/common/policies/secrets.py:137`), so **admin now passes only through keystone's
`admin → manager → member` chain**. Caracal's bootstrap builds it
(`keystone/cmd/bootstrap.py:189-198`), but `os_keystone_legacy_member_role_setup`
(`core/sdk_sh/modules/sdk_os.sh:1150`) only repairs `member→reader` and `_member_→member` —
never `manager→member`. A cluster upgraded from a release that deleted `member` could leave an
admin with a bare 403 on secret creation. Pre-existing, but this change makes it load-bearing;
I can open a separate ticket for the repair function.

Not rebuilt for this. The ISO in the description still covers the venv hop, the config rename
and the `db upgrade` no-op; the authorization row it carried is an intended inversion, so I
replaced it with the source references above rather than leave a table reading as if it
covered this head.

SekiXu:
Done — both sections added. The QA section is organised against the ticket's own three ACs, so
it is visible which rows prove which. One thing worth your eye: the Secure RBAC widening sits
under AC3, since a changed default is not a new feature — the section says so explicitly, so
QA does not read member-can-store-secrets as a defect.

core/modules/config_barbican.cpp
Eandalf-Bigstack: Great! I would handle the manager implied role missing issue.

Eandalf-Bigstack (approved):

PR #1387, opened 2026-08-29
feat(ceph): upgrade Ceph from quincy 17.2.6 to reef 18.2.8

### LINKED ISSUE #650: [Story] Upgrade Ceph from Quincy to Reef
## Description

As a CubeCOS operator, I want to resolve Ceph Quincy EOL security consern so that we could pass audits.

## Acceptance Criteria

- [x] Migrate configs
- [x] Ensure OSDs are correctly set up and identified
- [x] Ensure OSD pools are set up
- [x] Ensure we could perform RBD operations
- [x] Ensure we have CephFS ready
- [x] Ensure we have the RGW gateway ready for S3 object storage
- [x] Update ceph-csi to make it compatible

Delivered by **[PR #1387](https://github.com/bigstack-oss/cubecos/pull/1387)** — Ceph quincy 17.2.6 → **reef 18.2.8-1.el9s** (CentOS Stream 9 Storage SIG).

**ceph-csi needed no update.** 3.13.1's support matrix floors at `Pacific (>=v16.2.0)`, so reef is inside its window; it was tested rather than assumed and both drivers provision, bind and mount against a fully-reef cluster (row 6 below). The AC is met by verification, not by a version bump.

**Reef is a hop, not a destination — squid is the next CubeCOS version's move.** Reef is archived upstream too (18.2.8, 2026-03-20, is the last of the series), so this does not end the EOL exposure on its own. It is still the correct hop, because ceph upgrades across at most two majors: squid's notes say *"Upgrading from Quincy or Reef"*, while tentacle's say *"You must first upgrade to Reef (18.2.z) or Squid (19.2.z)"*. Quincy → tentacle is not a legal path; reef keeps both doors open.

**Three defects had to be fixed for a quincy cluster to reach reef intact**, each of which would have broken a real upgrade:

1. **Every OSD refuses to start on reef.** Reef added `bluestore_cache_kv_onode_ratio` to a sum CubeCOS already tuned to exactly 1.0, so the total became 1.04 and BlueStore refused to mount — `osd init failed: (22) Invalid argument`, on every start.
2. **The old `require-osd-release` call would have stranded a rolling upgrade permanently.** It read the release off the *local* CLI and pinned unconditionally; on the first node into reef ceph refuses, the refusal was never checked, and the marker was written anyway — so it never retried and the cluster would have kept `require_osd_release quincy` for good.
3. **Daemons came up in the wrong order** — mds and rgw ahead of the OSDs, and pools created before an OSD existed to place their PGs on.

## Story Points: 5

> | -- | -- | -- |

## QA Verification

|---|---|---|
| 1 | `ceph versions \| jq -r '.overall'` on any control node | A **single** entry, `ceph version 18.2.8 … reef (stable)`, counting every daemon (36 on a 3-node converged cluster). No quincy entry. |
| 2 | `ceph -s` | `health: HEALTH_OK`; mons in quorum, every OSD `up`/`in`, mds up, one rgw per host, `volumes: 1/1 healthy`. |
| 3 | `ceph mon dump \| grep min_mon_release` and `ceph osd dump \| grep require_osd_release` | `min_mon_release 18 (reef)` and `require_osd_release reef`. Both must read reef — `require_osd_release` still on quincy means the upgrade never finalised. |
| 4 | `ceph osd pool ls \| wc -l`, then `ceph osd tree` | Pool count unchanged from before the upgrade (25 on a stock cluster); every OSD listed under its own host with status `up`. |
| 5 | `rbd create cinder-volumes/qa --size 128` → `rbd snap create cinder-volumes/qa@s1` → `rbd snap protect …@s1` → `rbd clone …@s1 cinder-volumes/qaclone` → `rbd export cinder-volumes/qa /tmp/out` → `rbd rm` each | Every step exits 0; the exported file matches what was imported byte-for-byte (`cmp` / matching `md5sum`). |
| 6 | Create a PVC on `ceph-rbd` and one on `ceph-fs`, mount both in a pod, write a file to each | Both PVCs reach `Bound`; the pod reaches `Running`; both files read back. `kubectl get pods -A \| grep csi` shows rbd + cephfs provisioner and nodeplugin pods `Running`. |
| 7 | `echo x > /mnt/cephfs/.qa && cat /mnt/cephfs/.qa && rm /mnt/cephfs/.qa`; then `ceph fs status` | File round-trips; one MDS `active` on rank 0, standbys present on an HA cluster. |
| 8 | `radosgw-admin user list`; `systemctl is-active ceph-radosgw@rgw.$(hostname)` | Pre-upgrade users still listed; the unit is `active` on each control node. |
| 9 | `hex_cli -v -c cluster -c check` | Every service group `ok`, in particular `Storage ok [ ceph(v) ceph_mon(v) ceph_mgr(v) ceph_mds(v) ceph_osd(v) ceph_rgw(v) rbd_target(v) fc_link(v) ]`. |
| 10 | **Rolling upgrade, mid-roll:** with one node upgraded and the rest still quincy, run `ceph versions`, then `ceph osd dump \| grep require_osd_release` | `ceph versions` shows **both** releases in the mon and osd tiers (mixed-version is expected and healthy); `require_osd_release` must still read **quincy** — it may only flip after the *last* node. |
| 11 | **Cross-version RBD, mid-roll:** `rbd import` a file and snapshot it from an upgraded (reef) node, then `rbd export` it from a not-yet-upgraded (quincy) node | Exported file matches the imported one by `md5sum`; the quincy node can also `rbd snap create` on the same image. |
| 12 | **After the last node:** `hex_sdk ceph_finalize_release`, twice | First run sets `require_osd_release reef`; the second is a clean no-op (`rc=0`, no change). |
| 13 | `openstack --os-compute-api-version 2.30 server migrate --live-migration --host <other> --wait <vm>` | Migration reports `Complete`; `openstack server show <vm>` shows the new host and status `ACTIVE`. |
| 14 | `dnf repolist \| grep -i ceph` and `dnf check --dependencies` | Only the reef repo enabled (`centos-ceph-reef`); dependency check reports **no** problems — in particular no "missing requires of centos-release-ceph-quincy". |
| 15 | `logrotate /etc/logrotate.conf ; echo rc=$?` and `systemctl status logrotate.service` | `rc=0` and the unit is not `failed`. (Regression guard for the duplicate `/etc/logrotate.d/{mariadb,mysql}` entry fixed on this branch.) |

- [x] **QA Status declared at Done** — when this issue reaches `Done`, set the board's

- [x] **Handbook knowledge update** — landed in **[bigstack-handbook #428](https://github.com/bigstack-oss/bigstack-handbook/pull/428)** (merged), four commits:

  | Entry | Type | Why it exists |
  |---|---|---|
  | [`kb/cubecos/runbooks/ceph-major-version-upgrade.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/runbooks/ceph-major-version-upgrade.md) | Runbook | Nothing in the kb covered a ceph **major** upgrade. Records the mon→mgr→osd→mds→rgw order, why an A/B roll cannot follow it across the cluster, which single step is gated and which is deliberately not, the release-lifecycle constraint that forces quincy→reef→squid, and a Rollback section stating plainly that ceph has no downgrade. Written to be reused on the squid hop. |
  | [`kb/cubecos/known-issues/ceph-reef-bluestore-cache-ratio-osd-fail.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/ceph-reef-bluestore-cache-ratio-osd-fail.md) + scenario sidecar | Known-issue (`fixed-in:3.2.0`) | Bites on the *first* node of a roll, and the daemon dies before opening its own log so `/var/log/ceph/ceph-osd.N.log` is empty — the evidence is only in `journalctl`. Flags that a future release is equally free to add a fourth ratio term. |
  | [`kb/cubecos/known-issues/health-mysql-repair-unbounded-datadir-copies.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/health-mysql-repair-unbounded-datadir-copies.md) + scenario sidecar | Known-issue (`open`) | Found while preparing accept-3cc. Presents as a **storage** outage rather than a database one, which misdirects triage. Not fixed — see below. |
  | [`kb/cubecos/known-issues/caracal-service-users-missing-service-role.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/caracal-service-users-missing-service-role.md) | Amendment | `handbook dup` showed the live-migration RBAC issue was already covered, so the two new findings were appended there instead of duplicating: `neutron-server` must also recycle for the grant to take, and `migrate_keystone_service_role` can silently grant nothing. |

  Both diagnosed faults ship a `<note>.scenario.yaml` sidecar with the reproducible injection, the raw pre-diagnosis symptom and a checkable verification. `handbook doctor` reports `HIGH=0`.

## Notes for follow-up (out of scope here)

- **`_health_mysql_repair` copies the whole mysql datadir per attempt and never prunes** — six copies × three nodes = 78 GB filled `/` on accept-3cc and killed every `ceph-mon` with **exit 28 (ENOSPC)**. The repair is capable of causing the outage it is trying to fix. Deserves its own issue.
- **Rebuilding `ceph-mgr` against python 3.11 was surveyed and dropped.** `-DWITH_PYTHON3=3.11` is a first-class cmake option and all 370 reef mgr `.py` files compile clean under 3.11, but `ceph-mgr` links `Boost::python311` and **CS9 ships only `libboost_python39`** — so it needs a from-source Boost first — and it links `osdc client global-static ceph-common`, so there is no "just the mgr". Estimated ~1.5–2.5 h per image build to move one daemon's embedded interpreter, while CS9's `libpython3.9` still receives Red Hat CVE backports (latest 2026-08-13) until CS9 EOL on 2027-05-31. Worth revisiting on the squid hop, where Fedora 43 already builds ceph 19.2.6 against python 3.14.

Eandalf-Bigstack:
#### What type of PR is this?

/kind feature

Quincy went EOL 2025-01-13, which is the audit finding #650 exists to close. This moves Ceph to reef 18.2.8 from the CentOS Stream 9 Storage SIG, and fixes the three things that stop a quincy cluster reaching reef intact.

**Reef is a hop, not a destination, and that is deliberate.** Reef is archived upstream too — docs.ceph.com lists it beside quincy, and 18.2.8 (2026-03-20) says in as many words that it is the last release of the series. It is still the right hop, because ceph upgrades across at most two majors:

- squid's notes say **"Upgrading from Quincy or Reef"** — so quincy → squid is legal directly;
- tentacle's say **"You must first upgrade to Reef (18.2.z) or Squid (19.2.z)"** — quincy → tentacle is not.

Reef keeps both doors open, and squid is the next CubeCOS version's move. The SIG already carries `ceph-squid` and `ceph-tentacle`, so that hop is the same one-line repo change.

- **Every OSD refuses to start on reef without the bluestore fix.** `bluestore` will not mount if `meta_ratio + kv_ratio + kv_onode_ratio` exceeds 1.0. Quincy summed only meta and kv, so this tree's `0.8 + 0.2 = 1.0` was exactly legal; reef added `kv_onode` to the same check and its 0.04 default pushes the total to 1.04. Not a warning — `osd init failed: (22) Invalid argument`, every start. Measured on jim-1cc: both OSDs down, cluster at `945 stale+active+clean`. meta gives up the 0.04 (0.76 still spends three quarters there; trimming kv would cost rocksdb its read cache), and `kv_onode` is now written **explicitly** at its own default — leaving it implicit is what let the sum depend on a value upstream was free to change.

- **The old `require-osd-release` call would have stranded a rolling upgrade permanently.** `migrate_ceph` read the release off the *local* CLI and pinned the cluster to it unconditionally. On the first node to reboot into reef that asks a cluster still holding quincy OSDs to disallow pre-reef ones; ceph refuses, the refusal is never checked, and the `ceph_cluster_migrated` marker is written anyway — so it never retries and the cluster keeps `require_osd_release quincy` for good. `ceph_finalize_release` replaces it: it reads the release off the cluster, and does nothing unless a cluster answers (a no-op during FTS), every OSD reports one release (a no-op for the whole of a roll), and the monmap has already advanced. Idempotent, so it is safe from every ceph commit and from the end of a roll; whichever node finishes last completes the upgrade.

- **Daemons now come up in ceph's documented order** — mon, mgr, osd, mds, rgw. `Commit()` ran every `Setup*` in one condition and then committed mgr, mds, rgw, osd, which put both the mds and the radosgw ahead of the OSDs and created the rgw and cephfs pools before there was an OSD to place their PGs on. Survivable on a restart; wrong to hold across a major upgrade. Pool and filesystem creation moved after the OSDs, `SetupFS` still ahead of the mds step (it runs `ceph fs new`, and the mds claims rank 0 on a filesystem that exists). The pauserd/pausewr lift moved out of `CommitMds` up into `Commit()`, to just after the mon answers — it was only in `CommitMds` because the mds happened to be committed first, and with the OSDs now ahead of it they would have come up under the block that call exists to avoid.

**What is deliberately *not* gated: a node's OSDs coming up.** Holding them down until every other node upgraded would, on a 3-node cluster, put two thirds of the OSDs down the moment the second node reboots — below `min_size`, client I/O stopped. Ceph's tier-at-a-time advice assumes restarts measured in minutes; an A/B roll takes a node out for a reboot and a firmware switch, and mixed-version operation is the documented steady state during an upgrade. So ordering is enforced *within* a node, and only the one cluster-wide irreversible step is gated across it.

- **Ceph gets its own python 3.11 venv, which is then deleted.** Building the rados/rbd bindings used to mean `pip install "Cython<3"` into the antelope venv *and* the caracal venv, leaving a build-time compiler in two openstack runtime environments for the life of the image. They are built once in `/opt/ceph` as wheels and installed into the caracal venv, then the venv, the wheels and the unpacked source tree are removed, with `test ! -e` guards. The quincy version of this file never cleaned up at all — the shipped rootfs has been carrying the whole ceph source tree. The second (cpython-310) build is gone with it: cinder, glance and manila all live in the caracal venv now, so nothing in antelope talks to the RBD store.

  Two reef-only build details. `src/pybind/rbd/setup.py` gained a top-level `from packaging import version`, which with `--no-build-isolation` resolves against the building venv — without it the build dies at setup.py import time before compiling anything; `rados/setup.py` is byte-identical to quincy's. `Cython<3` is kept because that is what `ceph.spec.in` builds against. `ceph-base` also picks up a new `libfmt.so.8`, satisfied by EPEL's `fmt`, already enabled.

- **The dashboard `saml2.py` patch is deleted rather than carried over.** It commented out a `token.decode('utf-8')` on an already-str token; reef fixes that upstream with an `isinstance(token, bytes)` guard (18.2.8 `saml2.py:65`), so applying the old file would overwrite the fix. Verified against the file reef actually installs.

- **`python-magic`, `python3-saml`, `xmlsec` and the `jaraco-text` pin stay on system python 3.9, audited rather than moved.** `ceph-mgr`'s reef rpm carries a hard `libpython3.9.so.1.0` dependency — it embeds the interpreter, so every mgr module, the dashboard's SAML2 controller included, is imported by that 3.9 interpreter and cannot see a 3.11 venv. mon, osd, mds and radosgw declare no python dependency at all. The comment block now records this so it is not attempted again.

- **`heavyfs` stops erasing ceph release rpms.** The `centos-release-ceph-pacific` removal was dead code — only `centos-release-openstack-yoga` and `centos-release-ovirt45` require it, and we install neither. The quincy removal was the opposite problem: `centos-release-openstack-antelope` *requires* `centos-release-ceph-quincy`, so `rpm -e --nodeps` left an unsatisfied dependency for the life of the image (`dnf check`: "has missing requires of centos-release-ceph-quincy"). Disabling the repo gets what was wanted — 17.2.z out of dnf's view — with the package installed, the dependency satisfied and the rpm database clean. It is still needed: the two release rpms own differently-named `CentOS-Ceph-*.repo` files and neither Obsoletes the other. This line disappears on its own when `OPENSTACK_RELEASE` reaches caracal, whose release rpm requires the *reef* one.

- **One non-ceph fix rides along: `logrotate` has been failing since the MariaDB 10.6 → 10.11 bump.** `MariaDB-server` ships `/etc/logrotate.d/mariadb` and `config_mysql.cpp` writes `/etc/logrotate.d/mysql`; both glob `/var/log/mariadb/*.log`, and logrotate treats two entries for one file as a hard error — it skips that file but still **exits 1**, leaving `logrotate.service` permanently failed. That matters beyond mariadb: a failed unit is the signal anything watching rotation keys off, so a genuinely stuck rotation elsewhere is indistinguishable. Found on accept-3cc with `/var/log` at 19G across three nodes.

**ceph-csi needed no change.** #650 lists it, so it was tested rather than assumed: 3.13.1's support matrix floors at `Pacific (>=v16.2.0)`, and against the fully-reef cluster both drivers provision, bind and mount — details below.

Fixes #650

> Note

**Not rebuilt for this.** Everything was verified by in-place landings on jim-1cc and accept-3cc plus build-side checks in the `centos9_cubecos_jim_200.13` jail and its `heavy_rootfs` chroot. A full image build has not run, so the `ceph.mk` venv teardown and the `heavyfs` repo change are proven by executing the same steps in the chroot rather than by a green build.

**The seven commits are meant to be read in order** — each is separable and independently revertable:

```
build(ceph): upgrade ceph from quincy 17.2.6 to reef 18.2.8
feat(ceph): gate the cluster-wide upgrade step on the whole OSD tier
fix(ceph): keep the bluestore cache ratios summable under reef
fix(ceph): bring ceph daemons up in the documented upgrade order
fix(mysql): drop MariaDB's own logrotate file so logrotate stops failing
build(ceph): keep the ceph venv out of the shipped image
build(heavyfs): stop erasing ceph release rpms, disable the quincy repo instead
```

`fix(mysql)` is unrelated to the ceph story and can be cherry-picked onto its own branch if you would rather.

**Rebuilding ceph-mgr against python 3.11 was surveyed and dropped.** The lever is real — `-DWITH_PYTHON3=3.11` is a first-class cmake option, all 370 reef mgr `.py` files compile clean under 3.11, and there are no uses of anything 3.11 removed. Two things stop it. `ceph-mgr` links `Boost::python311`, and **CS9 has no such library** — only `libboost_python39`, in AppStream and in EPEL's `boost1.78-python3` alike — so it needs a from-source Boost first. And it links `osdc client global-static ceph-common`, so there is no "just the mgr": `ninja ceph-mgr` pulls in most of ceph's core C++. Estimated ~1.5–2.5 h for a first build on an 80-core box, added to every image build, to move one daemon's embedded interpreter — while CS9's `libpython3.9` still receives Red Hat CVE backports (latest 2026-08-13, CVE-2026-11940) until CS9 EOL on 2027-05-31. Worth revisiting on the squid hop, where Fedora 43 already builds ceph 19.2.6 against python 3.14.

**Two things found on accept-3cc that are *not* fixed here**, both out of scope:

- `_health_mysql_repair` copies the whole 4.3 GB datadir before each galera bootstrap, is retried 3× per call, and never prunes. Six copies × three nodes = 78 GB, which filled `/` and killed every `ceph-mon` with **exit 28 (ENOSPC)**. The repair is capable of causing the outage it is trying to fix.
- Live migration was blocked by `rule:create_port_binding is disallowed by policy` until the `role:service` grant from `da50b615` was applied to the **neutron** user (nova's `[neutron]` client authenticates as it). Applying it needed `neutron-server` restarted, not just nova; and `hex_sdk migrate_keystone_service_role` silently granted nothing when invoked — it wraps every call in `>/dev/null 2>&1` and `touch`es its marker unconditionally, so a failed run is indistinguishable from a successful one.

#### Additional documentation

```docs
Verified by in-place landings on jim-1cc (cc1, single node) and accept-3cc
(cc1/cc2/cc3, 24 OSDs); build-side checks in the centos9_cubecos_jim_200.13
jail and its core/heavyfs/heavy_rootfs chroot.


1. THE BLUESTORE RATIOS ARE A HARD STOP, NOT A WARNING

[cc1 ~]# systemctl restart ceph-osd.target        <- with the quincy values
[cc1 ~]# journalctl -u ceph-osd@0
  bluestore(/var/lib/ceph/osd/ceph-0) _set_cache_sizes
    bluestore_cache_meta_ratio (0.8) + bluestore_cache_kv_ratio (0.2)
    + bluestore_cache_kv_onode_ratio (0.04) = 1.04; must be <= 1.0
  osd.0 0 OSD:init: unable to mount object store
  ** ERROR: osd init failed: (22) Invalid argument
[cc1 ~]# ceph -s
  osd: 2 osds: 0 up, 2 in
  pgs: 945 stale+active+clean

  with meta = 0.76 and kv_onode written explicitly:
[cc1 ~]# ceph osd stat
  2 osds: 2 up (since 15s), 2 in


2. THE GATE, ACROSS A REAL 3-NODE ROLL   (accept-3cc, 36 daemons, 24 OSDs)

  step            mon tier      osd tier       min_mon_release  finalize  require_osd_release
  --------------  ------------  -------------  ---------------  --------  -------------------
  baseline        3 quincy      24 quincy      17 (quincy)      no-op     quincy
  cc1 -> reef     2q + 1 reef   16q + 8 reef   17 (quincy)      no-op     quincy
  cc2 -> reef     1q + 2 reef    8q + 16 reef  17 (quincy)      no-op     quincy
  cc3 -> reef     3 reef        24 reef        18 (reef)        FIRES     reef

[cc1 ~]# hex_sdk ceph_tier_releases osd          <- mid-roll
  quincy
  reef
[cc1 ~]# hex_sdk ceph_finalize_release ; ceph osd dump | grep require_osd_release
  require_osd_release quincy                     <- correctly did nothing

  this is exactly where the old migrate_ceph would have fired, been refused,
  written its marker and never retried.

  after the last node, and idempotent on a second run:
[cc3 ~]# hex_sdk ceph_finalize_release ; ceph osd dump | grep require_osd_release
  require_osd_release reef
[cc3 ~]# hex_sdk ceph_finalize_release ; echo rc=$?
  rc=0                                           <- no-op


3. THE DOCUMENTED ORDER, FROM THE COMPILED Commit()

  forced through a real config change so the ceph body ran (66.2s, exit 0):
[cc1 ~]# grep 'ceph commit' /var/log/hex_config.log
  ceph commit(o) took 66.2 secs                  <- (o) = executed, (x) = skipped

[cc1 ~]# grep hex_config /var/log/hex_config.log
  04:25:51  Executing: systemctl start ceph-mon@cc1            <- MON
  04:25:52  ceph: mon quorum answered after 0s                 <- WaitForMonQuorum()
  04:25:52  Executing: /usr/sbin/hex_sdk ceph_unpause_client_io   <- hoisted lift
  04:25:53  Executing: systemctl start ceph-mgr@cc1            <- MGR
  04:26:36  Executing: /usr/sbin/hex_sdk ceph_osd_restart 0
  04:26:38  Executing: /usr/sbin/hex_sdk ceph_osd_restart 1    <- OSD
  04:26:42  Executing: systemctl start ceph-mds@cc1            <- MDS
  04:26:48  Executing: systemctl restart ceph-radosgw@rgw.cc1  <- RGW

  afterwards: HEALTH_OK, 66 keystone endpoints, 0 malformed.


4. MIXED-VERSION RBD, WHILE GENUINELY MIXED

[cc1 ~]# ceph --version | awk '{print $3,$5}'    <- reef client
  18.2.8 reef
[cc3 ~]# ceph --version | awk '{print $3,$5}'    <- quincy client
  17.2.6 quincy

  write from the reef client:
[cc1 ~]# md5sum /tmp/xver.dat
  ab5eb20e7463f1a8c670595554d8a831
[cc1 ~]# rbd import --image-format 2 /tmp/xver.dat cinder-volumes/xvertest
[cc1 ~]# rbd snap create cinder-volumes/xvertest@fromreef

  read it back from the quincy client, and write to it:
[cc3 ~]# rbd export cinder-volumes/xvertest /tmp/xver.out ; md5sum /tmp/xver.out
  ab5eb20e7463f1a8c670595554d8a831               <- identical
[cc3 ~]# rbd snap create cinder-volumes/xvertest@fromquincy
[cc3 ~]# rbd snap ls cinder-volumes/xvertest
  5  fromreef    16 MiB
  6  fromquincy  16 MiB

  and after the whole cluster reached reef, the same pre-upgrade data:
[cc1 ~]# rbd export cinder-volumes/xvertest /tmp/final.out ; md5sum /tmp/final.out
  ab5eb20e7463f1a8c670595554d8a831


5. FINAL STATE OF THE ROLLED CLUSTER

[cc1 ~]# ceph versions | jq -r '.overall'
  { "ceph version 18.2.8 (efac5a54...) reef (stable)": 36 }
[cc1 ~]# ceph -s
  health: HEALTH_OK
  mon: 3 daemons, quorum cc1,cc2,cc3
  mgr: cc1(active), standbys: cc2, cc3
  mds: 1/1 daemons up, 1 standby, 1 hot standby
  osd: 24 osds: 24 up, 24 in
  rgw: 3 daemons active (3 hosts, 1 zones)
  volumes: 1/1 healthy
  pools: 25 pools, 1457 pgs

[cc1 ~]# hex_cli -v -c cluster -c check
  every row ok, including
    Storage  ok  [ ceph(v) ceph_mon(v) ceph_mgr(v) ceph_mds(v) ceph_osd(v)
                   ceph_rgw(v) rbd_target(v) fc_link(v) ]

  live migration across the rolled cluster: cc1 -> cc2 -> cc3, both Complete,
  instance ACTIVE throughout.


6. CEPH-CSI 3.13.1 AGAINST REEF -- NO CHANGE NEEDED

[cc1 ~]# kubectl get pods -A | grep csi
  ceph-csi-cephfs-nodeplugin-bz6vg          2/2  Running
  ceph-csi-cephfs-provisioner-...           5/5  Running
  ceph-csi-rbd-nodeplugin-7grp8             2/2  Running
  ceph-csi-rbd-provisioner-...              7/7  Running
[cc1 ~]# kubectl get pvc
  csi-rbd-reef  Bound   ceph-rbd
  csi-fs-reef   Bound   ceph-fs
[cc1 ~]# kubectl exec csi-reef-test -- sh -c 'cat /mnt/rbd/t.txt /mnt/cephfs/t.txt; df -h ...'
  reef-rbd-1787999213
  reef-fs-1787999213
  /dev/rbd0                      973.4M  28.0K  957.4M  0%  /mnt/rbd
  10.1.0.1:6789:/volumes/cephfs_default_sub_volume_group/csi-vol-e70d9ca9-...
                                    1.0G      0    1.0G  0%  /mnt/cephfs


7. BUILD SIDE, IN THE heavy_rootfs CHROOT WITH REEF HEADERS

[jail]# chroot $R rpm -q librados-devel librbd-devel
  librados-devel-18.2.8-1.el9s.x86_64
  librbd-devel-18.2.8-1.el9s.x86_64
[jail]# chroot $R /opt/ceph/bin/pip list | grep -Ei 'cython|packaging|setuptools|wheel'
  Cython 0.29.37   packaging 26.3   setuptools 75.6.0   wheel 0.48.0
[jail]# ... pip wheel --no-build-isolation --no-deps -w /usr/src/ceph/wheels .
  Successfully built rados
  Successfully built rbd
  rados-2.0.0-cp311-cp311-linux_x86_64.whl
  rbd-2.0.0-cp311-cp311-linux_x86_64.whl
[jail]# chroot $R /opt/openstack-caracal/bin/python -c "import rados, rbd"
  caracal venv OK   rados.cpython-311-x86_64-linux-gnu.so

  hex_config compiles and links clean at -Wall -Werror:
[jail]# make -C core/modules config_ceph.o && make -C core/main hex_config
  CXX config_ceph.o
  LD  hex_config
[jail]# strings -a core/modules/config_ceph.o | grep -cF 'bluestore cache kv onode ratio = 0.04'
  1


8. THE heavyfs REPO CHANGE, PROVEN ON A NODE

  the broken dependency the old `rpm -e --nodeps` leaves:
[cc1 ~]# dnf check --dependencies
  Error: Check discovered 1 problem(s)
  centos-release-openstack-antelope-1-5.el9s.noarch has missing requires of
    centos-release-ceph-quincy

  with the new approach (keep the package, disable the repo):
[cc1 ~]# dnf -y install centos-release-ceph-quincy ; dnf config-manager --set-disabled centos-ceph-quincy
[cc1 ~]# dnf check --dependencies
                                                 <- clean
[cc1 ~]# dnf repolist | grep -i ceph
  centos-ceph-reef   CentOS-9-stream - Ceph Reef
[cc1 ~]# ceph versions | jq -r '.overall|keys[0]'
  ceph version 18.2.8 ... reef (stable)


9. LOGROTATE

[cc1 ~]# systemctl status logrotate.service
  Process: ExecStart=/usr/sbin/logrotate /etc/logrotate.conf (code=exited, status=1/FAILURE)
[cc1 ~]# logrotate -d /etc/logrotate.conf 2>&1 | grep ^error
  error: mysql:1 duplicate log entry for /var/log/mariadb/mysql_error.log
  error: found error in file mysql, skipping
[cc1 ~]# rpm -qf /etc/logrotate.d/mariadb
  MariaDB-server-10.11.18-1.el9.x86_64

  after removing the vendor file, on all three nodes:
[cc1 ~]# logrotate -d /etc/logrotate.conf 2>&1 | grep -c ^error
  0
[cc1 ~]# logrotate /etc/logrotate.conf ; echo rc=$?
  rc=0
```


Eandalf-Bigstack:
Thanks @traviswu-bigstack — both review points were right, and the first turned out to be worse than described. Addressed in 89614cee.

## 1. `WaitForMonQuorum` — removed outright

You were right that it can block the first of three control nodes waiting for nothing. It was also miscounted: each iteration costs `timeout 10 ceph -s` **plus** `sleep 5`, but the loop counter only advanced by the sleep — so `timeoutSec = 300` was really **up to 900 s**. Measured on accept-3cc, one no-quorum probe takes the full 10 s, so 60 iterations is 15 minutes on exactly the path it was least able to help: a cold poweroff, where cc1 commits before cc2/cc3 have booted and a 3-mon quorum cannot form yet.

Checking whether it was *strongly* needed: it isn't. It replaced no retry — it only front-loaded ones that already existed.

- `SetupPools()` carries its own `for i in 1 2 3 4 5` loop against `ceph -s`
- `ceph_osd_create_map` reads local disks via `ceph-volume raw list`, no cluster needed
- starting a systemd unit doesn't require quorum

So rather than guard it with a marker, it's gone. The documented order is unaffected — from the compiled binary on jim-1cc, mon → mgr is now 2 s instead of a wait:

```
02:53:15  Executing: systemctl start ceph-mon@cc1
02:53:15  Executing: /usr/sbin/hex_sdk ceph_unpause_client_io
02:53:17  Executing: systemctl start ceph-mgr@cc1
02:53:57  Executing: /usr/sbin/hex_sdk ceph_osd_restart 0
02:53:59  Executing: /usr/sbin/hex_sdk ceph_osd_restart 1
02:54:02  Executing: systemctl start ceph-mds@cc1
02:54:05  Executing: systemctl restart ceph-radosgw@rgw.cc1
```

## 2. `ceph_finalize_release` — yes, upgrade-only, now marker-gated

Confirmed it is only ever needed for an upgrade: `require_osd_release` changes when the release does and at no other time. It does **not** block a bootstrap — the first `$CEPH -s` is bounded by `SRVSTO=10` and returns — but it was paying a pointless round trip on every commit forever: **3 s** against a healthy cluster, **11 s** when no mon answers.

It now runs only while `/etc/appliance/state/ceph_release_upgrade_pending` exists, dropped by a new `CONFIG_MIGRATE(ceph, MigratePrepare)` hook on a firmware switch — the `config_mysql.cpp` `BACKDIR` pattern you pointed at.

The marker is cleared only when the release is **actually pinned**, not merely when the call ran, so the function grew a real contract: `rc 0` = pinned, `rc 1` = not yet (no cluster, mixed OSD tier, mons behind, or the pin failed). Mid-roll it returns 1, the marker survives, the next commit retries — whichever node finishes the roll last clears its own. It also now confirms the pin by re-reading the osdmap rather than trusting `require-osd-release`'s exit status. `ceph_leave_rolling` and `migrate_ceph` call it with `|| true` so a normal mid-roll "not yet" can't fail either caller.

## 3. Your question about the elevation — and a third bug it exposed

You were right about the mechanism, and I'd been unable to find it because `bootstrap_cube_config` is a name for the flow rather than a defined function. Tracing it properly:

```
systemd-hexctl-user.service   (enabled, WantedBy=multi-user.target)
  └─ ExecStart=/usr/sbin/bootstrap
       ├─ line 12/241 : /usr/sbin/hex_config bootstrap $BOOTSTRAP_MODULE
       └─ line 73/302 : hex_config -p bootstrap standalone-done
```

and `CommitCheck` short-circuits on `IsBootstrap()` — it sets `s_bConfigChanged = true` **unconditionally**, before looking at any tuning. So the boot after a firmware switch always takes the normal path, and for the upgrade reboot itself the elevation is indeed redundant.

But writing the test for it surfaced a **third** bug that neither of us had spotted: in my first revision `FinalizeCephRelease()` sat inside `if (s_bConfigChanged)` *and* behind `CommitCheck`'s early return. Mid-roll, that bootstrap commit finds the gate shut (the mons aren't all on the new release yet), so the marker survives — and the commit that eventually retries it need not be a bootstrap. Under the original placement it would have been skipped and the cluster could have run reef with `require_osd_release` still reading quincy, silently.

So the shape is now as you asked — return immediately on an undef role, no `doCommit` boolean threaded through, and the marker-gated finalize on `CommitCheck`'s false path:

```cpp
if (IsUndef(s_eCubeRole))
    return true;

if (!CommitCheck(modified, dryLevel)) {
    FinalizeCephRelease();
    return true;
}
```

Cost when there's no marker is one `access()`.

## What was tested

On jim-1cc, with the compiled binary, three cases:

| case | ceph module | finalize | marker |
|---|---|---|---|
| no marker | `commit(o) 57.8s` | **0 invocations** | stays absent |
| marker + a config change | `commit(o) 58.9s` | ran, `release finalized` | cleared |
| marker + a **true no-op** `commit(x)` | **2.4 s** (vs `0.0 s` unmarked) | ran | cleared |

That last row is the one the first revision got wrong. `HEALTH_OK` and `require_osd_release reef` throughout.

Return contract on accept-3cc: `rc 0` already-pinned (4 s), `rc 1` with no quorum (11 s), and `ceph_leave_rolling` still `rc 0` when finalize declines.

Both nodes and the jail checkout were restored to clean state afterwards.

core/modules/config_ceph.cpp
```diff
@@ -1513,9 +1569,21 @@ Commit(bool modified, int dryLevel)
...
+    // interrupted release upgrade. CommitCheck() is true for every bootstrap --
+    // it sets s_bConfigChanged unconditionally there -- so the boot after a
+    // firmware switch takes the normal path below. This branch is for the case
+    // that is left over: the gate was shut when that boot ran (mid-roll the mons
+    // are not all on the new release yet), so the marker survived and the next
+    // commit to come along has to retry it. That one need not be a bootstrap.
+    if (!CommitCheck(modified, dryLevel)) {
+        FinalizeCephRelease();
+        return true;
+    }
```
traviswu-bigstack: In what case !CommitCheck(modified, dryLevel) require FinalizeCephRelease()?

core/sdk_sh/modules/sdk_ceph.sh
```diff
@@ -454,14 +454,152 @@ ceph_leave_rolling()
...
+        [ $waited -lt $timeout ] || {
+            log_warning "ceph_wait_upgrade_gate: $tier not fully on $want after ${timeout}s (running: $(ceph_tier_releases "$tier" | tr '\n' ' '))"
+            return 1
+        }
+        sleep 10
+        waited=$((waited + 10))
+    done
+    return 0
+}
+
```
traviswu-bigstack: I dont see who are the caller of these functions

core/sdk_sh/modules/sdk_ceph.sh
```diff
@@ -454,14 +454,152 @@ ceph_leave_rolling()
...
-        if [ -n "$rel" ] && ! $CEPH osd dump 2>/dev/null | grep -q "require_osd_release $rel" ; then
-            Quiet $CEPH osd require-osd-release "$rel" --yes-i-really-mean-it
-        fi
-    fi
+    # post-upgrade finalization; see ceph_finalize_release, which is also called
+    # from the per-node ceph commit so a cluster upgraded without a roll job
+    # (a 1-node appliance, a hand-landed node) still completes the upgrade.
+    # `|| true` because rc 1 there means "not this node's turn", which is a normal
+    # outcome mid-roll and must not fail leaving rolling mode.
+    ceph_finalize_release || true
```
traviswu-bigstack: Since ceph_leave_roll() is on the path of rolling restart, like Commit() of config_ceph.cpp, is it better we place a short-circuit here to avoid unnecessary check

core/modules/config_ceph.cpp
```diff
@@ -1447,6 +1463,46 @@ Validate()
...
+//
+// The marker is cleared only when ceph_finalize_release reports the release actually
+// pinned -- rc 0 -- not merely that it ran. Mid-roll the gate is shut and it returns
+// 1, so the marker survives and the next commit tries again; whichever node finishes
+// the roll last is the one that clears its own. Same shape as config_mysql.cpp's
+// BACKDIR, which also survives until the upgrade it represents has succeeded.
+static void
+FinalizeCephRelease()
+{
+    if (!IsControl(s_eCubeRole))
```
traviswu-bigstack: why not !IsMaster()

Eandalf-Bigstack: We would need all control nodes to be ready in Ceph Reef to unlock the tier of mon/mgr/osd/mds/rgw into Reef. This is similar to MariaDB and RabbitMQ, to have a version uniformality check among all control nodes, so we would need to run it on each control node to check if peers are ready to march into the new version.

core/sdk_sh/modules/sdk_ceph.sh
Eandalf-Bigstack: We would still need to check since this ceph_leave_roll would be run on control node power roll events. However, to finalize release, we would need all control nodes ready, so we could not directly issue the finalization order on any control nodes.

core/modules/config_ceph.cpp
Eandalf-Bigstack: This is mostly used to clean up the stale Ceph upgrade file marker `/etc/appliance/state/ceph_release_upgrade_pending`. For most of the time, this marker would be cleared on the third control node. While at the same time, we would have the file marker left on control node 1 and 2. The file marker could be cleaned up on control node 1 and 2 on a hex_config run.

core/sdk_sh/modules/sdk_ceph.sh
Eandalf-Bigstack: Yes, apparently, they are not referenced anywhere, I would look into it.

core/modules/config_ceph.cpp
traviswu-bigstack: but what ceph_finalize_release() does is only make sure osd not include other services mon/mgr/osd/mds/rgw

core/sdk_sh/modules/sdk_ceph.sh
Eandalf-Bigstack: Thank you for catching that. The dead code was removed in [f4beac7](https://github.com/bigstack-oss/cubecos/pull/1387/commits/f4beac7e9145dfe7a2a0fcd21300bd1ead7cf8dd). Those were used as probing functions for watching the ceph rolling events while testings during the development. Now, they were removed to clean up.

core/modules/config_ceph.cpp
traviswu-bigstack: Like we discussed before, this is for edge case handling, but the !commitcheck is called almost every time we make change via hex_config. that said, 99% of time of calling FinalizeCephRelease(); is meaningless.

core/sdk_sh/modules/sdk_ceph.sh
traviswu-bigstack: same reason as above, what's the point of this line when the cluster is not in a ceph migration path, like rolling restart. the work here is mostly pointless. that's why i would suggest adding a short circuit to bypass the finalization as soon as possible

core/modules/config_ceph.cpp
Eandalf-Bigstack: Got it. Would drop it here.

core/sdk_sh/modules/sdk_ceph.sh
Eandalf-Bigstack: We would still need to support cluster powercycle as an upgrade path. For cluster powercycle upgrade, we would still need to have the upgrade migration logics in the bootstrapping to cover the upgrade.

core/modules/config_ceph.cpp
Eandalf-Bigstack: Yes, that function is mostly pulling the gate for OSD when the MON is ready. Master would be the first one ready, but OSDs from the last Ceph release would need all MON ready, which means control 2 and control 3 would need to be ready. However, if we dedicate the call for master, we would never have the chance to have all MON ready and be able to call the function at the same time.

traviswu-bigstack: got it

core/sdk_sh/modules/sdk_ceph.sh
Eandalf-Bigstack: Fixed in [ed34c70](https://github.com/bigstack-oss/cubecos/pull/1387/commits/ed34c70039f694d82ec9bff095d6a062df7cd606).

core/modules/config_ceph.cpp
Eandalf-Bigstack: Fixed in [ed34c70](https://github.com/bigstack-oss/cubecos/pull/1387/commits/ed34c70039f694d82ec9bff095d6a062df7cd606).

traviswu-bigstack (approved):

PR #1401, opened 2026-09-01
fix(advisor): fold the cluster id the way the Advisor stores it

traviswu-bigstack:
#### What type of PR is this?

/kind bug

`cubesys.controller` is declared with `ValidateRegex` and `DFT_REGEX_STR` (`^.*$`), so CubeCOS constrains it **not at all** — its only de-facto shape comes from `UpdateHosts` writing it into `/etc/hosts`. A cluster whose controller is named `Controller01` is perfectly legal, and anything downstream that refuses it is imposing a rule on a value it does not own.

The Advisor now folds the id to lowercase and stores the folded form (bigstack-oss/cube-ai-advisor#129), accepting lowercase letters, digits, `.`, `-` and `_`, with alphanumeric ends and a 64-character limit. This folds the same way here, so **the id this node reports is the id the Advisor will hold**, and refuses only shapes that genuinely break a certificate CN, a URL path segment or `/etc/hosts` — refused locally, where the message can name the rule, instead of arriving as a remote error after a round trip.

Accepts `ky3haclust01` (the driver's UUID-derived id), `sky-140`, `cube.lab.01`, `foo_bar`, and folds `Controller01` → `controller01`, `ACME-PROD-01` → `acme-prod-01`. Refuses `a/b`, `has space`, `_leading`, `trailing_`, `-leading`, `.dot`, and anything over 64 characters.

None — follow-on to bigstack-oss/cube-ai-advisor#108 and #129, whose rule this mirrors.

- **This PR reversed direction.** Its first revision narrowed the local check to the Advisor's *pre-#129* rule, which would have locked out exactly the clusters it now admits. Travis caught the premise: the SaaS should honour what CubeCOS produces, not the other way round. The Advisor widened; this now mirrors the wider rule.
- Tests cover **both** failure directions — dropping the fold fails two cases (`Controller01`, `ACME-PROD-01`), and restoring the narrower charset fails another (`foo_bar`). Both were run before being restored, so the cases pin the behaviour rather than describing it.
- `CLUSTER_ENV` is a testability seam with a production default; nothing sets it outside the unit test.
- `bash test_sdk_advisor_verify.sh` → pass=27 fail=0.

#### Additional documentation

None; the reasoning is in the function's comment.

### DIFF
--- a/core/sdk_sh/modules/sdk_advisor.sh
+++ b/core/sdk_sh/modules/sdk_advisor.sh
@@ -147,8 +147,11 @@ advisor_cluster_id()
 {
     local id=""
 
-    if [ -r /etc/cube/phone-home-agent.env ] ; then
-        id=$(sed -n 's/^CUBE_CLUSTER_ID=//p' /etc/cube/phone-home-agent.env | head -1)
+    # CLUSTER_ENV is overridable so the unit test can supply a file; nothing
+    # in production sets it.
+    local env_file=${CLUSTER_ENV:-/etc/cube/phone-home-agent.env}
+    if [ -r "$env_file" ] ; then
+        id=$(sed -n 's/^CUBE_CLUSTER_ID=//p' "$env_file" | head -1)
     fi
     if [ -z "$id" ] ; then
         id=$(source /usr/sbin/hex_tuning /etc/settings.txt 2>/dev/null ; echo "$T_cubesys_controller")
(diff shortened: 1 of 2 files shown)


traviswu-bigstack:
Holding this in draft — the direction is wrong and will be rewritten.

`cubesys.controller` is declared with `ValidateRegex` + `DFT_REGEX_STR` (`^.*$`), so CubeCOS constrains it not at all; its only de-facto shape comes from `UpdateHosts` writing it into `/etc/hosts`. This PR narrowed the node-side check to match a rule the Advisor had imposed on a value CubeCOS owns — which would lock out a legitimate cluster whose operator set, say, `Controller01`. Hostnames are case-insensitive; refusing on case is spurious.

The Advisor is widening instead (lowercase-fold on input, accept `_`, keep refusing only what genuinely breaks a CN, a path segment or /etc/hosts). This PR will be rewritten to mirror that wider rule, so the node still gives an early, local error for the shapes that really are broken.

traviswu-bigstack:
Closing — Travis is right that this should not exist.

The node-side check catches nothing the SaaS does not already catch. `IssueToken` validates when the operator creates the cluster and issues the token, before a node is involved; `Redeem` folds the CSR's CommonName and compares it to the token's cluster, and admission folds too. The node's id must *match* the cluster the token was issued for, so validity is subsumed by that match — a wrong-shaped id fails as a mismatch regardless.

What it costs is real: a duplicated copy of a rule the Advisor owns, on values that are either machine-generated (`CUBE_CLUSTER_ID`) or hostname-shaped (`cubesys.controller`). It had already drifted within a day — this PR's first revision narrowed the node to the Advisor's pre-#129 rule, and would then have locked out exactly the clusters the Advisor had just decided to accept. A second enforcement point for someone else's rule reproduces that every time the rule moves.

develop keeps the pre-existing guard for shapes that are broken under any rule (slash, whitespace, over 64 characters) and the 'cannot tell which cluster this node belongs to' error, which is genuinely local knowledge. Neither duplicates a decision the Advisor makes.

Issue #643, opened 2026-01-27
[Story] [Parallel Group] Upgrade Watcher from Antelope to Caracal


jdjgya:
## Description

As a CubeCOS operator, I want Watcher be in Caracal so that we could upgrade to Caracal.

Issues labeled with "Parallel Group" could be tasked in parallel.

## Acceptance Criteria

> A detailed list of conditions that must be met for the story to be accepted. These statements must focus on functional behavior and constraints rather than specific code-level implementation.

- [ ] Migrate configs
- [ ] Pass health check
- [ ] Ignore the new features introduced in the new version

## Story Points: 3

>| Points | Effort / Resources | Description |
> | -- | -- | -- |
>| 1 | Low | Low complexity; can be resolved easily.
>| 3 | Medium | Moderate workload; requires some focus, but risks are well-managed.
>| 5 | High | High complexity; difficult to implement, requiring significant time and effort.
>| 8 | Very High | Extreme complexity or high uncertainty. If possible, this should be divided into multiple user stories.

## QA Verification

> Rows below are **checkable claims**: the exact command plus the expected outcome. Run them
> **from a control node as root**, with `source /etc/admin-openrc.sh` unless a row says otherwise.
> Shipped by cubecos **PR #1395** (merge `4efadb19`, fast-forward merged to `develop` on
> 2026-09-01), a single commit — `core/watcher/watcher.mk`, the `antelope_patch/` →
> `caracal_patch/` re-derivation of three upstream overrides, a regenerated
> `watcher.conf.sample`, and `project.mk`'s occupant list. **`config_watcher.cpp` needed no
> change**, which is itself part of what §1 checks.
>
> **Scope of this section.** A `User Story` (3 points): `python-watcher` hops from 10.0.0 in
> the python 3.10 antelope venv to **12.1.0 in the python 3.11 caracal venv**, the twelfth
> service of #594 to make the hop. Sections below are organised against this ticket's three
> Acceptance Criteria, so it is visible which rows prove which AC. Other services' caracal hops
> are their own tickets.
>
> **§3 is the row that decides the verdict** — not because the new feature is risky, but because
> `allocation_balance` is a CubeCOS-only strategy carried as a **whole-file override**, and a
> whole-file override is the one patch shape that can silently revert upstream's own additions.
> If §3 is wrong, the cluster loses a strategy while every health check stays green.

### The two things to hold in your head before testing

[...]

Issue #639, opened 2026-01-27
[Story] [Parallel Group] Upgrade Masakari from Antelope to Caracal


jdjgya:
## Description

As a CubeCOS operator, I want Masakari be in Caracal so that we could upgrade to Caracal.

Issues labeled with "Parallel Group" could be tasked in parallel.

## Acceptance Criteria

> A detailed list of conditions that must be met for the story to be accepted. These statements must focus on functional behavior and constraints rather than specific code-level implementation.

- [ ] Migrate configs
- [ ] Pass health check
- [ ] Ignore the new features introduced in the new version

## Story Points: 3

>| Points | Effort / Resources | Description |
> | -- | -- | -- |
>| 1 | Low | Low complexity; can be resolved easily.
>| 3 | Medium | Moderate workload; requires some focus, but risks are well-managed.
>| 5 | High | High complexity; difficult to implement, requiring significant time and effort.
>| 8 | Very High | Extreme complexity or high uncertainty. If possible, this should be divided into multiple user stories.

## QA Verification

> Rows below are **checkable claims**: the exact command plus the expected outcome. Run them
> **from a control node as root**, with `source /etc/admin-openrc.sh` unless a row says otherwise.
> Shipped by cubecos **PR #1394** (merge `4fb50b43`, fast-forward merged to `develop` on
> 2026-09-01), 5 commits over 37 files — `core/masakari/*` (`.mk`, two generated
> `.conf.sample`s and the `oslo-config-generator` inputs, `masakari_sudoers`, the
> `antelope_patch/` → `caracal_patch/` split), `core/modules/config_masakari.cpp`,
> `core/nova/nova.mk`, and eight comment-only files.
>
> **Scope of this section.** A `User Story` (3 points): masakari hops from 15.0.x in
> `/opt/openstack-antelope` to **17.0.0 in `/opt/openstack-caracal`**, and masakari-monitors to
> **17.0.1**, joining the twelve services already there. Sections below are organised against
> this ticket's three Acceptance Criteria, so it is visible which rows prove which AC. Other
> services' caracal hops are their own tickets.
>
> **§4 is the section that decides the verdict.** This PR deleted `/usr/bin/privsep-helper`,
> which was shared by sixteen `helper_command` values across seven services. If §4 is wrong,
> the damage is to cinder / glance / neutron / nova / manila / cyborg — not to masakari, and no
> other row here detects it.

### The one thing to hold in your head before testing

[...]

Issue #58, opened 2025-05-01
[Feature] License - To have a new field `name` in the license file and the related functions


jdjgya:
**Is your feature request related to a problem? Please describe.**

COS 3.0.0 needs the license name to show on UI,

but current license mechanism has not included the license `name` and hex related features

<br/>

From the existing use case, 

the value of `name` is probably from

- the license's filename

- or set by license manager during license generating.

<br/>

---

<br/>

**Describe the solution you'd like**

probably might be 2 solutions

1). Add a new field - `name` in the current license mechanism including the license info and the related features.

2). or if the effort of (1) is huge, probably we could try to make a convention first like `name = current field A + current field B` ... and so on, so API can use it to generate the name for the time being

<br/>

---

<br/>

**Describe alternatives you've considered**

<br/>

---

<br/>

**Additional context**

PR #1404, opened 2026-09-01
feat(horizon): upgrade horizon from openstack antelope to caracal

### LINKED ISSUE #636: [Story] Upgrade Horizon from Antelope to Caracal
## Description

As a CubeCOS operator, I want Horizon be in Caracal so that we could upgrade to Caracal.

## Acceptance Criteria

- [ ] Migrate configs
  - [ ] Ensure we could log in the Horizon dashboard.
  - [ ] Ensure we could perform volume operations on the dashboard.
  - [ ] Ensure we could perform network operations on the dashboard.
  - [ ] Ensure we could perform VM operations on the dashboard.
  - [ ] Ensure other panels are functioning on the dashboard.

## Story Points: 3

> | -- | -- | -- |

## QA Verification

> Rows below are **checkable claims**: the exact command plus the expected outcome. Unless a row
> says otherwise, run the CLI ones **from a control node as root** with
> `source /etc/admin-openrc.sh`, and the browser ones against `https://<shared-ip>/horizon/`
> (`WEBROOT` is `/horizon/`, `local_settings.in:481`).
> Shipped by cubecos **PR #1404** (fast-forward merged to `develop` on 2026-09-02) —
> `core/horizon/*` (`.mk`, `local_settings.in`, the unit, `openstack-dashboard.conf`,
> `gunicorn-config.py`), the masakari-dashboard patch pair, `core/main/proj_functions`, and nine
> `.mk` files whose parked-in-antelope notes are now past tense.
>
> **Scope of this section.** A `User Story` (3 points): **horizon 23.1.1 → 24.0.2 (django 3.2 →
> 4.2)** into `/opt/openstack-caracal`, its eight dashboard plugins pinned to their 2024.1
> releases, and **`/usr/bin/openstack` 6.2.1 → 6.6.1** moving with them along with the eight osc
> plugin clients that were only in antelope because the CLI was. Sections below are organised
> against this ticket's three Acceptance Criteria, so it is visible which rows prove which AC.
>
> **§2 is why this ticket is `Ready to QA` rather than `Not needed`.** Dev's sweep proves every
> panel *loads* (23 authenticated paths, zero 500s). It does **not** prove a create/delete
> actually succeeds through the dashboard, and it did not touch **Cube Account (Keycloak) login**
> at all — the sweep authenticated with `auth_type=credentials`, which is not the form's default.
> Those two gaps are the load-bearing part of this section.

Three things moved at once and they had to: the dashboard follows its plugins, and every osc
plugin client was pinned to whichever venv the CLI ran in. After this ticket **no OpenStack
service is left in the antelope venv** — only monasca, ospurge and ceph's bindings.

Two mechanical changes are easy to mistake for unrelated breakage if you hit them:

- **Static assets are now served by httpd directly** (`Alias /horizon/static`, with
  `ProxyPass /horizon/static !` at `openstack-dashboard.conf:31` ordered *above* the
  `ProxyPass /horizon` at line 41). Apache honours the first matching ProxyPass, so that ordering
  is load-bearing — get it wrong and every asset 404s while pages still render.
- **collectstatic and compress moved to `ExecStartPre`**, so a restart does real work before the
  socket is up. It completes in ~8s against `TimeoutStartSec=5min`; a restart that appears to hang
  for a few seconds is that, not a failure.

And one latent bug this hop forced into the open: `ALLOWED_HOSTS` had always been the bare string
`'*'`. Django 4.0 added it to `tuple_settings`, so under 4.2 that raises
`ImproperlyConfigured: The ALLOWED_HOSTS setting must be a list or a tuple` on **every** settings
load. It worked by accident under 3.2. It is now `['*']`.

### Preconditions

- A CubeCOS cluster installed from an ISO that contains PR #1404. Verify before starting:
  `/opt/openstack-caracal/bin/pip show horizon \| head -2` must report `24.0.2`. **On a build
  without this PR every row below either fails or passes for the wrong reason — check this first.**
- A browser that can reach the cluster VIP over https, plus admin credentials.
- **Cube Account (Keycloak) must be configured** for §2.1 row 3. On a cluster where it is not, that
  row is not applicable — say so in the result rather than marking it passed.
- §2.2-§2.4 create and delete a volume, a network and an instance. All are removed in **Cleanup**;
  run it even if a row fails.

### 1 · Migrate configs

|---|---|---|
| 1 | `/opt/openstack-caracal/bin/pip show horizon \| head -2` | `Name: horizon` / `Version: 24.0.2`, under `python3.11/site-packages` |
| 2 | `grep ALLOWED_HOSTS /etc/openstack-dashboard/local_settings` | `ALLOWED_HOSTS = ['*']` — a **list**. This file is regenerated from `.in` on every `Commit()`, so it is the value that actually reaches django |
| 3 | `systemctl is-active openstack-dashboard` | `active`. This is itself the runtime proof of row 2: django 4.2 refuses to start on the old string |
| 4 | `ps -ef \| grep '[g]unicorn' \| head -3` | master + 4 workers, running under `/usr/local/bin/python3.11` |
| 5 | `/opt/openstack-caracal/bin/pip list \| grep -Ei 'heat-dashboard\|manila-ui\|octavia-dashboard\|ironic-ui\|designate-dashboard\|masakari-dashboard\|watcher-dashboard\|neutron-vpnaas-dashboard'` | eight rows: `11.0.0`, `11.0.1`, `13.0.1`, `6.3.0`, `18.0.0`, `10.0.0`, `11.0.0`, `10.0.0` — **no `0.0.0`**, which is what a leftover `setup.py`-installed git checkout reports |
| 6 | `readlink -f /usr/bin/openstack; openstack --version` | `/opt/openstack-caracal/bin/openstack`, version `6.6.1` |
| 7 | `/opt/openstack-antelope/bin/pip list \| grep -Ei 'horizon\|dashboard\|openstackclient\|^python-(heat\|manila\|octavia\|designate\|masakari\|watcher\|cyborg\|barbican)client'` | **empty** — no services, no dashboards, no clients left behind |
| 8 | `/opt/openstack-caracal/bin/python3 -c "from importlib.metadata import entry_points; print(sorted(e.name for e in entry_points(group='openstack.cli.extension')))"` | all nine: `accelerator`, `baremetal`, `dns`, `ha`, `infra_optim`, `key_manager`, `load_balancer`, `orchestration`, `share`. A missing one is how `DNSaaS NG [ designate(9 api not all up) ]` was once produced with every designate unit up |
| 9 | `openstack segment list` in a shell **without** credentials | `Missing value auth-url` — i.e. the subcommand resolved. `openstack: 'segment' is not an openstack command` means the plugin did not |
| 10 | `ls /etc/openstack-dashboard/policies/*.yaml \| wc -l` | ≥7 — keystone / nova / cinder / glance / neutron / octavia / masakari, each non-empty (`dump_default_policies` ran against every namespace) |

### 2 · Adjust service arrangements to fulfil basic functionalities

Dev's evidence for this AC is a sweep of 23 authenticated paths, all 200. **A 200 means the panel
renders, not that the operation works** — these rows are the part that has not been done.

#### 2.1 · Ensure we could log in the Horizon dashboard

|---|---|---|
| 1 | `curl -skI https://<shared-ip>/horizon/` | `302` to `/horizon/auth/login/?next=/horizon/` — the prefix is preserved through the proxy. A redirect that has lost `/horizon` means `SCRIPT_NAME` is not reaching gunicorn |
| 2 | In a browser: log in with **Keystone credentials** (admin) | lands on `/horizon/project/`, sidebar renders Project / Admin / Identity |
| 3 | Log out, then log in with **Cube Account** (Keycloak, the form's default `auth_type`) | lands on the same place. **Dev did not test this path** — it is the one most likely to differ |
| 4 | `grep -c IndexError /var/log/horizon/horizon.log` | `0` — a stripped `SCRIPT_NAME` produces one per request |
| 5 | View source on the login page, note an asset URL, `curl -skI` it | `200`, and the filename is hashed (e.g. `output.b597bc3299cf.css`) — offline compression's manifest is in use |
| 6 | `grep -c ' 404 ' /var/log/httpd/*access*` (during the session above) | `0` for `/horizon/static/…` — the `ProxyPass … !` ordering holds |

#### 2.2 · Ensure we could perform volume operations on the dashboard

|---|---|---|
| 1 | Project → Volumes → **Create Volume** (`qa-636-vol`, 1 GiB) | reaches `Available`; no red toast |
| 2 | `openstack volume show qa-636-vol -f value -c status` | `available` — the CLI agrees with what the panel showed |
| 3 | Volumes → **Create Snapshot** on it, then delete the snapshot | both complete from the panel |
| 4 | Admin → Volumes | the volume is listed with its project |

#### 2.3 · Ensure we could perform network operations on the dashboard

|---|---|---|
| 1 | Project → Networks → **Create Network** (`qa-636-net`) with a subnet `192.0.2.0/24` | created, both network and subnet visible |
| 2 | Project → Routers → **Create Router**, attach the subnet as an interface | interface shows `Active` |
| 3 | `openstack network show qa-636-net -f value -c status` | `ACTIVE` |
| 4 | Project → Security Groups → add a rule, then delete it | both operations succeed from the panel |

#### 2.4 · Ensure we could perform VM operations on the dashboard

|---|---|---|
| 1 | Project → Instances → **Launch Instance** onto `qa-636-net` | reaches `Active` |
| 2 | Instance → **Console** tab | the noVNC console attaches and paints a screen |
| 3 | Instance → Attach the volume from §2.2, then detach it | both complete; `openstack server show` agrees |
| 4 | Instance → **Soft reboot**, then **Delete** | returns to `Active`, then disappears from the list |

#### 2.5 · Ensure other panels are functioning on the dashboard

|---|---|---|
| 1 | Visit each plugin panel: Stacks, Shares, Load Balancers, DNS Zones, Reverse DNS, VPN, Bare Metal, Goals, Audits, Action Plans, Instance-HA | every one renders, **zero 500s**. These are the eight plugins that were repinned |
| 2 | `/masakaridashboard/` → Segments / Hosts / Notifications / VMoves | all render. A `NoReverseMatch` here takes down **every** page, not one panel — that is the failure the carried patch exists to prevent |
| 3 | Identity → Projects / Users, and Settings | render |
| 4 | `grep -c 'Internal Server Error\|Traceback' /var/log/horizon/horizon.log` after the sweep | `0` |

### 3 · Ignore the new features introduced in the new version

|---|---|---|
| 1 | `grep default_panel /opt/openstack-caracal/lib/python3.11/site-packages/masakaridashboard/dashboard.py` | `default_panel = 'segments'` — the carried patch was re-derived against 10.0.0, whose context had shifted (`vmoves` added two lines above, `ugettext_lazy` → `gettext_lazy`). Upstream's `'default'` names a panel with no index, and the sidebar 500s on it |
| 2 | `grep -n "'vmoves'" …/masakaridashboard/dashboard.py` | present — **the one addition taken deliberately**, and only because its API already shipped with masakari 17.0.0 in #639. The panel was the missing half, not new scope |
| 3 | `/opt/openstack-caracal/bin/pip show heat-dashboard \| head -2` | `11.0.0` — **not 11.0.1**, which `heat.mk`'s comment used to name and which does not exist on PyPI |
| 4 | `ls /etc/openstack-dashboard/enabled/ \| wc -l` | unchanged from before the hop — the 18 `cp -f` registration lines were verified against the caracal wheels' file sets, so no glob should have started missing |

### 4 · The failure mode this change introduces — test it deliberately

The service used to run under `WSGIScriptAlias` with mod_wsgi. It now runs as **gunicorn behind a
unix socket**, with static files served by httpd and `collectstatic`/`compress` in `ExecStartPre`.
Each of those is a new way to be wrong that no panel check detects.

|---|---|---|
| 1 | `systemctl restart openstack-dashboard; systemd-analyze blame \| grep openstack-dashboard` | `ExecStartPre` completes well inside `TimeoutStartSec=5min` (~8s measured); the unit reaches `active` |
| 2 | Immediately after: log in and load `/horizon/project/` | 200 within ~10s — a restart must not leave the dashboard serving stale or missing assets |
| 3 | `grep ALLOWED_HOSTS /etc/openstack-dashboard/local_settings.in` | `['*']` — the **template** carries the list, so the next `Commit()` regenerates `local_settings` (§1 row 2) correctly rather than reintroducing the bare string. Fixing only the generated file would survive until the next config change and then take the dashboard down |
| 4 | `sed -n '28,45p' /etc/httpd/conf.d/openstack-dashboard.conf` | `ProxyPass /horizon/static !` appears **before** `ProxyPass /horizon` |

1. **The systemd unit is `openstack-dashboard`, not `horizon`.**
2. **`python-monascaclient` deliberately gets no caracal line.** It registers no
   `openstack.cli.extension`, only the standalone `monasca` script, which stays pointed at the
   antelope venv alongside the monasca service that is not moving. Monasca is removed outright in
   #672.
3. **`python-ironicclient` is not named explicitly.** ironic-ui and python-watcher both declare it
   and both are now in the caracal venv — the same two-declarer guarantee `ironic.mk` relied on.
4. **The antelope venv still exists** and still holds monasca, ospurge and ceph's bindings. That is
   what `OPENSTACK_RELEASE` is now blocked on; it is not leftover from this ticket.
5. **`cube_edge_cfg()` only runs on an edge deployment.** On a normal cluster the designate/watcher
   panel-move path is simply not exercised.

### Cleanup

```bash
source /etc/admin-openrc.sh
openstack server delete qa-636-vm            # if §2.4 left it behind
openstack volume delete qa-636-vol
openstack router remove subnet qa-636-router qa-636-subnet && openstack router delete qa-636-router
openstack network delete qa-636-net          # deletes its subnet with it
```

### Already verified on the 1cc — and what that did not cover

Dev verified §1, §3, §4 and the *load* half of §2 on a 1cc VM (`cc1`, 10.1.0.1) built from ISO
`CUBE_3.1.20_20260901-0822_c0bd078` — build #58, `distclean iso test`, `RC_BASE=v3.1.20`,
`E2ETST=1cc`, green end to end (`build` / `hex` / `e2e` / `publish` all `SUCCESS`). The sweep
covered 23 authenticated paths including every plugin panel, plus the login POST, the static-asset
404 count, the `SCRIPT_NAME` prefix, offline compression, all nine osc plugins, and a restart
round trip.

**What that did not cover — and what the rows above are therefore load-bearing for:**

- **Any create/delete performed through the dashboard.** The sweep is `GET`-only. §2.2-§2.4 are
  entirely new ground.
- **Cube Account (Keycloak SAML) login.** The sweep authenticated with `auth_type=credentials`;
  Keycloak is the form's default and was never exercised. §2.1 row 3.
- **`cube_edge_cfg()`**, which reruns collectstatic and compress after moving the designate and
  watcher panels aside — edge deployments only.
- **A multi-control cluster**, where the dashboard sits behind the VIP with more than one gunicorn
  master.

- [x] **QA Status declared at Done** — this is a `User Story`, which carries a `QA Status`, and the
  one — **do not reopen this ticket**. §2.1 row 3 and §2.2-§2.4 are where that is most likely.

SekiXu:
#### What type of PR is this?

/kind feature

horizon was the last thing serving out of `/opt/openstack-antelope` that a user sees, and `/usr/bin/openstack` was the last thing holding other services to that venv after they had already moved away from it. Both hop here, and they have to hop together: the dashboard stays where its plugins are, and every osc plugin client was pinned to whichever venv the cli ran in.

**horizon 23.1.1 → 24.0.2, which is django 3.2 → 4.2.** The caracal venv already held a 24.0.2 copy — a down payment taken in #643 so that `dump_default_policies` could run under the interpreter that owns the caracal services' `oslo.policy` entry points. Same version, so the dependency set landed then and this install replaces the antelope one rather than adding to it. The two policy-namespace lists collapse back into one for the same reason: the antelope list has been empty since octavia and masakari hopped, and its loop was a no-op.

The eight dashboard plugins, all pinned to their 2024.1 releases:

| Plugin | From | To |
| -- | -- | -- |
| heat-dashboard | 9.0.0 | 11.0.0 |
| manila-ui | 9.0.1 | 11.0.1 |
| octavia-dashboard | 11.0.1 | 13.0.1 |
| ironic-ui | 6.1.0 | 6.3.0 |
| designate-dashboard | git clone | 18.0.0 |
| masakari-dashboard | git clone | 10.0.0 |
| watcher-dashboard | git clone | 11.0.0 |
| neutron-vpnaas-dashboard | git clone | 10.0.0 |

`heat.mk` named 11.0.1 as the version to move to. **There is no 11.0.1** — heat-dashboard's caracal series is 11.0.0 alone and PyPI has no such release — so the pin is 11.0.0 and the comment is corrected with it.

**The four clones had to become pins anyway.** Their version was whatever `unmaintained/2023.1` pointed at on build day, and watcher-dashboard's was worse still: a `2023.1-eol` *tag*, because that project deleted both of its 2023.1 branches. Nothing reads `$(OPS_GITHUB_BRANCH_02)` for a dashboard now, and the build directories the clones needed go with them.

**The 18 `cp -f` lines that register panels, settings snippets and policy files are untouched.** Every filename they name was checked against the caracal wheel of all eight plugins: the `enabled/`, `local_settings.d/` and `conf/` file sets are identical across the hop, so none of the globs miss.

**masakari-dashboard's carried patch is re-derived rather than moved.** `default_panel = 'default'` names a panel whose `urls.py` has no index, so the sidebar raises `NoReverseMatch` and every page 500s; upstream has not fixed it at 10.0.0. But 10.0.0 adds `'vmoves'` to the panels tuple two lines above the change and renames `ugettext_lazy` to `gettext_lazy`, so the 8.0.0 hunk's context no longer applies. `antelope_patch/` is gone and masakari's patch pairs are one set again. 10.0.0 also brings the vmoves panel itself — the API behind it arrived with masakari 17.0.0 in #639, only the panel was missing.

**python-openstackclient 6.2.1 → 6.6.1**, and with it the eight clients that were sitting in antelope for no reason other than the cli's location, each already carrying a comment saying it could go once the cli moved:

| Client | osc plugin | Also installs |
| -- | -- | -- |
| python-heatclient | `orchestration` | `/usr/bin/heat` |
| python-manilaclient | `share` | (the antelope duplicate is deleted) |
| python-octaviaclient | `loadbalancer` | |
| python-designateclient | `dns` | |
| python-masakariclient | `ha` | |
| python-watcherclient | `optimize` | `/usr/bin/watcher` |
| python-cyborgclient | `accelerator` | `/usr/bin/cyborg` |
| python-barbicanclient | `secret` | |

Every one is **named explicitly** on the caracal side rather than left to whatever pulls it in transitively. `designate.mk` carries the story of losing the `dns` plugin exactly that way, which is what `cluster check` reported as `DNSaaS NG [ designate(9 api not all up) ]` while every designate unit was up.

`python-monascaclient` gets **no** caracal line. It registers no `openstack.cli.extension` at all, only the standalone `monasca` console script (`metric` comes from aodhclient and gnocchiclient), so the cli move leaves nothing behind: `/usr/bin/monasca` stays pointed at the antelope venv alongside the monasca service that is not moving, and `hex_sdk`'s only caller — `sdk_os.sh`'s `monasca metric-name-list` — keeps working unchanged. Monasca is being removed outright in #672 anyway. `python-ironicclient` needs no line either: ironic-ui and python-watcher both declare it and both are now in the caracal venv, the same two-declarer guarantee `ironic.mk` relied on in antelope.

**Three things run `manage.py` and all three follow the interpreter:** `openstack-dashboard.service`, `config_horizon.cpp`'s migrate, and `cube_edge_cfg()` in `core/main/proj_functions`, which reruns collectstatic and compress after it moves the designate and watcher panels aside on an edge deployment. gunicorn is now named explicitly too — the copy the dashboard had been using came from `core/monasca`, which is not moving.

**One ordering fix:** the heavyfs install block moves below the caracal venv bootstrap. Double-colon rules run in file order, and it used to sit above it — installing into `$(CARACAL_OPENSTACK_HOME_DIR)` before anything had created it.

**`ALLOWED_HOSTS` had to become a list.** Django added it to the `tuple_settings` check in 4.0, so the bare string `local_settings.in` has always carried now raises `ImproperlyConfigured: The ALLOWED_HOSTS setting must be a list or a tuple` on every settings load under django 4.2. Under 3.2 nothing checked the type, and `validate_host()` iterating a one-character string yields the same `'*'` pattern a one-element list does — so it worked by accident, and the comment two lines above already said "list". This fails the *build*, not just the dashboard: the `dump_default_policies` loop loads `openstack_dashboard.settings`, and a settings module that will not load leaves django's command registry holding only the core commands, so the loop dies as `KeyError: 'dump_default_policies'` before it reaches a namespace.

**Nine `.mk` files carried notes** saying their dashboard plugin or osc client was parked in antelope until horizon and the cli moved. They have, so the notes are rewritten in the past tense — the constraint that put each piece there is worth keeping, the claim that it is still in force is not. The load-bearing ones: `heavyfs/Makefile`'s setuptools ceiling (converting the four dashboards to pinned pip removed the last `setup.py install` caller in the tree; the ceiling is kept so a new one fails on review), `ironic.mk`'s two-declarer argument (pointed at the wrong venv), `manila.mk`'s second-consumer claim (manila-ui hopped too), and `project.mk`'s occupant list — whose "leaves a piece behind" paragraph is replaced with what is actually left in antelope: **monasca, ospurge and ceph's bindings. No service**, so `OPENSTACK_RELEASE` is now blocked on those three rather than on a hop. `project.mk` also gains masakari, which #639 moved without adding itself to that list.

**On "ignore the new features".** The one behavioural addition taken deliberately is masakari-dashboard's vmoves panel, and only because its API already shipped in #639 — the panel was the missing half, not a new feature. Nothing else in the eight plugins' 2024.1 releases is enabled beyond what the existing `cp -f` registration already covered.

#### Verification

ISO `CUBE_3.1.20_20260901-0822_c0bd078` — build **#58**, `distclean iso test`, `RC_BASE=v3.1.20`, `E2ETST=1cc`. Green end to end, 3h27m: `build` 10631s, `hex` (sys/iso) 692s, `e2e` 1093s, `publish` all `SUCCESS`.

> That ISO was built from `c0bd078f`, which is this branch plus one commit since dropped — a change to the source-side prerequisites of the `$(HEAVYFS)` rule in `core/heavyfs/Makefile`. It belongs to a separate build-performance story and is parked on its own branch. It cannot affect this ISO: `distclean` builds from an empty tree, so make's up-to-date judgement never runs and the prerequisite list is not consulted. The only file differing between `c0bd078f` and `df172fcb` is that Makefile's prereq expression.

On the 1cc that ISO installed (`cc1`, 10.1.0.1):

| AC / risk | Command | Result |
| -- | -- | -- |
| **Migrate configs** — the setting that gated the build | `grep ALLOWED_HOSTS /etc/openstack-dashboard/local_settings` | `ALLOWED_HOSTS = ['*']`. This is the file `config_horizon.cpp` regenerates from `.in` on every `Commit()`, so it is the value that actually reaches django |
| **Migrate configs** — policy files | `dump_default_policies` over all seven namespaces | keystone / nova / cinder / glance / neutron / octavia / masakari, each a non-empty yaml |
| **The service runs at all** | `systemctl is-active openstack-dashboard` | `active`, gunicorn master + 4 workers under `/usr/local/bin/python3.11`. **This is itself the runtime proof of the ALLOWED_HOSTS fix** — django 4.2 refuses to start on the old value |
| Version actually shipped | `pip show horizon` in the caracal venv | `24.0.2`, in `python3.11/site-packages` |
| **Log in** | POST `/horizon/auth/login/` with `auth_type=credentials` | HTTP 302, session established, `/project/` returns 200 |
| **Volume / network / VM panels** | GET each panel authenticated | `/project/volumes/`, `/project/networks/`, `/project/routers/`, `/project/instances/`, `/admin/volumes/`, `/admin/networks/`, `/admin/instances/`, `/admin/hypervisors/` — **all 200** |
| **Other panels** | 23 authenticated paths swept | **all 200, zero 500** — stacks, shares, load_balancer, dnszones, reverse_dns, vpn, ironic, goals, audits, action_plans, identity, settings |
| **masakari's carried patch** | `grep default_panel .../masakaridashboard/dashboard.py` | `default_panel = 'segments'`, and `panels` contains `'vmoves'`. `/masakaridashboard/` 200, `hosts` / `notifications` / `vmoves` 200. The sidebar renders Project / Admin / Identity / Instance-ha — **`NoReverseMatch` does not occur**, which is the failure this patch exists to prevent and it takes down every page, not one panel |
| **Static assets** (new: httpd serves them directly) | `curl` a collected css + httpd access log | HTTP 200; **404 count 0**. `ProxyPass /horizon/static !` sits at line 31 against `ProxyPass /horizon` at line 41, so the ordering ProxyPass requires holds |
| **Offline compression** | asset URLs on the login page | `output.b597bc3299cf.css` — hashed names, so `COMPRESS_OFFLINE`'s manifest is in use |
| **SCRIPT_NAME** (new: gunicorn, not `WSGIScriptAlias`) | `curl -I /horizon/` | 302 → `/horizon/auth/login/?next=/horizon/` — prefix preserved through the proxy. `IndexError` count in horizon.log: **0**, which is what a stripped prefix would produce on every request |
| osc cli moved | `readlink -f /usr/bin/openstack`; `openstack --version` | `/opt/openstack-caracal/bin/openstack`, **6.6.1** |
| **All nine osc plugins resolve** | `importlib.metadata` over `openstack.cli.extension` | `accelerator`, `baremetal`, `dns`, `ha`, `infra_optim`, `key_manager`, `load_balancer`, `orchestration`, `share` all registered. `openstack segment list` answers `Missing value auth-url` — i.e. resolved, just uncredentialed |
| Antelope venv emptied of services | `pip list` in the antelope venv | zero openstack services, zero dashboard packages |
| **Restart survives** (new: collectstatic/compress moved to `ExecStartPre`) | `systemctl restart openstack-dashboard` | ExecStartPre completes in **8s** against `TimeoutStartSec=5min`; login back to 200 within ~10s; a full login → `/project/` → `/masakaridashboard/` round trip still 302/200/200 |
| Regression: whole-cluster health | 1cc e2e through `test_30` | green (build #58) |

**Still to check by hand, and not covered above:** performing an actual create/delete of a volume, network and instance *through the dashboard* (the sweep proves the panels load, not that operations succeed); logging in through **Cube Account** (Keycloak SAML), which is the form's default `auth_type` where the sweep used `credentials`; and `cube_edge_cfg()`, which only runs on an edge deployment.

Close #636

`ALLOWED_HOSTS = '*'` had been wrong since it was written; django 3.2 simply never checked. It is worth a look at whether anything else in `local_settings.in` is in `tuple_settings` — `INSTALLED_APPS`, `TEMPLATE_DIRS`, `LOCALE_PATHS`, `SECRET_KEY_FALLBACKS` — for the same class of latent type error.

The systemd unit is `openstack-dashboard`, not `horizon`.

### DIFF
--- a/core/barbican/barbican.mk
+++ b/core/barbican/barbican.mk
@@ -32,25 +32,21 @@ rootfs_install::
 	$(Q)# clean up dns configurations after downloading packages
 	$(Q)rm -f $(ROOTDIR)/etc/resolv.conf
 
-# python-barbicanclient stays in the *antelope* venv -- TEMPORARY, see below
+# python-barbicanclient
 #
-# It is the only part of barbican that does not follow the service into the caracal venv,
-# because it is the one part the openstack cli loads rather than the one barbican runs.
-# python-barbicanclient contributes an [openstack.cli.extension] entry point plus sixteen
-# [openstack.key_manager.v1] commands (secret store/get/list/delete/update, the container
-# and consumer families, and the order family), and a stevedore entry point is only visible
-# to the interpreter it was installed under. /usr/bin/openstack is the antelope venv's
-# (core/heavyfs/Makefile links it there), so installing the client into the caracal venv
-# would take `openstack secret ...` off the node entirely.
+# It is kept in a block of its own because it was the one part of barbican that could not
+# follow the service into the caracal venv: it is what the openstack cli loads rather than
+# what barbican runs. python-barbicanclient contributes an [openstack.cli.extension] entry
+# point plus sixteen [openstack.key_manager.v1] commands (secret store/get/list/delete/
+# update, the container and consumer families, and the order family), and a stevedore entry
+# point is only visible to the interpreter it was installed under, so it has to sit next to
+# /usr/bin/openstack or `openstack secret ...` leaves the node entirely.
 #
-# This is the first caracal hop to hit that. keystone, glance, cinder and nova escaped it
+# #632 was the first caracal hop to hit that. keystone, glance, cinder and nova escaped it
 # because `openstack image|volume|server ...` are osc built-ins -- python-keystoneclient and
 # python-cinderclient contribute no [openstack.cli.extension] at all -- so their clients
-# could move with the service. The same split is already spelled out for designate in
-# core/designate/designate.mk, from the era when the cli was the system python's.
-#
-# TEMPORARY: this goes away when the cli itself reaches the caracal venv, at which point
-# this install block moves back into the one above.
+# could move with the service. #636 moved the cli, so the split is over; the block stays
+# separate only because the reasoning above is worth keeping next to the install.
 rootfs_install::
 	$(Q)# enable dns in the rootfs for downloading packages
 	$(Q)cp -f /etc/resolv.conf $(ROOTDIR)/etc/
--- a/core/barbican/barbican.mk
+++ b/core/barbican/barbican.mk
@@ -54,8 +54,8 @@ rootfs_install::
 rootfs_install::
 	$(Q)# enable dns in the rootfs for downloading packages
 	$(Q)cp -f /etc/resolv.conf $(ROOTDIR)/etc/
-	$(Q)chroot $(ROOTDIR) bash -c "source $(OPENSTACK_HOME_DIR)/bin/activate && \
-		pip install -c $(OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
+	$(Q)chroot $(ROOTDIR) bash -c "source $(CARACAL_OPENSTACK_HOME_DIR)/bin/activate && \
+		pip install -c $(CARACAL_OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
 		python-barbicanclient"
 	$(Q)# clean up dns configurations after downloading packages
 	$(Q)rm -f $(ROOTDIR)/etc/resolv.conf
--- a/core/cyborg/cyborg.mk
+++ b/core/cyborg/cyborg.mk
@@ -74,13 +74,13 @@ rootfs_install::
 rootfs_install::
 	$(Q)# enable dns in the rootfs for downloading packages
 	$(Q)cp -f /etc/resolv.conf $(ROOTDIR)/etc/
-	$(Q)chroot $(ROOTDIR) $(OPENSTACK_HOME_DIR)/bin/pip install \
-		-c $(OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
+	$(Q)chroot $(ROOTDIR) $(CARACAL_OPENSTACK_HOME_DIR)/bin/pip install \
+		-c $(CARACAL_OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
 		python-cyborgclient
 	$(Q)# clean up dns configurations after downloading packages
 	$(Q)rm -f $(ROOTDIR)/etc/resolv.conf
-	$(Q)# Link the cli plugin's console script, which is the antelope venv's
-	$(Q)chroot $(ROOTDIR) ln -sf $(OPENSTACK_HOME_DIR)/bin/cyborg /usr/bin/cyborg
+	$(Q)# Link the cli plugin's console script
+	$(Q)chroot $(ROOTDIR) ln -sf $(CARACAL_OPENSTACK_HOME_DIR)/bin/cyborg /usr/bin/cyborg
 
 rootfs_install::
 	$(Q)[ -d $(CYBORG_PATCHDIR) ] && cp -rf $(CYBORG_PATCHDIR)/* $(CYBORG_SRCDIR)/ || /bin/true
--- a/core/cyborg/cyborg.mk
+++ b/core/cyborg/cyborg.mk
@@ -53,24 +53,22 @@ rootfs_install::
 	$(Q)chroot $(ROOTDIR) ln -sf $(CARACAL_OPENSTACK_HOME_DIR)/bin/cyborg-status /usr/bin/cyborg-status
 	$(Q)chroot $(ROOTDIR) ln -sf $(CARACAL_OPENSTACK_HOME_DIR)/bin/cyborg-wsgi-api /usr/bin/cyborg-wsgi-api
 
-# the osc plugin stays in the antelope venv -- TEMPORARY
+# the osc plugin
 #
 # python-cyborgclient owns the `openstack accelerator ...` commands, and a
 # stevedore entry point is only visible to the interpreter it was installed
-# under. /usr/bin/openstack is the antelope venv's (core/heavyfs/Makefile), so
-# the plugin has to stay next to it or those commands disappear from the cli.
-# Every service whose client owns an osc plugin meets this on its way over.
+# under, so the plugin has to sit next to /usr/bin/openstack or those commands
+# disappear from the cli. That held it in the antelope venv while the cli was
+# there; #636 moved the cli, so it sits with the service again. Every service
+# whose client owns an osc plugin met this on its way over.
 #
 # Nothing in hex_sdk drives this one -- health_cyborg_check() only asks systemd
 # whether the three units are running -- so no health check depends on it. What
 # would go quiet is the operator-facing cli.
 #
 # It is a pip install rather than a git checkout for the same reason the service
-# is, and no version is named: os-antelope-pip-upper-constraints.txt already
-# carries python-cyborgclient (2.1.0), so a version here could only drift from
-# that file.
-#
-# It goes once /usr/bin/openstack is the caracal venv's (#636).
+# is, and no version is named: os-caracal-pip-upper-constraints.txt already
+# carries python-cyborgclient, so a version here could only drift from that file.
 rootfs_install::
 	$(Q)# enable dns in the rootfs for downloading packages
 	$(Q)cp -f /etc/resolv.conf $(ROOTDIR)/etc/
--- a/core/designate/designate.mk
+++ b/core/designate/designate.mk
@@ -5,14 +5,15 @@ ROOTFS_DNF += bind bind-utils
 
 # python-designateclient owns the "dns" osc plugin entry point, which hex_sdk's
 # health_designate_check() drives as `openstack dns service list`, counting the
-# api/central/worker/producer/mdns rows that report UP. Since #634 it is installed by
-# its own block, the one that keeps it in the antelope venv while the service moves to
-# caracal -- see the note there for why it cannot follow.
+# api/central/worker/producer/mdns rows that report UP. It is installed by its own
+# block further down, which #634 split out to hold it in the antelope venv while the
+# service moved to caracal; #636 took /usr/bin/openstack to caracal too, so the block
+# now installs alongside the service and only the separation remains.
 #
 # It used to be the yoga python3-designateclient rpm under the *system* python 3.9,
 # because /usr/bin/openstack was itself `#!/usr/bin/python3` and a stevedore entry point
 # is only visible to the interpreter it was installed under. core/heavyfs moved the cli
-# into the venv, so the plugin follows it and the rpm is gone.
+# into a venv, so the plugin follows it and the rpm is gone.
 #
 # It is still named explicitly rather than left to designate-dashboard's requirements,
 # for the same reason it had to be named as an rpm: it once arrived only as a side effect
@@ -72,22 +73,20 @@ rootfs_install::
 	$(Q)# clean up dns configurations after downloading packages
 	$(Q)rm -f $(ROOTDIR)/etc/resolv.conf
 
-# the osc plugin and the dashboard stay in the antelope venv -- TEMPORARY
+# the osc plugin and the dashboard
 #
 # python-designateclient owns the "dns" osc plugin entry point, and a stevedore entry
-# point is only visible to the interpreter it was installed under. /usr/bin/openstack
-# is the antelope venv's, so the plugin has to stay next to it or
-# health_designate_check()'s `openstack dns service list` cannot run at all -- which is
-# how this was first found, as "DNSaaS NG [ designate(9 api not all up) ]" while every
-# designate unit was active. See the note at the top of this file; #632 has the same
-# constraint for python-barbicanclient.
+# point is only visible to the interpreter it was installed under, so the plugin has
+# to sit next to /usr/bin/openstack or health_designate_check()'s `openstack dns
+# service list` cannot run at all -- which is how this was first found, as
+# "DNSaaS NG [ designate(9 api not all up) ]" while every designate unit was active.
+# That held it in the antelope venv from #634 until #636 moved the cli here; #632 had
+# the same constraint for python-barbicanclient.
 #
 # designate-dashboard is a horizon plugin: core/horizon/horizon.mk copies its enabled
 # panels out of $(HORIZON_VENV_SP), which is the site-packages of whichever venv
 # horizon runs in, so the dashboard goes where horizon goes. #636 took horizon to
 # caracal, so the panel is a caracal-venv install now.
-#
-# The osc plugin goes once /usr/bin/openstack is the caracal venv's (#636).
 rootfs_install::
 	$(Q)# enable dns in the rootfs for downloading packages
 	$(Q)cp -f /etc/resolv.conf $(ROOTDIR)/etc/
--- a/core/designate/designate.mk
+++ b/core/designate/designate.mk
@@ -91,8 +91,8 @@ rootfs_install::
 rootfs_install::
 	$(Q)# enable dns in the rootfs for downloading packages
 	$(Q)cp -f /etc/resolv.conf $(ROOTDIR)/etc/
-	$(Q)chroot $(ROOTDIR) $(OPENSTACK_HOME_DIR)/bin/pip install \
-		-c $(OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
+	$(Q)chroot $(ROOTDIR) $(CARACAL_OPENSTACK_HOME_DIR)/bin/pip install \
+		-c $(CARACAL_OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
 		python-designateclient
 	$(Q)# --no-build-isolation because this pulls horizon; see core/heavyfs/Makefile.
 	$(Q)chroot $(ROOTDIR) $(CARACAL_OPENSTACK_HOME_DIR)/bin/pip install \
--- a/core/designate/designate.mk
+++ b/core/designate/designate.mk
@@ -24,9 +24,14 @@ ROOTFS_DNF += bind bind-utils
 NAMED_CONF_FILES := /etc/named*
 NAMED_APP_DIR := /var/named
 
-# https://releases.openstack.org/teams/designate.html
-# Only the dashboard is still built from git; designate itself is a pinned pip install.
-DESIGNATE_DASHBOARD_REPO_URL := https://github.com/openstack/designate-dashboard.git
+# https://releases.openstack.org/caracal/index.html#caracal-designate-dashboard
+# Horizon plugins are not in the upper-constraints (that file only covers libraries),
+# so the pin is explicit. This was a $(OPS_GITHUB_BRANCH_02) clone until #636: the
+# branch name was chosen while the dashboard had to match an antelope horizon, and a
+# branch resolves to whatever its tip is on build day. Now that the panel follows
+# horizon into the caracal venv the version has to change anyway, so it changes to a
+# number.
+DESIGNATE_DASHBOARD_VER := 18.0.0
 
 DESIGNATE_CONF_DIR := /etc/designate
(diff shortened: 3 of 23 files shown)


core/monasca/monasca.mk
```diff
@@ -83,6 +83,31 @@ rootfs_install::
...
+# python-monascaclient (2.8.0), so a version here could only drift from that file.
+rootfs_install::
+	$(Q)# enable dns in the rootfs for downloading packages
+	$(Q)cp -f /etc/resolv.conf $(ROOTDIR)/etc/
+	$(Q)chroot $(ROOTDIR) $(CARACAL_OPENSTACK_HOME_DIR)/bin/pip install \
+		-c $(CARACAL_OPENSTACK_INSTALLED_PIP_CONSTRAINT) \
+		python-monascaclient
+	$(Q)# clean up dns configurations after downloading packages
+	$(Q)rm -f $(ROOTDIR)/etc/resolv.conf
+
```
Eandalf-Bigstack: We would not need this CLI anymore. We are going to remove Monasca in #672 since OpenStack community decided to deprecate Monasca in Bobcat, which is before Caracal.

SekiXu: Dropped in 31b3449a, PR description updated. Thanks!

Eandalf-Bigstack (approved):

Issue #1405, opened 2026-09-01
[Bug]: health_dns_check's timeout is defeated by SSH connection reuse, stalling the boot-time cluster check for 31 minutes


traviswu-bigstack:
## Description

`health_dns_check` wraps its per-node probe in `timeout $SRVSTO` (10s), so with 5 nodes the loop should cap at ~50s. During the `cluster powercycle` recovery on a 3c2p cluster it ran for **31 minutes**, blocking the boot-time `cluster check` and, through it, the master's `finalizing` phase.

The timeout is inert. `sdk_health.sh:265` merges the remote stderr into the *local* pipe:

```bash
local real=$(timeout $SRVSTO ssh root@$node time arp -a 2>&1 | grep real | awk '{print $2}') || ERR_CODE=1
```

CubeCOS configures SSH connection reuse (`config_cubesys.cpp:119-124`: `ControlMaster auto`, `ControlPath /run/cube-ssh-%C`, `ControlPersist 120s`). When a client reuses an existing master it hands its stdio to that master over the control socket, so killing the client closes nothing — `grep`/`awk` never see EOF and the pipeline lasts as long as the *remote* command does. Caught in the act on the affected cluster, the hung pipe had exactly two holders:

```
pid=2328096 fd=0  grep real                             <- reader
pid=245042  fd=8  ssh: /run/cube-ssh-f998cf...[mux]     <- persistent master holds the write end
```

Because the probe times `arp -a`, and DNS was still broken during boot, the remote command itself took ~31 minutes — exactly what the 10s cap was supposed to prevent.

## How to Reproduce

Both conditions are required — `2>&1` **and** a pre-existing mux master:

| form | master exists | elapsed (timeout 5s) |
|---|---|---|
| `timeout 5 ssh host 'sleep 40' \| grep -c .` | no | 5s |
| `timeout 5 ssh host 'sleep 40' \| grep -c .` | yes | 5s |
| `timeout 5 ssh host 'sleep 40' 2>&1 \| grep -c .` | no | 5s |
| `timeout 5 ssh host 'sleep 40' 2>&1 \| grep -c .` | **yes** | **40s** |

```bash
ssh -O exit root@<peer> ; ssh root@<peer> true          # establish a master
time (timeout 5 ssh root@<peer> 'sleep 40' 2>&1 | grep -c .)   # takes 40s, not 5s
```

## Expected behavior

[...]

Issue #1390, opened 2026-08-31
[Bug]: rolling_update does not proactively drain a node from haproxy backends before reboot, causing sporadic control-plane 503s


chase-chou:
## Describe the bug

When `rolling_update` reboots each node in sequence, it relies purely on haproxy's passive health checks (`check inter 2000 rise 2 fall 5`) to detect that a node's services have gone down, instead of proactively removing that node from haproxy's backend lists before rebooting it. Since the reboot is triggered by `rolling_update` itself and the timing is fully known ahead of time, this passive-detection window (up to roughly 10 seconds, since a backend needs 5 consecutive failed checks before being marked DOWN) is entirely avoidable. During that window, any control-plane API request that haproxy's `balance source` (source-IP hashing to a fixed backend) happens to route to the node that is currently rebooting will get a connection failure or a 503, and depending on the nature of the request this can produce consequences that do not self-heal. In this run, a VM delete request landed in that exact window and the VM ended up stuck in `ERROR` because the resource-usage sync step of its deletion failed, requiring manual intervention.

## Affected Component / Service

`sdk_power.sh` (the `power_roll_advance`/`power_roll_start` rolling-update/restart relay); the affected haproxy `listen` blocks include but are not limited to `nova_placement` (port 8778).

## Version

- Product/Fixpack version: CubeCOS 3.1.0 + Fixpack001/002/003, rolling_update to 3.1.10 RC2 (`CUBE_3.1.10_20260824-1451_27df138.pkg`)
- Component version: the `sdk_power.sh` shipped via the Fixpack003 backport, containing the same rolling logic as 3.1.10 RC2

## To Reproduce

1. Prepare a cluster with at least one compute node and trigger `hex_sdk power_roll_start upgrade <pkg>`.
2. While rolling_update is rebooting one of the control nodes, continuously send requests from a compute node to a control-plane API (for example placement on port 8778).
3. Watch the rebooting control node enter `bootstrapping` and restart its local OpenStack services. If the compute node's requests happen to be hashed by haproxy to that specific node, they will get connection failures or 503s for the tens of seconds it takes haproxy's health check to accumulate enough failures and reroute to a surviving node.

## Expected behavior

Before `rolling_update` reboots a node, it should proactively remove that node from every relevant haproxy backend list (disable it), then add it back (enable it) once the node is confirmed healthy again after rebooting — it should not rely solely on passive health-check detection for an event it already knows is about to happen.

## Root Cause (if known)

[...]

PR #1412, opened 2026-09-02
fix(health): make wait_for_http_endpoint say why it is still waiting

traviswu-bigstack:
#### What type of PR is this?

/kind bug

Follow-up to #1406 (merged). That PR bounded `wait_for_http_endpoint` by the clock; this one makes it say *why* it is waiting.

Today the wait throws away everything it learns — `$CURL -sf http://$endpoint:$port > /dev/null` keeps neither the exit code nor the HTTP status. During the `cluster powercycle` post-mortem, keystone's wait ran **361 s** and apache2's **377 s** on each control node, and nothing on the waiting node recorded whether the endpoint had refused the connection, timed out, or answered 503. Answering that required going to *another* node's keystone access log — which showed keystone healthy and serving **288 requests**, including 26 `POST /v3/auth/tokens`, for the whole window. The endpoint was never down; the probes were not reaching it.

That distinction matters, and it is the difference between this being a slow-endpoint problem and a reachability one (#1390). Six minutes of silent retries should not need cross-node log archaeology to classify.

Now: the reason is logged on the first failure, every 30th after that, and once on give-up:

```
WARN: wait_for_http_endpoint 127.0.0.1:18099 not ready after 1 tries (curl rc=7 http=000)
WARN: wait_for_http_endpoint 127.0.0.1:18099 gave up after 4s (curl rc=7 http=000)
```

`curl rc=7` is connection refused — the line that would have pointed straight at reachability rather than the endpoint.

Context for #1390 — it does not fix the `balance source` behaviour there, it makes the next occurrence self-diagnosing. #1405 was closed by #1406.

> Note

- Behaviour change: the probe now uses `-s -o /dev/null -w '%{http_code}'` and treats 2xx/3xx as success, rather than relying on `-f`. Verified both ways: a live endpoint returns `rc=0` immediately, and a refused one returns `rc=1` at exactly its budget with the reason logged.
- Rate-limited deliberately — one line at the start, then every 30th attempt — so a genuinely slow cold boot does not flood the console the way an unbounded writer would (see the console-stall note in the handbook).
- `log_warning` is the sdk's own helper (`sdk_log.sh`), already used across `sdk_ceph.sh`, `sdk_git.sh`, `sdk_ovn.sh`.

#### Additional documentation

```docs

```

### DIFF
--- a/core/main/proj_functions
+++ b/core/main/proj_functions
@@ -911,14 +911,24 @@ wait_for_http_endpoint()
     # $timeout is seconds: bound the loop by the clock, not by attempt count -- each
     # probe can cost curl's own timeout (SRVTO, up to 60s) on top of the sleep
     local deadline=$(( $(date +%s) + timeout ))
-    local rc=1
+    local rc=1 tries=0 crc=0 code=
     while [ $(date +%s) -lt $deadline ] ; do
-        if $CURL --connect-timeout 2 -m 5 -sf http://$endpoint:$port > /dev/null ; then
+        code=$($CURL --connect-timeout 2 -m 5 -s -o /dev/null -w '%{http_code}' http://$endpoint:$port 2>/dev/null)
+        crc=$?
+        if [ $crc -eq 0 ] && [ "${code:-0}" -ge 200 ] && [ "${code:-0}" -lt 400 ] ; then
             rc=0
             break
         fi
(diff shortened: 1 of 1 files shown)


traviswu-bigstack:
Dropping this one. The branch is left in place, so it can be reopened or picked up later if the approach is wanted.

PR #1418, opened 2026-09-03
feat(health): check that lb backend membership matches node readiness

traviswu-bigstack:
#### What type of PR is this?

/kind bug

`health_haproxy_check` and `health_haproxy_ha_check` only ask whether the process is running. Neither looks at **backend membership**, which is where #1390 lives: `balance source` pins a client to one server and `check inter 2000 rise 2 fall 5` keeps a dead one in rotation for ~10s, so that client's control-plane traffic goes into a hole while every existing check still reports healthy.

The cold-boot post-mortem showed the same exposure with no node rebooting at all: during bootstrap, haproxy-ha on the VIP holder was restarted **4 times across 10 process generations in 16 minutes** by module commits, and a freshly started haproxy treats every server as up until its own checks run.

This adds `lb_backend`, in the `VirtualIp` group beside `vip` and `haproxy_ha`:

| code | condition | why it matters |
|---|---|---|
| 1 | server is `UP` in haproxy but the node is not ready | the #1390 window — source-hashed traffic is being sent to a node that cannot serve it |
| 2 | server is `MAINT` but the node is ready | a drain that was never restored; capacity missing silently, and the failure mode a proactive `disable server` can leave behind |

`_health_lb_backend_auto_repair` fixes both over the runtime API — `disable server` / `enable server` on the admin socket the config already defines at `level admin` (`config_haproxy.cpp:71,250`). No config is rewritten and no process is restarted.

Part of #1390 — it does not replace the proactive drain proposed there (see the limitation below).

> Note

- **Naming:** the component is `lb_backend`, not `lbaas`, because `S[LBaaS]` already denotes octavia in this tree (`cli_cluster.cpp:188`). Happy to rename if you would rather it read `lbaas` and octavia's group be renamed instead.
- **Limitation worth weighing:** the health framework gates auto_repair on `cube_cluster_ready && ! is_cluster_rolling`, so this repair deliberately will **not** fire during a rolling update — which is #1390's original scenario. It reports there, and repairs in the other cases (cold boot, ad-hoc node failure, stale drains). The rolling-update window still wants the proactive `disable server` in the relay that #1390 proposes; this is the safety net around it, not a replacement.
- **Validated against a live cluster** (3-node lab, read-only): the parser returns 38 servers per node with `svname` equal to the node hostnames, so the readiness lookup keys correctly; statuses observed were 99 `UP` / 15 `DOWN`. A healthy cluster is silent — `DOWN` on a ready node is not flagged, since a service that is simply not running there is another check's business.
- Field 18 is the server state (`UP`/`DOWN`/`MAINT`); the pre-existing `haproxy_stats` helper reads field 37, which is only the last check's verdict (`L7OK`/`L4CON`). The new helper uses 18 and keeps 37 as detail.
- Readiness is resolved once per node rather than once per proxy — every listen block repeats the same three servers, and `remote_run` is bounded by `is_sshable`'s `timeout 2 ssh` with a 1-minute cache.

#### Additional documentation

```docs

```


traviswu-bigstack:
Dropping this one. The branch is left in place, so it can be reopened or picked up later if the approach is wanted.

Issue #1369, opened 2026-08-26
[Feature] Unified resize action with plan-then-confirm (LIVE/COLD)


traviswu-bigstack:
## Description

`resize` and `live-resize` are two separate operator-facing actions today, and the operator has
to know in advance which one will work. Fold them into **one** action with plan-then-confirm:
Skyline badges each candidate flavor **LIVE** or **COLD** with the reason, and the confirm text
states the consequence — no reboot, versus restart plus a possible host move and a
confirm/revert step.

**Never a silent fallback.** Cold reboots the guest, cold-migrates it, and lands in
`VERIFY_RESIZE` with auto-confirm disabled, holding allocations on two hosts until someone acts.
That is downtime the operator must consent to, not a fallback we pick for them. Spike #1366
measured this; both VMware and Nutanix behave the same way, and where Nutanix's Terraform
provider *does* silently restart a VM, users filed it as a bug.

Design settled by spike **#1366**:

- The plan call returns the **instance's capability envelope once** — real ceiling, boot source,
  current shape, blocked specs — at **zero RPC cost**, because the ceiling is recorded in
  `system_metadata` at spawn. The UI evaluates every candidate flavor locally. Per-flavor calls
  would cost ~46 ms each plus N round-trips to the instance's host per dialog open.
- One authoritative re-check at submit for the chosen flavor only, closing the staleness race
  between opening the dialog and acting.
- Policies stay separable: enforcement is an imperative `context.can()` in the controller, so a
  second check after mode resolution needs no restructuring.
- **LIVE means "no reboot will be triggered", never "the guest will get the resources"** —
  guest-side onlining is invisible to the host (#1361).

## Definition of Done (DoD)

- [x] Ceiling recorded at spawn and exposed through a plan endpoint, no compute RPC
- [x] Plan response carries boot source, so the root-disk rule is applied correctly for
      boot-from-volume (see skyline-console#127)
- [x] Single action resolves mode server-side and enforces `:live_resize` or `:resize` accordingly
- [x] A live-impossible request with no cold consent returns 400 with the reason — never a silent
      downgrade
- [x] Skyline badges LIVE/COLD per flavor with the reason, replacing the silent grey-out
- [x] Confirm text states the real consequence and does not overclaim what LIVE guarantees
- [x] `VERIFY_RESIZE` surfaced in the UI with confirm/revert
- [ ] Unit tests for mode resolution and both policy paths

## Technical Notes

- **Feature Owner:** travis.wu
[...]

PR #1429, opened 2026-09-05
feat(nova): let an image-backed instance grow its root disk live

traviswu-bigstack:
## What type of PR is this?

/kind feature

## What this PR does / why we need it

`root_gb` was blocked outright for image-backed instances, so every flavor in a ladder that grows disk — which is every stock ladder — resolved to COLD for them. Only boot-from-volume roots could grow live, and only because cinder was doing the growing.

The libvirt image backends already know how to resize in place: `resize_image()` is implemented by `Flat`, `Qcow2`, `Lvm`, `Rbd` and `Ploop`, and nova already pairs it with a guest block-resize in `extend_volume()`. This reuses both, so it works on whatever `images_type` a deployment runs rather than only the ceph pool CubeCOS happens to set.

- `root_gb` is now **grow-only** rather than fixed, and only for image-backed — a volume-backed root's `root_gb` is ignored by nova entirely, so it stays unconstrained.
- `check_live_resize` refuses up front when the backend never overrode `resize_image`, joining the vCPU and memory ceilings as a clean pre-hotplug failure.
- The grow runs **after** the hotplugs. Growing is irreversible while a DIMM attach or `setVcpusFlags` rolls back, and `_lr_revert_allocations` restores the old flavor wholesale — growing first would leave a disk bigger than `DISK_GB` claims after a refused hotplug, which is the common failure.
- `set_allocations` takes an optional `root_gb`, so `DISK_GB` moves only for an image-backed root the compute is about to grow. A boot-from-volume root is cinder's; claiming host disk for it would double-count. The revert passes it too, or a failed resize leaks the larger claim.

## Behaviour matrix

`images_type = rbd` is set unconditionally under a role check (`config_nova.cpp`), so an **image-backed root is always ceph** — the cinder default volume type cannot reach it. The storage axis only affects boot-from-volume.

| | **BFV** (root = cinder volume) | **Image-backed** (root = ceph rbd) |
| :-- | :-- | :-- |
| **CoS / Skyline** | | |
| any cinder backend | LIVE · CPU+mem · *disk unchanged* | LIVE · CPU+mem+**disk** |
| **CMP** | | |
| ceph (CubeStorage) | LIVE · CPU+mem+**disk** | not reachable from the instances UI |
| non-ceph, online-extend | LIVE · CPU+mem+**disk** | not reachable |
| non-ceph, no online-extend | LIVE · CPU+mem · *disk unchanged* (degraded) | not reachable |

- nova never grows a BFV root — it ignores the flavor's `root_gb` for volume-backed instances. cubecmp's worker is what extends the volume, which is why the same instance resized from Skyline keeps its disk.
- CMP's k8s nodes **are** image-backed (Rancher's OpenStack driver, no `--boot-from-volume`), but they never enter CMP's `instances` collection, so its resize path cannot reach them. Skyline can, and after this change a live resize there grows a node's ephemeral root without replacing the node.
- Every cell that grows disk grows the **block device only**. `growpart`/`resizefs` are cloud-init boot-time modules, so the guest keeps its partition table and filesystem until it reboots or runs `growpart /dev/vda N && xfs_growfs /`.

## Which issue(s) this PR fixes

Refs #1369. Builds on #1371.

## Special notes for your reviewer

Verified on the sky 3cc lab, image-backed guest `lr.small` → `t2.xlarge` with a shell open on it:

```
BEFORE  boot=2026-09-05 04:29:35  cpus=2  mem=3G   vda=20G
AFTER   boot=2026-09-05 04:29:35  cpus=4  mem=15G  vda=160G
        vda: detected capacity change from 41943040 to 335544320
```

Same boot timestamp, session never dropped. The same resize returned `400 root disk size must not change` before the change.

`core/nova/tests/test_cube_live_resize.py`: 25/25 against the node's caracal venv, including four covering `DISK_GB` — untouched unless asked for, claimed when asked, never invented when the consumer has no `DISK_GB` allocation, and left alone on the CPU-only claim that precedes an auto-migrate.

Two defects were caught by writing the matrix rather than re-reading the code, and are fixed in the second commit: an unconditional shrink guard that would have blocked a legitimate BFV resize, and a disk lookup that used `Image.path` (an rbd connection string `blockResize` rejects) so the guest was never told its disk had grown.

### DIFF
--- a/core/nova/caracal_patch/compute/api.py.patch
+++ b/core/nova/caracal_patch/compute/api.py.patch
@@ -1,6 +1,6 @@
 --- compute/api.py
 +++ compute/api.py
-@@ -49,6 +49,7 @@ from nova.compute import utils as comput
+@@ -49,6 +49,7 @@
  from nova.compute.utils import wrap_instance_event
  from nova.compute import vm_states
  from nova import conductor
@@ -8,7 +8,7 @@
  import nova.conf
  from nova import context as nova_context
  from nova import crypto
-@@ -761,7 +762,8 @@ class API:
+@@ -761,7 +762,8 @@
  
          # Target disk is a volume. Don't check flavor disk size because it
          # doesn't make sense, and check min_disk against the volume size.
@@ -18,7 +18,7 @@
              # There are 2 possibilities here:
              #
              # 1. The target volume already exists but bdm.volume_size is not
-@@ -4160,9 +4162,55 @@ class API:
+@@ -4160,9 +4162,59 @@
  
      # TODO(stephenfin): This logic would be so much easier to grok if we
      # finally split resize and cold migration into separate code paths
@@ -35,8 +35,12 @@
 +                    "name": f.name, "spec": b})
 +        if new.vcpus < old.vcpus or new.memory_mb < old.memory_mb:
 +            _bad(_("live resize is grow-only"))
-+        if new.root_gb != old.root_gb and not is_bfv:
-+            _bad(_("root disk size must not change"))
++        # root_gb is grow-only. An image-backed root is grown in place by the
++        # driver, which refuses in check_live_resize when its image backend
++        # cannot -- the same clean pre-hotplug failure the vCPU/memory ceilings
++        # already use. A boot-from-volume root is cinder's.
++        if new.root_gb < old.root_gb:
++            _bad(_("root disk size cannot shrink"))
 +        if new.ephemeral_gb != old.ephemeral_gb:
 +            _bad(_("ephemeral disk size must not change"))
 +        # ceiling the running domain was booted with (config default or
--- a/core/nova/caracal_patch/compute/api.py.patch
+++ b/core/nova/caracal_patch/compute/api.py.patch
@@ -39,7 +39,7 @@
 +        # driver, which refuses in check_live_resize when its image backend
 +        # cannot -- the same clean pre-hotplug failure the vCPU/memory ceilings
 +        # already use. A boot-from-volume root is cinder's.
-+        if new.root_gb < old.root_gb:
++        if new.root_gb < old.root_gb and not is_bfv:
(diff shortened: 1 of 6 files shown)


core/nova/caracal_patch/compute/api.py.patch
```diff
@@ -35,8 +35,10 @@
 +                    "name": f.name, "spec": b})
 +        if new.vcpus < old.vcpus or new.memory_mb < old.memory_mb:
 +            _bad(_("live resize is grow-only"))
-+        if new.root_gb != old.root_gb and not is_bfv:
-+            _bad(_("root disk size must not change"))
++        # grow-only, and only for image-backed: nova ignores root_gb for a
++        # volume-backed root. The driver refuses in check_live_resize.
++        if new.root_gb < old.root_gb and not is_bfv:
++            _bad(_("root disk size cannot shrink"))
```
Eandalf-Bigstack, line 41: The condition did not match what the comment is saying. We are guarding from a root_gb shrinking case, which is bfi (not is_bfv) at the same time.

Eandalf-Bigstack (approved):

core/nova/caracal_patch/compute/api.py.patch
traviswu-bigstack: Correct — the comment stated three rules and the line implements one. The `check_live_resize` sentence in particular belongs to a different case: the driver refuses a *grow* when the image backend never overrode `resize_image`, which has nothing to do with a shrink.

Reworded to match the condition in ec79fb95:

```python
# an image-backed root can grow but not shrink; nova ignores
# root_gb for a volume-backed root
if new.root_gb < old.root_gb and not is_bfv:
    _bad(_("root disk size cannot shrink"))
```

The condition is unchanged — a shrink on an image-backed root is the only thing rejected here, and a boot-from-volume root's `root_gb` stays unconstrained because nova ignores it.

Issue #648, opened 2026-01-27
[Story] [Optional] Upgrade Kapacitor from 1.5.7 to 1.8.6


jdjgya:
## Description

As a **CubeCOS operator**, I want **Kapacitor kept on a current upstream release** so that **the alerting and metric-relay path is not left on a build several minor versions behind the one our own package channel serves**.

We ship `kapacitor-1.5.7-1.x86_64.rpm`, pinned by an explicit download URL in `core/kapacitor/kapacitor.mk`. `repos.influxdata.com/rhel/9/x86_64/stable` currently serves **1.8.6**.

Kapacitor is not only the alerting engine, so a version bump has to be verified against all three of its roles:

1. **Alerting** — 11 static tasks and 2 templates in `core/kapacitor/tasks` and `core/kapacitor/templates`, plus the per-project VM CPU/memory tasks instantiated from those templates, the `alert_events` re-entry loop, and the smtp / slack / exec handlers.
2. **Write proxy** — every metric and event producer POSTs InfluxDB line protocol to Kapacitor on `:9092`, not to InfluxDB on `:8086`. If Kapacitor stops, ingestion stops.
3. **Replication** — on an HA cluster the four generated `relay_<db>_<rp>` tasks (`telegraf` and `events` × `def` and `hc`) each carry one `influxDBOut()` per control node, because InfluxDB 1.x OSS has no clustering.

This story is scoped to the version bump only. Retiring Kapacitor in favour of Prometheus Alertmanager is #767; moving InfluxDB's workload to Prometheus is #674. The InfluxDB upgrade this issue originally described is not being done — see the decision comment on this issue.

## Acceptance Criteria

> A detailed list of conditions that must be met for the story to be accepted. These statements must focus on functional behavior and constraints rather than specific code-level implementation.

- [x] Upgrade Kapacitor to the latest version — 1.5.7 → **1.8.6** in [#1468](https://github.com/bigstack-oss/cubecos/pull/1468), taken from the signed InfluxData repo rather than the unsigned `dl.influxdata.com` rpm, and pinned in both `ROOTFS_DNF` and `LOCKED_DNF`
- [x] Ensure alert mechanisms are still working — verified end to end on `accept-3cc`: `crits_triggered=1`, `alert_events` processed the re-entrant point, and `CPU00006C` landed in the `events` database with its message fully interpolated

## Outcome

Done in [#1468](https://github.com/bigstack-oss/cubecos/pull/1468), two commits.

[...]

PR #1441, opened 2026-09-07
feat(metrics): scrape the phase 1 prometheus exporters, with the fqdn label Watcher keys on

Eandalf-Bigstack:
#### What type of PR is this?

- feature

Phase 1 of replacing Monasca: get the metrics Monasca supplied into Prometheus, from exporters the community maintains, so the downstream consumers keep working. No collector of our own is written here.

- **Source the exporters, and establish how few are needed.** Four services already speak Prometheus and were checked on a live node rather than assumed — haproxy 2.8 is built `+PROMEX`, rabbitmq 3.11 ships `rabbitmq_prometheus`, influxdb 1.12 already serves `/metrics`, zookeeper 3.8 has a `PrometheusMetricsProvider`. So there is no `haproxy_exporter` and none of the other three. What remains is `node_exporter` from EPEL plus four tarball exporters fetched and digest-verified in `core/exporters/Makefile`, installed into the rootfs by `exporters.mk`
- **Configure, enable and scrape them.** Each exporter's `ARGS` is written per node, the units are enabled per role, and the scrape jobs are generated. `node_exporter` and `ipmi_exporter` run on every role because that is the point of them; memcached and apache only where those services exist. Every listener binds one address rather than `0.0.0.0`
- **Carry the label the consumers key on.** Targets come from a cron generator writing `file_sd` files, because compute and storage are not in `cubesys.control.addrs` and a joining node must not need a `hex_config` commit before it is scraped. Each target carries `fqdn` as well as `instance`, matching the node's hostname exactly as `cubectl node list` reports it — Monasca's per-host series were dimensioned `hostname` and Watcher's `metric_map` keys `host_cpu_usage` / `host_ram_usage` on it

Provenance was checked against the GitHub API rather than the prometheus.io catalogue, which still lists two of these at their old homes. `apache_exporter` is the one entry that is not org-backed (`Lusitaniae/`); it is what the official catalogue lists for Apache, has no org-backed alternative, and feeds Grafana middleware panels only, so it is the one that could be dropped without breaking a consumer. `danielqsj/kafka_exporter` was rejected on the same test — single maintainer, five months stale — in favour of `prometheus/jmx_exporter`, which covers Kafka from the official org and is a separate change because it attaches to the JVM rather than being a scrape target.

- Refs #672 — phase 1, the exporter half of replacing Monasca

> **This PR stacks on `jim.lin/feat/replace-monasca-2` (#1439), and on nothing else** — the branch has been rebased, so #1433 and #1431 are no longer beneath it. The base here is `develop`, so until #1439 merges this PR's commit list also shows its nine commits. Only the seven below belong to this PR:
>
> - `build(exporters): package the prometheus exporters phase 1 needs`
> - `build(exporters): ensure /etc/default exists for the exporter units`
> - `feat(prometheus): generate per-node exporter scrape targets`
> - `feat(haproxy): expose haproxy's own prometheus exporter`
> - `feat(prometheus): configure and scrape the phase 1 exporters`
> - `feat(rabbitmq): enable the shipped prometheus plugin and scrape it`
> - `fix(prometheus): write the exporter targets cron with an explicit mode`
>
> Review it after #1439, or diff against `jim.lin/feat/replace-monasca-2`.
>
> The last of those closes the one CodeQL alert on the branch (`cpp/world-writable-file-creation` on `WriteExporterTargetsCronJob`'s `fopen`), which was also the only review comment. It matters beyond the static analysis: cron refuses to run an entry whose file is group- or world-writable, so a umask that ever let this out at 0664 would freeze the exporter target lists at whatever the commit seeded, with membership changes never picked up and nothing logged.

**Three things were wrong until a live cluster said so, and all three would have shipped silently broken.** `apache_exporter` was pointed at port 80, which is the front-end proxy — it answers `/server-status` with a 302 to https and the exporter reports that as `apache_up 0` rather than as an error; httpd's own vhost is 8080. The haproxy job assumed one stats listener when there are two, the per-node local one on 9100 and the pacemaker singleton on the VIP on 9000, so it is two jobs. And the blackbox module pinned `valid_http_versions`, which made Octavia — whose API answers HTTP/1.0 — report `probe_http_status_code 200` alongside `probe_success 0`, a false failure that would have left `health_octavia_check` permanently NG once it moves onto this series in phase 3.

**Why zookeeper is not here.** 3.8 does carry a `PrometheusMetricsProvider`, but the class lives in `zookeeper-prometheus-metrics.jar` and Kafka's bundled distribution ships only `zookeeper.jar` and `zookeeper-jute.jar` — setting `metricsProvider.className` took zookeeper down with `status=2/INVALIDARGUMENT` on `jim-1cc`. It needs `prometheus/jmx_exporter`, as does kafka. Recorded so it is not tried again.

**Why influxdb is scraped for what it will report, not what it reports today.** 1.12's `/metrics` is 116 lines of go, process and promhttp series with not one influx-specific metric — no shards, no series counts, no writes, no queries, because 1.x keeps those in the `_internal` database. 2.x exposes them properly and #648 moves us there, so wiring the job now means that lands with no further change here.

**Why `node_exporter` does not get its registered port.** CubeCOS haproxy already binds 9100 for the stats listener, and moving that would change an operator-facing URL, so the exporter moves instead.

**Why a missing `EnvironmentFile` had to be ruled out.** The units carry `EnvironmentFile=/etc/default/<name>` and nothing in the packages creates those files — `rpm -qf` reports the four tarball exporters' as owned by no package. A missing one does not mean "start with no arguments": systemd refuses the unit with `Result: resources` and says nothing about the cause. `config_prometheus` writes each file before enabling its unit, so the shipped path is ordered correctly; the build only guarantees the directory.

**Why the blackbox exporter is the one bound to loopback.** For the others the management address is where a Prometheus on another node scrapes them. For blackbox it is the probe that crosses the cluster, not the scrape, so the whole exporter is local. Its job needs `relabel_configs` rather than plain targets: `__address__` has to become the exporter while the original target survives as a label, or every series would be labelled with the exporter instead of the service.

#### Additional documentation

For the requirement "the exporters are sourced reproducibly", in the jail:

```
make -C core/exporters  ->  fetches and digest-verifies blackbox 0.28.0, ipmi 1.10.1,
                            memcached 0.17.0, apache 1.1.1
                        ->  four ELF 64-bit x86-64 binaries in the build tree
```

For the requirement "every target is scraped", from Prometheus's own target list:

```
jim-1cc     16 of 16 targets up
accept-3cc  33 of 33 targets up across twelve jobs
```

For the requirement "the label Watcher keys on is present and correct", on `accept-3cc`:

```
each of the four per-node target files enumerates all three nodes
fqdn=cc1 / cc2 / cc3, against the right management address and port
```

For the requirement "Watcher's host metrics are answerable from node_exporter":

```
host_ram_usage on cc1  ->  27.8 GiB
free -g on cc1         ->  27 used
```

For the requirement "haproxy reports its own counters, both listeners":

```
/metrics  ->  1465 series on the local listener of all three control nodes, and on the VIP's
haproxy -c    accepts the generated config
/haproxy_stats still answers 200 with its existing credentials
```

For the requirement "enabling rabbitmq_prometheus does not disturb the cluster", on `accept-3cc` with a rolling restart:

```
plugin enabled on all three nodes, each binding its own management address
cluster_status still lists all three running nodes
health_rabbitmq_check 0, health_kafka_check 0
rabbitmq_identity_info reporting from cc1, cc2 and cc3
```

`ipmi_up` is 0 on every node on both clusters, as expected on VMs with no BMC — the same behaviour Monasca's `ipmi_sensors` plugin had there. It needs real hardware to validate.

The target generator is idempotent on both clusters: a second run produces byte-identical files.


SekiXu (approved):
LGTM.

Went looking for the two things that usually bite in this area; neither is a problem here:

- Unauthenticated `/metrics` on the stats listeners is consistent with what the tree already does. `ceph_prometheus_ep` proxies the ceph mgr `/metrics` on VIP:9285 with no auth in the same file's srvlist table.
- `_prometheus_targets_write` is a faithful generalization of `lachesis_prometheus_targets`: same `timeout 20 cubectl node list -j`, same "empty output means etcd said nothing, so fail rather than truncate a live list", same `.tmp` + `cmp -s` write. Ports appearing in both the module and the sdk is how 8086 / 9092 / 9095 / 2181 / 11211 already work here.

`Commit()`s `IsUndef(s_eCubeRole)` guard means the two unconditional `SystemdCommitService(true, ...)` calls only reach nodes that have a role, which is what the comment above them claims.

Establishing what does *not* need an exporter, on a live node rather than by assumption, is worth as much as the four that do. The zookeeper note in `exporters.mk` in particular saves the next person a day: "3.8 has a PrometheusMetricsProvider but Kafka's bundled distribution does not ship the jar" is exactly the kind of dead end that gets re-attempted.

Issue #141, opened 2025-06-04
[Bug]: Ceph show “health error”


jimchen77:
## Describe the bug

Upon the shutdown of CC1, CC2 took over. However, Ceph's health status showed "Health error" from 2025/06/04 11:31 to 2025/06/04 11:41.

## Version 

CubeCOS 3.0.0 (RC2)

## To Reproduce

N/A

## Expected behavior

N/A 
## Screenshots

![Image](https://github.com/user-attachments/assets/a2ac39ef-e7ff-4750-9f39-ccb196cf9f36)
![Image](https://github.com/user-attachments/assets/9a1ebcc7-c3d2-4b5c-9b5d-01f3e6738f82) 

## Additional context

- Hardware [e.g. CPU, RAM, Disk Controller, Disks]
- Supporting logs

PR #1442, opened 2026-09-07
fix(metrics): take ceph-mgr off influxdb, and re-land the influxdb exposure change it blocked

Eandalf-Bigstack:
#### What type of PR is this?

- bug
- cleanup

Ceph telemetry had two routes into the monitoring stack and the wrong one was load-bearing. This PR retires it, moves the consumers, and — because that route was the blocker — finally re-lands the InfluxDB exposure change that was reverted a year ago.

- **Take ceph-mgr off InfluxDB, because that route was never HA.** The `influx` mgr module wrote to `shared_id:8086`, straight at InfluxDB, bypassing Kapacitor `:9092` where the peer relay lives — so its points were never replicated. On `accept-3cc` the `ceph` database held 5 measurements on `cc2` and **zero** on `cc1` and `cc3`, while Grafana reads `localhost`. This is a defect fix, not parity work
- **Migrate the clusters that already have it,** because the A/B partition switch does not clear it. The module enablement lives in the mon quorum's mgr map and `mgr/influx/*` in the mon config store — neither is on the rootfs, so a rolling upgrade carries both forward
- **Move the consumers and delete what is now dead.** The Grafana storage dashboard goes to PromQL, the `influxdb-ceph` datasource is removed, and the `stats_storage_*` functions, `join_ceph.tick`, the `ceph` database's relay tasks and its `CreateDBs` entry all go with it
- **Bind the mgr prometheus endpoint per daemon,** which is what makes the replacement route survive failover
- **Re-land [#125](https://github.com/bigstack-oss/cubecos/pull/125).** InfluxDB binds `127.0.0.1` and haproxy publishes it on the management address and the VIP, with writes denied on the VIP

- Refs #672 — the ceph half of phase 3, brought forward because it was a live defect
- Re-lands [#125](https://github.com/bigstack-oss/cubecos/pull/125), reverted by [#143](https://github.com/bigstack-oss/cubecos/pull/143)

> **This PR stacks on `jim.lin/feat/replace-monasca-3` ([#1441](https://github.com/bigstack-oss/cubecos/pull/1441)), which stacks on [#1439](https://github.com/bigstack-oss/cubecos/pull/1439) — and on nothing else.** The branches have been rebased, so #1433 and #1431 are no longer beneath this one. The base here is `develop`, so until #1441 merges this PR's commit list also shows the 16 commits those two contribute. Only the ten below belong to this PR:
>
> - `feat(ceph): bind the mgr prometheus endpoint per daemon, and export perf counters`
> - `feat(grafana): move the ceph storage panels onto prometheus`
> - `refactor(stats): drop the dead ceph storage stats and their kapacitor join`
> - `refactor(influxdb): stop pointing ceph-mgr at influxdb`
> - `fix(ceph): retire the influx mgr module on upgraded clusters`
> - `fix(health): stop the ceph-mgr repair from resurrecting the influx module`
> - `refactor(influxdb): drop the dead ceph database plumbing`
> - `fix(ceph): drop the inherited cluster-wide mgr prometheus address`
> - `feat(influxdb): expose influxdb on mgmt ip and vip, not the wildcard`
> - `fix(haproxy): emit the influxdb sections last, not mid-frontend`
>
> Review it after #1441, or diff against `jim.lin/feat/replace-monasca-3`.
>
> The last commit fixes a placement bug in the one before it, found in review: the new `backend influxdb` and `frontend influxdb_mgmt` were emitted *inside* `frontend cube_cos_http`, so haproxy — whose sections run to the next section header — handed that frontend's four closing lines to `influxdb_mgmt`. `bind :80` was left with no `/api/` or `/saml/` route and **no `default_backend`**, so everything outside `/horizon`, `/grafana`, `/opensearch-dashboards`, `/prometheus` and `/ceph/` would have returned 503. `:443` was unaffected because `cube_cos_https` carries its own copy of those lines. The two sections are now emitted last in `WriteLocalConfig`, after both `cube_cos` frontends are complete.

**The history behind the last commit, since it re-lands a revert.** [#125](https://github.com/bigstack-oss/cubecos/pull/125) *"feat(security): set up authentication for mongodb and only expose mongodb and influxdb on mgmt ip and vip"* merged 2025-06-02 and was reverted the next day inside [#143](https://github.com/bigstack-oss/cubecos/pull/143), whose body records the reason verbatim: ceph's points *"would not go through the existing mechanism to copy InfluxDB inputs to other InfluxDB copies other control node by Kapacitor"*, so ceph graphs went blank. That reason was real, and it is exactly the bypass this PR removes. Two corrections to how it was written down, both checked against the code: the module is a plain HTTP client (`InfluxDBClient(hostname, port…)` in `module.py`), not *"performing file operations on `/var/lib/influxdb/ceph/`*"; and the bind change did not create the split, since `shared_id` was the module's target both before and after #125 — it moved *which* node held the data away from the one serving VIP reads, because that listener balances on `source` rather than roundrobin. Note the revert commit predates issue [#141](https://github.com/bigstack-oss/cubecos/issues/141) by a day, so #141 is not its trigger despite the timing.

**Why the migration is not free with the A/B switch.** `hex_config`'s usual escape — a new partition without the marker files — does not apply, because none of this state is on the rootfs. Worse, the marker that gated the original *enable* (`ceph_mgr_influx_enabled`) was itself `CONFIG_MIGRATE`d, so its absence proves nothing. `migrate_ceph_mgr_influx` carries no marker of its own: as `migrate_ceph`'s own comment already records, a one-shot per-node marker cannot retry, and this needs a serving mgr a freshly booted node often lacks. `config_ceph.cpp`'s `Commit()` owns the gate instead and clears it only on rc 0 — the same shape as `FinalizeCephRelease` beside it, including not being `IsControl`-gated so every node can clear its own.

**Why `mgr/prometheus/server_addr` is scoped per daemon.** It is one value shared by every mgr, and a process can only bind an address that exists on its host — so a single specific IP set cluster-wide is unbindable everywhere else and the endpoint dies the moment the active mgr moves. Setting cluster-wide `10.1.0.1` took the endpoint down on all three `accept-3cc` nodes because the active mgr was `cc3`. The inherited cluster-wide `0.0.0.0` is removed rather than migrated, which is safe mid-roll: with no key at any level ceph's own default takes over, declared `::` and measured to bind `*:9283` and answer HTTP 200 on the IPv4 management address.

**Why `exclude_perf_counters=false` is not new collection.** The per-OSD series the dashboards need — `ceph_osd_stat_bytes`, the op latency sum/count pairs — are excluded by default. The influx module always shipped exactly those counters, so the mgr already gathers them; excluding them here only threw away data already paid for. Measured cost on `accept-3cc`: `/metrics` goes from 1841 lines in ~7 ms to 7183 in ~11 ms.

**One health repair would have undone the migration.** `health_ceph_mgr`'s `ERR_CODE 3` ran `ceph mgr module disable influx` followed by `enable`, so any cluster where ceph reported a failed influx module would have put it straight back. `ERR_CODE 2` was already dead code: it read `$influx_hn`, which is assigned nowhere in the tree, so it ran `ceph config set mgr mgr/influx/hostname` with an empty argument. Both codes are removed and left as **numbering gaps** rather than renumbered, so 4/5/6 keep the meaning they already have in recorded events.

**Why writes are denied on the VIP but allowed on the per-node frontend.** A write through the VIP lands on one backend and is never relayed — that is the trap this PR is closing, so it is closed structurally rather than by convention. Reads stay because every control node holds a full replica and 42 call sites reach InfluxDB as `influx -host $(shared_id)`; six are health checks reading monasca `http_status`, so making them local-only would turn one node's dead InfluxDB into nova, glance, cinder, heat, octavia and designate all reporting down. The per-node frontend must allow writes, because that is what the Kapacitor peer relay and the event inserts use.

**Two things the VIP listener needs that the srvlist loop would not give it.** `mode http` is set explicitly, because the `defaults` section sets no mode and `http-request` is invalid in tcp mode. And it lists every peer with **no `CUBE_MIGRATE` pin**: the loop narrows to the master for the length of a rolling upgrade because a half-upgraded peer is dangerous for wsrep state or a database mid-migration, but reading replicated points is not that — pinning would leave one checked backend, so while that node reboots the route would have no server at all, and InfluxDB rebuilds its TSI on start (why `TimeoutStartSec` is 1800), making that window minutes.

**The `ceph` database is deliberately not dropped.** Its retention policies are already set, so with the writes stopped the data ages out on its own. Deleting an operator's history outright is not this change's call to make.

#### Additional documentation

For the requirement "the influx route was the non-HA one", before any change:

```
accept-3cc  ceph database:  cc1  0 measurements
                            cc2  5 measurements
                            cc3  0 measurements
            mgr prometheus: complete on all three
```

For the requirement "the migration takes an upgraded cluster to a clean state", on `jim-1cc`, which was in exactly the pre-upgrade state (module enabled, all three `mgr/influx/*` keys set):

```bash
[cc1 ~]# hex_sdk migrate_ceph_mgr_influx ; echo rc=$?
rc=0
[cc1 ~]# ceph mgr module ls -f json | jq -r '.enabled_modules[]' | grep -E 'influx|prometheus'
prometheus
[cc1 ~]# ceph config dump -f json | jq -r '.[] | select(.name|startswith("mgr/influx/")) | .name'
                                          # empty
[cc1 ~]# hex_sdk migrate_ceph_mgr_influx ; echo rc=$?      # idempotent
rc=0
```

With an unreachable mgr it returns 1 and leaves the state alone, so the marker survives and the next commit retries.

For the requirement "the writes actually stop":

```
last point in the influx ceph database   14:11:27Z
module disabled at                       14:11:34Z
nothing written in the 242s since        (four 60s intervals)
mgr prometheus still serving             1593 ceph_ series
```

For the requirement "the per-daemon bind wins, and dropping the inherited key is safe", on `jim-1cc` with both keys present:

```bash
[cc1 ~]# ceph config dump -f json | jq -r '... server_addr ...'
mgr      = 0.0.0.0
mgr.cc1  = 10.1.0.1
[cc1 ~]# ss -lntp | grep 9283
LISTEN  10.1.0.1:9283            # the per-daemon key wins
                                 # with no key at any level:
LISTEN  *:9283                   # ceph's own default, and 10.1.0.1:9283 -> HTTP 200
```

`sysctl net.ipv6.bindv6only = 0` and IPv6 not disabled, which is why the `::` default still answers on IPv4.

For the requirement "the Grafana migration is faithful", on the storage dashboard: 21 targets and 2 template variables converted to PromQL, six deprecated `table-old` panels rebuilt as `table`, a latency unit corrected from ns to s, 38 datasource uid references repointed — and the eight telegraf-sourced targets left **byte-identical**. Queried through the pinned datasource after the change:

```
ceph_health_status        1               (matches the cluster's HEALTH_WARN)
ceph_cluster_total_bytes  52846133248
ceph_osd_up               2 series
```

For the requirement "the retired datasource actually goes away on an already-provisioned cluster", from Grafana 12.4.10's log on restart:

```
logger=provisioning.datasources level=info msg="deleted datasource based on configuration" name=influxdb-ceph
```

and the API then lists only the five remaining datasources, health `ok`.

For the requirement "the health surface is correct after the removal":

```
health_ceph_mgr_check      rc=0, reports healthy
health_errcode_dump        ceph_mgr 1 / 4 / 5 / 6   (2 and 3 skipped)
health_errcode_lookup 2    "2 undefined"            (degrades, does not mislabel)
```

For the requirement "the loopback bind breaks no local client", on `jim-1cc` with `bind-address = "127.0.0.1:8086"`:

```
bare influx CLI (config_influxdb.cpp x8)   OK
Grafana's four datasources                 OK  (28 / 120 / 4 / 9 measurements)
monasca-api, kapacitor                     OK  (0 error lines)
```

`getent hosts localhost` returns `::1` first and `http://[::1]:8086/ping` is refused, but `http://localhost:8086/ping` returns 204 — clients fall back to IPv4.

For the requirement "kapacitor keeps writing across the bind change", which is the one that would fail silently:

```
writes in a 100s window          +136
  source ::1                      frozen at 54397
  source 127.0.0.1                179 -> 318
kapacitor error lines             0
telegraf points still landing     yes
```

For the requirement "haproxy publishes it correctly, and does not mask an outage":

```
10.1.0.1:8086/ping      204          /metrics  200, 116 lines, identical to loopback
prometheus influxdb target           health=up
influxdb stopped ->     503          health_influxdb_check rc!=0 (detects it)
influxdb restarted ->   204          backend recovers on its own within one check interval
```

For the requirement "the VIP listener serves reads and refuses writes", running the generated stanza directly:

```
haproxy -c              Configuration file is valid
GET  /ping              204
POST /query             200
POST /api/v2/query      200
GET  /metrics           200
POST /write             403
POST /api/v2/write      403
```

and the denied point never reached InfluxDB.

**Not verified here:** the VIP listener on a real 3-node cluster. `jim-1cc` is single-node and `config_haproxy.cpp` returns before the srvlist section when `ha` is false, so that stanza is never emitted there — it was exercised by running the generated block under a standalone haproxy on a spare port. The live VIP listener, the `balance source` distribution and the peer chain (VIP haproxy → peer management address → peer haproxy → peer loopback) still want a pass on `accept-3cc`.

For the requirement "no generated haproxy section adopts another section's lines" — both candidate layouts built from a real `haproxy.cfg` and judged by haproxy's own parser, since the misplaced form is not a syntax error:

```bash
[cc1 ~]# # as emitted before the fix
acl api_path / use_backend cube_cos_api / default_backend cube_cos_ui  ->  influxdb_mgmt
frontend cube_cos_http  default_backend:  NONE          # -> 503 outside the listed path_begs
haproxy -c: Configuration file is valid
[cc1 ~]# # as emitted now
frontend cube_cos_http   default_backend: yes
frontend cube_cos_https  default_backend: yes
frontend influxdb_mgmt   holds only its own five lines
haproxy -c: Configuration file is valid
```

Both valid, which is the whole reason this needed a reader rather than a check. Section headers in the built object now read `cube_cos_api, cube_cos_ui, cube_cos_http, cube_cos_https, influxdb, influxdb_mgmt`, on a clean `CXX config_haproxy.o` under `-Wall -Werror` and a clean `LD hex_config`.


SekiXu (commented):
The ceph-mgr-off-influx reasoning holds up, and the cleanup half checks out: nothing in cubecos,
cube-cos-api, cube-cos-ui, cube-cos-openapi or triangle calls the deleted `stats_storage_*`, no
product code pins a storage.json panel id, and handling the stale row with `deleteDatasources`
rather than just dropping the entry is the right move for an already-provisioned appliance.
Leaving errcodes 2 and 3 as gaps rather than renumbering is the right call too.

One placement problem in `WriteLocalConfig()`. The new `backend influxdb` and
`frontend influxdb_mgmt` sections are emitted between `use_backend ceph_dashboard_backend` and
the four lines that follow it, and on develop those four belong to `frontend cube_cos_http`:

    acl api_path path_beg /api/
    acl saml_path path_beg /saml/
    use_backend cube_cos_api if api_path or saml_path
    default_backend cube_cos_ui

In the generated haproxy.cfg they therefore end up inside `frontend influxdb_mgmt`.
`frontend cube_cos_http` (`bind :80`) is left with no `/api/` or `/saml/` route and no
`default_backend`, so anything on :80 outside /horizon, /grafana, /opensearch-dashboards,
/prometheus and /ceph/ gets a 503. `cube_cos_https` has its own copy of those lines further down,
so :443 is unaffected - and the config still parses and haproxy still starts, so this would not
show up at startup.

Moving the two new sections below `frontend cube_cos_http`'s `default_backend` and its trailing
blank line fixes it.

Eandalf-Bigstack:
@SekiXu thank you — and this one should not have reached you. **An obvious bug like this has no business slipping into a PR**: two new sections dropped into the middle of an existing `frontend`, taking its closing lines with them. Nothing subtle about it once someone reads which lines the insertion separates, which is exactly what you did. Fixed in 3ab67ebc.

Your description was right in every part. The four lines

```
acl api_path path_beg /api/
acl saml_path path_beg /saml/
use_backend cube_cos_api if api_path or saml_path
default_backend cube_cos_ui
```

belong to `frontend cube_cos_http`, and because haproxy sections run to the next section header they were adopted by `frontend influxdb_mgmt`. `bind :80` was left with no `/api/` or `/saml/` route and no `default_backend` at all, so everything outside `/horizon`, `/grafana`, `/opensearch-dashboards`, `/prometheus` and `/ceph/` would have returned 503, with `:443` unaffected because `cube_cos_https` carries its own copy further down.

On your suggestion I moved them below `cube_cos_http`'s `default_backend` first, then took it one step further on Jim's preference: they now go **last** in `WriteLocalConfig`, after both `cube_cos` frontends are complete. Same fix, and it keeps the http/https pair adjacent instead of split by an unrelated listener. It also makes the diff a plain deletion plus an append rather than a move that reshuffles those four lines.

Verified with haproxy's own parser rather than by eye, because as you noted the broken form is not a syntax error — both layouts built from a real `haproxy.cfg` and judged:

```bash
[cc1 ~]# # as emitted before
acl/use_backend/default_backend  ->  influxdb_mgmt
frontend cube_cos_http  default_backend:  NONE
haproxy -c: Configuration file is valid
[cc1 ~]# # as emitted now
frontend cube_cos_http   default_backend: yes
frontend cube_cos_https  default_backend: yes
frontend influxdb_mgmt   holds only its own five lines
haproxy -c: Configuration file is valid
```

Both "valid" is the part worth keeping in mind: neither `haproxy -c` nor a successful start would ever have reported this, and on :443 everything would have looked normal.

Two notes on scope, so they are on the record rather than assumed:

- **`frontend influxdb_mgmt` itself happened to survive.** Its unconditional `use_backend influxdb` precedes the adopted rules and an unconditional `use_backend` wins, so `:8086` kept working — including `/api/v2/query`, which would otherwise have matched the adopted `path_beg /api/` and been routed to `cube_cos_api`. The damage was confined to `:80`.
- **The comment now carries the reason, not just the placement.** A bare "keep this here" would be re-broken by the next person adding a listener, so it says why: a section emitted above another's closing lines silently adopts them, and the result still parses.

What I could **not** exercise, so you are not left assuming otherwise: I did not regenerate the config with the fixed binary on a live node, nor curl `:80` to observe the 503. `jim-1cc`'s config predates this PR and has no influxdb sections at all, and the node is currently thrashing on memory — 232 MB free of 32 GB, no swap, load 210/434/361, sshd resetting handshakes — so restarting haproxy there was not a reasonable thing to do tonight. The evidence above is haproxy's parser on both layouts plus the compile and link. Happy to do the live regeneration on `accept-3cc`, or on `jim-1cc` once it recovers, if you would rather see it before approving.

Thanks also for confirming the cleanup half — no remaining `stats_storage_*` callers across cubecos, cube-cos-api, cube-cos-ui, cube-cos-openapi and triangle, no product code pinning a storage.json panel id, and `deleteDatasources` being the right handling for an already-provisioned appliance. That was the part with the widest blast radius and the least to show for it if it had been wrong.

SekiXu (approved):
3ab67ebc fixes it, and putting both sections last is better than the move I suggested.

`frontend cube_cos_http` keeps its own closing lines (`config_haproxy.cpp:216-219`) and `cube_cos_https` keeps its copy (`:236-239`); the new `backend influxdb` and `frontend influxdb_mgmt` are the last things written before `fclose` (`:253-266`, `:268`). The VIP half is on a section boundary too - its own `listen influxdb` at the end of `WriteConfig` (`:511-534`).

One thing the write-deny does not cover, pre-existing rather than introduced here. The guard is path-based on `/write` and `/api/v2/write` (`config_haproxy.cpp:515`), and InfluxQL mutations ride POST `/query`: `sdk_stats.sh:34`, `:47` and `:64` all issue `drop series` at `-host $(shared_id)`. With `balance source` that lands on exactly one replica and the series survives on the other two. That is already the behaviour today, because shared_id:8086 currently reaches the VIP holder's influxdb directly, so nothing changes with this PR - only the claim that the deny stops a future writer from rediscovering the trap doesn't extend to that class. Not asking for a change.

PR #1448, opened 2026-09-08
feat(watcher): move the decision engine onto prometheus and supply the instance metrics it needs

### LINKED ISSUE #673: [Story] Move Watcher Data Source from Monasca to Prometheus
## Description

As a CubeCOS operator, I would like to keep Watcher working in Caracal after we change the metric collecting mechanism so that we could still have Watcher provide resource optimization services in Caracal.

## Acceptance Criteria

- [x] Change data source of Watcher from Monasca to Prometheus — **done in [#1448](https://github.com/bigstack-oss/cubecos/pull/1448)**. `[watcher_datasources] datasources = prometheus`, served by the datasource backported from 2025.1 into Caracal's watcher 12.1.0, which predates it. The two upstream files are taken **verbatim**, so the backport is trivially re-portable and drops out entirely on Epoxy, which ships them; the only edits are the registrations Caracal needs. It could not be pointed at Prometheus directly: the datasource builds `host:port/api/v1/...` and `python-observabilityclient` accepts only a host and a port, while our Prometheus runs with a `/prometheus` route prefix — so a bare `/api/v1/query` is a 404 on both `:9091` and the thanos querier's `:10904`. `config_haproxy.cpp` writes a loopback listener that prefixes the path and reuses `prometheus_backend`, so Watcher inherits the UI's routing: the local Prometheus on one node, the deduplicating thanos queriers in HA.
- [x] Remove the patch in v3.1.0 for Watcher with Monasca data source to work — **done in [#1448](https://github.com/bigstack-oss/cubecos/pull/1448)**. `common/clients.py` (which only ever built the monasca client), `decision_engine/datasources/monasca.py` and `metric_map.yaml` are all removed, along with **PR [#548](https://github.com/bigstack-oss/cubecos/pull/548)'s `× 1024.0`** in `workload_balance`. What remains under `core/watcher/` is two verbatim upstream files, two minimal registration edits, `allocation_balance.py`, and a `workload_balance.py` carrying only PR [#515](https://github.com/bigstack-oss/cubecos/pull/515)'s `workload_cache.get()` guard and its diagnostics — that guard is generic robustness and stays, since instances really can be absent from the cache.
- [x] Verify Watcher with Prometheus data source could correctly provide resource optimization services, create and apply patches if needed — **done in [#1448](https://github.com/bigstack-oss/cubecos/pull/1448)**. All three strategies CubeCOS has verified are exercised on both datasources with matching outcomes; see QA Verification below.

**Also landed in [#1448](https://github.com/bigstack-oss/cubecos/pull/1448)** (not in the original criteria — surfaced by this work)

- [x] **The two instance metrics Watcher needs did not exist anywhere, so they are now produced.** `instance_cpu_usage` and `instance_ram_usage` map to `ceilometer_cpu` and `ceilometer_memory_usage`, and nothing in the stack emitted either. They come from a `node_exporter` **textfile** written by `hex_sdk watcher_instance_metrics` — no new service, port or scrape job, and the node's existing target already carries the `fqdn` label. No packaged libvirt exporter passes the provenance bar the phase 1 fleet was chosen under: the upstream one has had no commit since June 2021 and prometheus-community [declined to adopt it](https://github.com/prometheus-community/community/issues/50), a harder fail than the `kafka_exporter` already rejected. No nova call is needed for the label either — nova sets the libvirt domain UUID to the instance UUID, so `virsh list --uuid` is already the value `instance_uuid_label` names.
- [x] **Watcher's metric names are hardcoded and `metric_map.yaml` cannot redirect them.** `METRIC_MAP` fixes the names, and `_build_prometheus_query` then dispatches on the meter name, raising `Cannot process prometheus meter` for anything else. `metric_map_path` *does* override the map (`manager.py:182`), but renaming a Prometheus meter only breaks the dispatch — which is why the collector emits the `ceilometer_*` names and the backported datasource stays byte-identical to upstream.
- [x] **The missing-metric failure is silent, and that is the operability finding.** `get_backend` only rejects a datasource when a metric maps to `None`; `ceilometer_cpu` *is* mapped, merely absent. So before the collector existed the audits still SUCCEEDED, still produced action plans, and decided from no per-instance data with **no warning logged**. This is why the datasource move and the collector are a pair that should not be split across releases.
- [x] **Instance memory is the guest's figure, not the host resident set.** `available - unused`, which is what Ceilometer's `inspect_memory_usage` computes for `memory.usage` and what Monasca's libvirt check computes for `vm.mem.used_gb` — the series `metric_map.yaml` pointed `instance_ram_usage` at, so an existing `threshold` carries over. `balloon.rss` is a different quantity Monasca publishes separately as `vm.mem.resident_gb`; on a 128 MiB guest it measured **133 MiB** against a guest-used **28 MiB**. Where a guest has no virtio-balloon the series is omitted rather than substituted.
- [x] **`allocation_balance` ([#1128](https://github.com/bigstack-oss/cubecos/pull/1128)) is unaffected by construction** — its own header reads *"compute model only, no metrics datasource"*. It behaved identically on both datasources, which makes it the control for isolating datasource-caused regressions.

**Found but deliberately not fixed here**

- [ ] **Monasca's libvirt check cannot import, so it has produced no per-VM metric since v3.1.20.** `checks_d/libvirt.py` imports `libvirt`, `netaddr`, `neutronclient.v2_0` and `novaclient`, none of which `monasca-agent`'s `requirements.txt` lists — they arrived because nova *and* neutron shared `/opt/openstack-antelope`, and the caracal hop moved both out. The agent logs `Initialization failed for checks: dict_keys(['libvirt'])` on every node, **no measurement carries a `resource_id` dimension**, and `workload_balance` therefore resolved a workload for zero instances while still reporting SUCCEEDED. Same defect as [#1348](https://github.com/bigstack-oss/cubecos/issues/1348) one venv over. Naming the four packages restores it and brings 38 `vm.*` measurements back (119 → 210) — but that fix is **not landed**, because Monasca is being removed. It was applied by hand to the lab nodes only, to establish a baseline the Prometheus path could be measured against.

## Notes

The actual goals of this user story depends on the conclusion from #646

**What upstream is doing, for the record.** The Prometheus datasource is **deprecated as of 2026.1** in favour of **Aetos**, a Keystone-authenticated RBAC reverse proxy in front of Prometheus, and the default `watcher_datasources` changed accordingly. Backporting the deprecated datasource is still right for us — Aetos is an additional service plus a Keystone endpoint, and our path is yoga → antelope → caracal → epoxy — but the next Watcher uplift should revisit it. Upstream also assumes **Ceilometer + sg-core** produce the instance metrics, which is why the names are `ceilometer_*`; adopting those two services for two metrics is not worth it, so we emit the names instead.

## Story Points: ?

Although the complexity is moderate, the risk is unknown.

> | -- | -- | -- |

## QA Verification

|---|---|---|
| 1 | `grep -A1 '^\[watcher_datasources\]' /etc/watcher/watcher.conf` | `datasources = prometheus` |
| 2 | `curl -s -o /dev/null -w '%{http_code}' 'http://127.0.0.1:9099/api/v1/query?query=up'` | `200` — the haproxy listener Watcher queries through |
| 3 | Boot one instance, wait two minutes, then `cat /var/lib/node_exporter/textfile/cube_instance_metrics.prom` on its compute node | one `ceilometer_cpu{resource="<instance-uuid>"}` and one `ceilometer_memory_usage{resource="<instance-uuid>"}` line; nodes with no instance show only the `# HELP` / `# TYPE` headers |
| 4 | `virsh dommemstat <instance-uuid>` and compare `(available - unused) / 1024` with the emitted `ceilometer_memory_usage` | the two agree (28 MiB on the lab guest) |
| 5 | `openstack optimize audittemplate create at1 workload_balancing --strategy workload_balance` then `openstack optimize audit create -a at1 -p threshold=0.05 -p period=310 -p metrics=instance_cpu_usage` | audit reaches `SUCCEEDED` and an action plan is created |
| 6 | Repeat 5 with `-p metrics=instance_ram_usage` | audit reaches `SUCCEEDED` |
| 7 | `openstack optimize audittemplate create at2 workload_balancing --strategy allocation_balance` then `openstack optimize audit create -a at2` | audit `SUCCEEDED`; on a cluster with allocation skew the plan is `RECOMMENDED` with at least one migration |
| 8 | `grep 'Host usage for' /var/log/watcher/decision-engine.log` on **every** control node — the audit is dispatched by RPC and may run on any of them | the node hosting the instance reports non-zero `host_cpu_usage_percent` and `host_ram_usage_percent`; nodes without instances report `0.0`. A run where **every** node reports `0.0` for RAM means the metric is not resolving, even though the audit succeeded |
| 9 | `grep -c 'No values returned\|No workload in cache' /var/log/watcher/decision-engine.log` | `0` |

> **Note for QA on #8.** A Watcher audit reports `SUCCEEDED` and produces an action plan whether or not it read any metric, so "the audit passed" is not evidence. The per-host workload numbers are the check that matters.

- [x] **Handbook knowledge update** — landed in bigstack-handbook [#580](https://github.com/bigstack-oss/bigstack-handbook/pull/580), merged:

  - **Topic** — [`architecture/watcher-metrics-on-prometheus.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/architecture/watcher-metrics-on-prometheus.md) — how Watcher gets its metrics now: the four implemented metrics and where each comes from, the hardcoded names and why `metric_map.yaml` cannot redirect them, the route-prefix mismatch that 404s a direct target, the textfile collector and why not an exporter, the unit contracts, guest-used versus RSS, and the two silent-failure modes.
  - **Correction** — [`architecture/watcher-optimization-goals.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/architecture/watcher-optimization-goals.md) — repointed off Monasca, `metric_map.yaml` marked removed with the reason not to reintroduce it, and a further correction: the entry had established that the *host* mappings do not resolve, but the *per-instance* ones had no producer at all on 3.1.20.
  - **Extension** — [`known-issues/masakari-lost-libvirt-python-in-caracal-hop.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/known-issues/masakari-lost-libvirt-python-in-caracal-hop.md) — monasca's libvirt check is the same defect one service over, with no health check to catch it, so the Lesson is widened to "what is left in the old venv, and what does it import".
  - **Correction** — [`architecture/watcher-allocation-balance-strategy.md`](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cubecos/architecture/watcher-allocation-balance-strategy.md) — reads no datasource at all, came through the move untouched, and is therefore the control for isolating datasource regressions.

Eandalf-Bigstack:
#### What type of PR is this?

- feature
- bug

Watcher is the last consumer that still reads Monasca. This moves its decision engine onto Prometheus and supplies the only metrics that move did not already have, so Monasca can be deleted without taking Watcher with it. This is issue #673.

- **Backport the Prometheus datasource into Caracal.** Watcher 12.1.0 predates it; it landed upstream in 2025.1. The two upstream files are taken **verbatim**, so the backport stays trivially re-portable and drops out entirely on Epoxy
- **Publish Prometheus where the datasource can actually reach it.** It builds `host:port/api/v1/...` and `python-observabilityclient` accepts only a host and a port — there is nowhere to put our `/prometheus` route prefix, and Watcher validates the host against a hostname regex so one cannot be smuggled in. A loopback haproxy listener prefixes the path
- **Supply the two missing series.** `ceilometer_cpu` and `ceilometer_memory_usage` come from a `node_exporter` textfile written by `hex_sdk watcher_instance_metrics`, cronned by `config_nova.cpp` on compute-capable nodes. No new service, port or scrape job
- **Retire the Monasca-shaped patches.** `clients.py`, `datasources/monasca.py` and `metric_map.yaml` all go, along with PR [#548](https://github.com/bigstack-oss/cubecos/pull/548)'s unit conversion
- **Carry what remains as reviewable diffs**, the convention `core/nova` and `core/masakari` already use

- Fixes #673 — Watcher reads Prometheus end to end
- Refs #672 — the Watcher item of phase 3, and it answers what phase 2 actually had to deliver

**The first two commits are a pair and should not be split across releases.** The datasource move alone leaves `instance_cpu_usage` and `instance_ram_usage` resolving to `None`, and the failure is silent: `get_backend` only rejects a datasource when a metric maps to `None`, but `ceilometer_cpu` *is* mapped — it is simply absent from the store. So audits still SUCCEED, still produce action plans, and decide from no per-instance data with nothing logged. The unit change below has the same coupling in the other direction.

**Watcher's metric names are not configurable, and `metric_map.yaml` cannot help.** `METRIC_MAP` hardcodes `instance_cpu_usage -> ceilometer_cpu` and `instance_ram_usage -> ceilometer_memory_usage`, and `_build_prometheus_query` then dispatches on the meter name, raising `Cannot process prometheus meter` for anything it does not recognise. `metric_map_path` *does* override the map (`manager.py:182`), but renaming a Prometheus meter only breaks the dispatch. That is why the collector emits the ceilometer names rather than our own: it keeps the backported datasource byte-identical to upstream. `metric_map.yaml` is removed rather than repointed, and the option's default path is safe to leave dangling because upstream's `load_metric_map` guards with `os.path.exists`.

**Why `× 1024.0` had to go, and why it is coupled.** PR #548 added it because Monasca's `mem.used_gb` is GiB while `workload_balance` compares against `node.memory` in MiB. `ceilometer_memory_usage` is already MiB. Measured on the same node: **424.446 %** with the multiply still deployed, **0.4145 %** without. Land the datasource without the strategy change and memory reads 1024× high; land the strategy change without the datasource and the Monasca era reads 1024× low. Neither fails loudly — both just migrate wrongly.

**Memory is the guest's own figure, not the host resident set.** `available - unused`, which is exactly what Ceilometer's `inspect_memory_usage` computes for `memory.usage` and what Monasca's libvirt check computes for `vm.mem.used_gb` — the series `metric_map.yaml` pointed `instance_ram_usage` at, so an existing `threshold` carries over unchanged. `balloon.rss` is a different quantity that Monasca publishes separately as `vm.mem.resident_gb`; on a 128 MiB guest it measured **133 MiB** against a guest-used **28 MiB**. Where a guest has no virtio-balloon neither field is reported and the series is **omitted**, as Monasca omits it, rather than silently substituted.

**Why a textfile and not an exporter.** No libvirt exporter passes the provenance bar the phase 1 fleet was chosen under: the upstream one has had no commit since June 2021 and prometheus-community [declined to adopt it](https://github.com/prometheus-community/community/issues/50) — a harder fail than the `kafka_exporter` already rejected. The textfile collector needs no new service, port or scrape job, and the node's `node_exporter` target already carries the `fqdn` label. No nova call is needed for the instance label either: nova sets the libvirt domain UUID to the instance UUID, so `virsh list --uuid` is already the value `instance_uuid_label` names.

**Addressing review — where the cron lives.** It now sits in `config_nova.cpp`, gated on `IsCompute`, because it reads the local libvirt domains and so belongs to the module that runs where the domains are — the same gate nova's own compute services use a few lines below. It was in `config_prometheus.cpp` for a weaker reason: `config_watcher.cpp` returns early unless `IsControl` while the instances live on compute, and `config_prometheus.cpp`'s `Commit()` happened to run on every role. That made it work rather than made it right, and it wrote a cron on control-only and storage nodes where the generator has no `virsh`. `config_prometheus.cpp` keeps only what is genuinely its own: node_exporter's `--collector.textfile.directory` flag and the scrape. The writer stays in `sdk_watcher.sh`, since Watcher is its only consumer and `hex_sdk` resolves the module from the function's first token.

**Addressing review — the patch format.** The three modified files are now unified diffs at `<rel>.py.patch` beside the pristine `<rel>.py.orig`, with the three brand-new files installed verbatim: the `core/nova` and `core/masakari` convention. The old `cp -rf` overlay took 617 lines of whole-file copies to express 12 hunks, and would have silently overwritten a future 12.2.0 file with 12.1.0 content on a version bump. **One deliberate difference from nova and masakari:** both guard the apply with `|| exit 1` and rely on `--forward` for idempotence, but `--forward` only skips the *hunks* — measured here, it still exits 1 when every hunk is already applied, so a re-run against an already-patched tree aborts the build. It never fires for them because each build reinstalls the venv from pip first; it would fire in an incremental workspace. A `--dry-run --reverse` pre-check detects that case and skips it. Say the word if you would rather match nova byte-for-byte.

**What upstream is doing, for the record.** The Prometheus datasource is deprecated as of 2026.1 in favour of **Aetos**, a Keystone-authenticated RBAC proxy in front of Prometheus, and the default `watcher_datasources` has changed accordingly. Backporting the deprecated datasource is still right for us — Aetos is an additional service plus a Keystone endpoint, and our path is yoga → antelope → caracal → epoxy — but the next Watcher uplift should revisit it. Upstream also assumes Ceilometer + sg-core produce the instance metrics, which is why the names are `ceilometer_*`; we deliberately do not adopt those two services for two metrics.

#### Additional documentation

For the requirement "the backport loads in Caracal":

```
datasources registered: ['gnocchi', 'ceilometer', 'monasca', 'grafana', 'prometheus']
check_availability:     available
```

For the requirement "the datasource can reach Prometheus at all", which is what the haproxy listener exists for:

```
127.0.0.1:9091/api/v1/query             -> HTTP 404   (route prefix)
10.1.0.1:10904/api/v1/query             -> HTTP 404   (thanos querier, same)
127.0.0.1:9091/prometheus/api/v1/query  -> HTTP 200
127.0.0.1:9099/api/v1/query             -> HTTP 200   (the listener)
```

For the requirement "host metrics resolve", which needed no new collection:

```
jim-1cc      host_cpu_usage 45.21 %   host_ram_usage 32090844 KiB   (32 GB box)
accept-3cc   cc1 50.81 % / 28.92 GiB   cc2 34.89 % / 27.57 GiB   cc3 32.56 % / 25.61 GiB
             cc1 free -g: 28 GiB used
```

All three `fqdn` labels are discovered through the thanos queriers on the HA cluster.

For the requirement "the collector produces the two missing series", on the node hosting the instance:

```
ceilometer_cpu{resource="e3b8df03-68bb-49af-94a5-7df427079253"} 219945287000
ceilometer_memory_usage{resource="e3b8df03-68bb-49af-94a5-7df427079253"} 28
```

and nothing on the nodes with no domains. Cross-checked against libvirt directly:

```
virsh dommemstat -> available=112676 KiB  unused=83196 KiB
                    (112676-83196)/1024 = 28 MiB          == the emitted value
cpu.time         -> 439926000 ns over 181 s on 1 vcpu
                    = 0.243 %                              == what the datasource returned
```

For the requirement "the strategies behave as they did on Monasca", the same three audits run on both datasources. Node `26e1681d` is the one hosting the instance:

```
                     host_cpu_usage_percent   host_ram_usage_percent
monasca baseline     0.0 (idle VM)            0.0594
prometheus           0.0244                   0.0873
```

Both memory figures are guest-used; they differ only by sample moment (19.1 MiB vs 28 MiB of a 128 MiB guest). The other two nodes report `0.0` throughout, having no instances. Audit outcomes, identical across both datasources and unchanged after the review refactor:

```
workload_balance  metrics=instance_cpu_usage   SUCCEEDED
workload_balance  metrics=instance_ram_usage   SUCCEEDED
allocation_balance                             SUCCEEDED, action plan RECOMMENDED (1 migration planned)
```

`allocation_balance` is unaffected by the move by construction — its own header reads *"compute model only, no metrics datasource"*.

For the requirement "removing `metric_map.yaml` breaks nothing", with the file deleted on all three nodes:

```
decision engines:        active on cc1, cc2, cc3
metric_map complaints:   0 on all three
the three audits:        SUCCEEDED
readings:                unchanged at 0.0873 / 0.0244
```

For the requirement "the patch format applies and re-applies", against a tree reconstructed from the `.orig` files in the jim build container:

```
first run:   all three patches apply cleanly              rc=0
second run:  every hunk reported already applied, skipped rc=0
verbatim:    prometheus.py, prometheus_client.py, allocation_balance.py installed
```

For the requirement "it builds": clean `CXX` under `-Wall -Werror` and a clean `LD hex_config` for `config_nova.o`, `config_prometheus.o`, `config_watcher.o` and `config_haproxy.o`, with the cron literals now in `config_nova.o` and **zero** references remaining in `config_prometheus.o`.

**Not verified here, in two places.** An instance without virtio-balloon, where the memory series is omitted by design — the lab guests are Cirros images that all report balloon statistics. And the control-only path where the cron should be *absent*: every node on both lab clusters is `control-converged`, so `IsCompute` matches everywhere and no topology available here reaches the unlink branch.


core/modules/config_prometheus.cpp
```diff
@@ -684,6 +718,10 @@ Commit(bool modified, int dryLevel)
     // that is the whole point of them -- so their config and their units are handled on
     // every role, and only the scrape side below is control-only.
     WriteExporterDefaults(enabled, G(MGMT_ADDR));
+    // Same reasoning: written on every role, because the VMs it reports on are not on the
+    // control nodes. The generator no-ops where there is no virsh.
+    WriteInstanceMetricsCronJob();
+    HexUtilSystemF(0, 30, HEX_SDK " watcher_instance_metrics");
```
traviswu-bigstack: I feel the cron job should be coded in config_nova.cpp and installed only for compute-capable roles. what do you think

traviswu-bigstack (commented):
Let's use .diff for all following openstack patches and drop the .py file

core/modules/config_prometheus.cpp
Eandalf-Bigstack: Moved to `config_nova.cpp` gated on `IsCompute` — agreed, and for a better reason than I originally had. The cron runs `hex_sdk watcher_instance_metrics`, which reads the local libvirt domains, so it belongs to the module that runs where the domains are; nova gates its own compute services the same way a few lines below.

It was in `config_prometheus.cpp` because `config_watcher.cpp` returns early unless `IsControl` while the instances live on compute, and `config_prometheus.cpp`'s `Commit()` happened to run on every role. That made it work rather than made it right — and it wrote a cron on control-only and storage nodes too, where the generator has no `virsh` and produced an empty file. Those nodes now get no cron at all, and the writer unlinks it when the role or the service is disabled. `config_prometheus.cpp` keeps only node_exporter's `--collector.textfile.directory` flag and the scrape.

Verified on accept-3cc and jim-1cc: all four nodes are `control-converged` so `IsCompute` matches and each gets the cron, the node hosting the instance emits both series while the rest emit none, and re-running the three verified strategies gives readings unchanged from before the refactor (`host_ram_usage_percent 0.0873`, `host_cpu_usage_percent 0.0244`). One gap I could not close here: every node on both labs is converged, so no topology available reaches the control-only branch where the cron should be absent.

I took the second point in the same pass — the watcher patches are now unified diffs beside pristine `.orig` files, the `core/nova` / `core/masakari` convention. One deliberate difference is called out in the PR body: `--forward` still exits 1 when every hunk is already applied, so a re-run against an already-patched tree would abort the build, and I added a `--dry-run --reverse` pre-check to skip that case. Happy to match nova byte-for-byte instead if you would rather keep them identical.

SekiXu (approved):
Verified the three claims this PR rests on against the sources rather than taking them from the body.

**The patch convention checks out, byte for byte.**

- All three `.orig` files are byte-identical to upstream `python-watcher` 12.1.0 (`watcher/conf/__init__.py`, `watcher/decision_engine/datasources/manager.py`, `watcher/decision_engine/strategy/strategies/workload_balance.py`), so "pristine" is literally true and a 12.2.0 bump will conflict loudly instead of being absorbed.
- `datasources/prometheus.py` and `conf/prometheus_client.py` are byte-identical to upstream `stable/2025.1`. The "taken verbatim" claim holds, which is what makes the backport droppable on Epoxy.
- All three `.py.patch` files apply to their `.orig` at rc=0. On a second run the `patch --dry-run --reverse` pre-check in `core/watcher/watcher.mk` correctly reports every one as already applied and skips it — so the deliberate divergence from `core/nova`'s `|| exit 1` + `--forward` does what the comment says it does, and an incremental workspace no longer aborts the build.
- Applying the three patches reproduces develop's whole-file copies with exactly one difference: the `* 1024.0` removal in `workload_balance.py`. Nothing else changed under cover of the format switch.

`python-observabilityclient` is pinned — `===0.1.1` is in `core/heavyfs/os-caracal-pip-upper-constraints.txt`, so the `-c` on the `pip install` covers it and the unpinned-looking name in `watcher.mk` is not actually unpinned.

**Three non-blocking notes.** The first two are already fixed by #1457, so they only surface if this ships on its own:

1. `core/sdk_sh/modules/sdk_watcher.sh` — the `# HELP ceilometer_memory_usage Resident memory of the instance in megabytes.` line contradicts both the code (`available - unused`) and the file's own header, which explicitly rejects `balloon.rss` *because* it is the resident set. It also says megabytes where the header says MEBIBYTES. #1457 rewrites this line correctly.
2. Same file — `$(( (avail - unused) / 1024 ))` truncates to whole MiB (28.8 → 28). On the 128 MiB lab guest that is nearly a percentage point. #1457's awk does this in floating point.
3. Still present in both: a `virsh list` timeout is indistinguishable from "no running domains" — the `for` loop just yields nothing, and the good textfile is then overwritten with headers only. `rate()`/`avg_over_time` absorb a one-sample gap so the impact is small, but a failed read and an empty host are worth telling apart.

Style nit: the `watcher_prometheus` frontend is emitted above the comment block in `config_haproxy.cpp` that says "Add new sections here, at the end." It parses fine and it is after both `cube_cos` frontends, so it does not hit the hazard that comment describes — but the instruction and the placement now disagree.

LGTM. This one should merge first; #1457 is stacked on it.

traviswu-bigstack (approved):

PR #1450, opened 2026-09-09
Collect and report per-VM and per-pool storage usage

### LINKED ISSUE #1449: [Feature] Report per-VM and per-pool storage usage, and alarm on capacity
## Description

Collect storage usage on a fixed interval and make it readable — per VM and per backend — then
alarm when a pool or a VM filesystem approaches full.

Three numbers are collected and kept deliberately separate, because they measure different things
and disagree by large factors in both directions:

| Number | Source |
|---|---|
| Provisioned | nova flavor / cinder volume size |
| Physically allocated | `rbd du`, `qemu-img`, or the array |
| In-guest filesystem used | qemu-guest-agent |

Thin provisioning never shrinks, so allocated stays high after a guest deletes files unless discard
passed through end to end; and the guest figure counts nothing outside a mounted filesystem (swap,
raw devices handed to a database, LVM extents, snapshots). A single blended "used" column would be
neither comparable between VMs nor summable against the cluster total.

Reduction ratios are **pool-level only**: BlueStore's compression counters are per-pool and Ceph
exposes no per-image stat, so a per-VM compression ratio is not obtainable. Neither VMware (vSAN
reports it cluster-wide) nor Nutanix (Storage Container level) ships one either.

Scope is the collector, the CLI surface and the alarms. Serving it to the GUIs is follow-on work:
cube-cos-api metrics surface, then the Horizon panel and the cubecmp portal columns. Dell SC and
NetApp adapters, and enabling Ceph compression, are each separate.

## Definition of Done (DoD)

- [ ] Per-disk and per-pool usage sampled on a tunable interval by one writer per cluster, with
      per-VM in-guest usage where the guest agent answers
- [ ] Samples readable via `hex_cli -c storage -c usage [<instance>]`, reporting sample age
- [ ] Pool capacity and over-subscription events fire on threshold crossing, not per sample
- [ ] VM filesystem alarm fires per filesystem, delivered through the existing per-project alert
      triggers and receivers
- [ ] Backends with no adapter of their own still report honest pool totals
- [ ] Unit tests pass with no cluster required
- [ ] Passing a code review

## Technical Notes

- **Feature Owner:** travis.wu
- **Frontend:** none in this issue (Horizon panel and cubecmp columns are follow-on)
- **Backend:** `sdk_storage.sh` collector + adapters; `hex_config` tunables and cron;
  `hex_cli` command; kapacitor template and event codes
- **Infra:** writes route through Kapacitor's influx-compatible endpoint, which relays to influx on
  every control node — writing to influx directly repeats the bypass recorded in cubecos#672

Risks and constraints worth review:

1. **The VM alarm must be a `batch` kapacitor task.** `kapacitor.conf` disables subscriptions on
   every influxdb target, so a `stream` task never receives a point. This also means the shipped
   per-VM cpu/memory alert templates cannot fire — a pre-existing defect, worth its own issue.
2. **Replication factor is derived per pool, never assumed.** Pools legitimately differ; a
   hardcoded 3 makes every figure on a `size=2` pool 50% wrong, and it fails silently.
3. **Attribution enumerates per project.** No `openstack` list command exposes a volume's or
   server's owning project at any `--long` level, and `-c Project` is silently dropped.
4. **No Cinder driver is patched.** Pool capacity is already published by every driver; per-volume
   footprint comes from the backend, reusing credentials already in the storage model.
5. Compression stays off, so the reduction ratio reports as unavailable rather than 1.0 — enabling
   it is a separate, benchmark-gated decision.

## UI/UX Mockups (Optional)

None

## QA Verification

|---|---|---|
| 1 | `hex_cli -c storage -c usage` | one row per pool with `stored`, `raw`, `capacity`, `provisioned`, `repl` and a `sampled=` timestamp |
| 2 | `hex_cli -c storage -c usage <instance-uuid>` | that VM's disks with `provisioned`/`allocated`/`snapshots`, plus a `guest-fs` line, or `guest-fs no agent data` |
| 3 | `hex_cli -c storage -c usage <nonexistent-uuid>` | exits cleanly reporting no agent data; no traceback, no error |
| 4 | Compare `raw`/`stored` from step 1 against `ceph osd pool get <pool> size` | the reported `repl` equals the pool's configured size, per pool |
| 5 | Run the collector on a control node that does not own the VIP | returns immediately having written nothing |
| 6 | Set `storage.usage.threshold.soft` to a value below current pool usage, run the collector twice | one `STO00001W` event, not two |
| 7 | Let pool usage fall more than 5 points below soft, run again | one `STO00003I` clear event |
| 8 | Fill a filesystem above 85% inside a VM whose guest agent answers | `DSK00005W` in `events.instance`, naming project, VM and mountpoint |
| 9 | `hex_sdk -v alert_vm_event_list` | lists `DSK00006C` and `DSK00005W` with their thresholds alongside the cpu/memory codes |
| 10 | `hex_cli -c cluster -c check` after install | all service groups `ok` |

- [ ] **QA Status declared at Done**

- [ ] **Handbook knowledge update** — `kb/cubecos/architecture/storage-usage-collection.md`
      (the four RBD pools, per-pool replication, the attribution constraint, the single-writer
      rule, the Kapacitor write path) plus
      `kb/cubecos/known-issues/kapacitor-stream-tasks-inert-without-subscriptions.md`.
      Both are written and staged in the paired handbook PR.

traviswu-bigstack:
#### What type of PR is this?

/kind feature

Samples storage usage on a tunable interval, exposes it through `hex_cli`, and alarms on pool
capacity and VM filesystem fullness.

Three numbers are collected and deliberately kept as three, because they measure different things
and disagree in both directions: **provisioned** (nova/cinder), **physically allocated** (the
backend), and **in-guest filesystem used** (qemu-guest-agent). Thin provisioning never shrinks, so
allocated stays high after a guest deletes files unless discard passed through end to end; the guest
figure counts nothing outside a mounted filesystem. On one measured VM those read 171.8G / 37.8G /
21.1G — a single blended "used" column would be neither comparable between VMs nor summable against
the cluster total.

One adapter per backend, all emitting the same record shape:

- **ceph** — `rbd du` across the four pools CubeCOS creates (`ephemeral-vms`, `cinder-volumes`,
  `glance-images`, `k8s-volumes`), plus `ceph df detail` for stored-vs-raw and the compression
  counters. `glance-images` is shared overhead and never charged to a VM, or COW clones
  double-count their parent.
- **nfs** — `qemu-img` on the share (sparse raw, per the builtin model).
- **cinder** — pool capacity every driver already publishes, so a backend with no adapter of its
  own still reports honest totals rather than a blank.
- **guest** — `guest-get-fsinfo`, deduped by device and with pseudo filesystems dropped.

Per-VM compression ratio is not obtainable — BlueStore's counters are per-pool and Ceph has no
per-image stat — so reduction ratios are pool-level only. **No Cinder driver is patched.**

Fixes #1449

Five decisions worth a look, each of which fails *silently* if done the obvious way:

1. **Appended to `sdk_storage.sh` rather than a new module file.** `hex_sdk` resolves a module by
   globbing `sdk_<first-token>*.sh`; a second matching file makes `[ -e f1 f2 ]` error and return
   false, which would push every `storage_*` call — including the existing multipath and
   disk-selection ones — down the source-all-modules fallback.
2. **The replication factor is derived per pool, never assumed.** Pools legitimately differ, and a
   hardcoded 3 makes every figure on a `size=2` pool 50% wrong with no error.
3. **Attribution enumerates per project.** No `openstack` list command exposes a volume's or
   server's owning project at any `--long` level, and `-c Project` is silently dropped, so the
   obvious join always yields null.
4. **Single writer via VIP ownership, tested locally against the node name.** `is_vip_active()`
   cannot express this — it reports the resource Started *anywhere*, true on every control node, so
   using it would run the collector three times over.
5. **Writes go to the Kapacitor write proxy on `:9092`, not to InfluxDB.** That one ingress both
   relays the write to the peer control nodes and feeds the stream tasks the alert templates are
   built on; writing to `:8086` reaches neither and repeats the bypass that left ceph metrics on a
   single node (#672). `influx -execute "INSERT ..."` is unusable regardless — one line per call.

**Rebased onto the monasca retirement.** #1457 removes `MONASCA_DB` and its relay task, so the
per-filesystem series now lands in `telegraf.def` and the alarm is a **stream** task there — the
same shape and ingress the cpu/memory templates move to in that PR, rather than a parallel
mechanism. Verified on a 3-node lab after the change: 24 disk rows and 30 filesystem rows landing
identically on all three nodes, nothing written to `monasca`, and the stream task consuming them
(`processed=30`, 27 past the tenant filter, alerts delivered).

**Merge #1457 first** — both PRs touch `core/main/cube.mk`, and this one assumes the retirement.

**Both original blockers are now cleared.** `hex_config` was built at the cluster's own sha
(`d2ee24c5`, not `develop`) and hot-deployed with `cluster check` unchanged at 27 `ok`; and the NFS
adapter has run against a real NFS-backed Cinder volume — which is how the `cinder_get_storage`
silent failure below was found.

**Why pool thresholds are tunable and VM ones are not.** The VM disk alarm hardcodes 75/85/95 in
the tick, exactly like `tpl_alert_vm_cpu`, `tpl_alert_vm_mem` and the host disk alarm — none of
those declare threshold vars, and `alert_vm_event_list` scrapes the literals back out for display.
The pool alarms (`STO0000x`) do take tunables, deliberately: "filesystem 95% full" means the same
thing everywhere, "pool 85% full" does not, and the over-subscription ratio has no natural constant.

**Changed under review.** Ceph pool figures come from ceph-mgr's exporter (`ceph_pool_*`) rather
than a second `ceph df detail`, and the duplicate `cube_storage_pool_*` series for Ceph are gone.
The sampler is a telegraf exec input, not a cron entry, printing line protocol under `-f line` like
`network_virtual_router_stats`. Tunables are read through `hex_tuning` at the point of use, so the
bespoke env file is gone — it was also hiding a dead tunable, `storage.usage.exact` never took
effect. `rbd du` now runs `--exact`: the object map over-reported 87 GiB against the 11.4 GiB
`ceph df` says the pool stores. The VM disk alarm falls back to the block layer when no guest agent
answers, so an agentless instance can alarm at all.

**Grafana, added after review feedback.** Grafana reads Prometheus, so the same collection pass now
also writes a node_exporter textfile — the influx line protocol exists only because Kapacitor cannot
read Prometheus and the alarm needs a stream. Same shape as `instance_metrics_collect`: `.tmp` then
rename, and a null is omitted rather than written as `0` (an absent series reads as "not measured",
a `0` reads as "measured, empty").

The shipped `Storage` dashboard covered ceph cluster and OSD health and nothing per-pool or per-VM.
It gains a **VM and pool usage** row: top 10 VMs by allocated storage, top 10 filesystems by usage,
top 10 VMs by thin-provisioning gap, pool stored vs raw, and pool provisioned over capacity.

Verified on a 1cc lab: 226 series in the textfile, and Prometheus serving them —
`topk(10, sum by (vm_name) (cube_storage_disk_allocated_bytes))` returns real VMs.

166 assertions across 9 self-contained suites; they mock `rbd`, `ceph`, `qemu-img`, `virsh`,
`openstack` and `curl`, so `bash core/sdk_sh/tests/test_storage_usage_*.sh` needs no cluster.

#### Additional documentation

```docs
kb/cubecos/architecture/storage-usage-collection.md
kb/cubecos/known-issues/kapacitor-stream-tasks-inert-without-subscriptions.md
```


traviswu-bigstack:
### Companion PRs

This collector is one of five pieces; the other four are open as drafts:

| Repo | PR | What |
|---|---|---|
| cube-cos-openapi | bigstack-oss/cube-cos-openapi#115 | `storageUsage` in the `metricType` enum |
| cube-cos-api | bigstack-oss/cube-cos-api#663 | serves the rank; **also repoints 6 VM statements off the retired monasca bucket** |
| cube-cos-ui | bigstack-oss/cube-cos-ui#875 | "Disk Usage" in the VM ranking panel on `/home/chart` |
| skyline-console | bigstack-oss/skyline-console#134 | Usage column on the instance list, Allocated on the volume list |

Merge order is spec → api → ui, with skyline-console independent. This PR carries the `skyline.yaml.in` `prometheus_endpoint` fix that skyline-console#134 depends on.

### Worth flagging from lab validation

**All VM metrics were already broken on Prometheus-migrated clusters**, not just the new one. Six statements in cube-cos-api still queried `Bucket("monasca")` after #1457 retired it, so every VM rank and history — cpu, memory, iops, network — returned empty and the VM Ranking card was blank. Fixed in cube-cos-api#663.

**Block allocation is not filesystem usage, and cannot be substituted for it.** Measured here: the block layer reported 85% on an instance whose guest reported 14.8%. Running `fstrim` in the guest reported 32.1 GiB trimmed with exit 0 and moved `rbd du` by nothing — verified by counting RADOS objects rather than trusting the object map. `rbd_skip_partial_discard` frees an object only when the entire 8 MiB is discarded, and free space on an aged filesystem is too scattered for that. This is why the collector keeps the three numbers separate and carries `confidence`, and why the UI marks agent-less rows as estimates instead of colouring them on the alarm scale.

core/modules/config_cinder.cpp
```diff
@@ -124,6 +131,12 @@ CONFIG_TUNING_SPEC_STR(NOVA_USERPASS);
 // parse tunings
 PARSE_TUNING_BOOL(s_enabled, CINDER_ENABLED);
 PARSE_TUNING_BOOL(s_debug, CINDER_DEBUG);
+PARSE_TUNING_BOOL(s_usageEnabled, STORAGE_USAGE_ENABLED);
+PARSE_TUNING_UINT(s_usageInterval, STORAGE_USAGE_INTERVAL);
+PARSE_TUNING_UINT(s_usageSoft, STORAGE_USAGE_SOFT);
+PARSE_TUNING_UINT(s_usageHard, STORAGE_USAGE_HARD);
+PARSE_TUNING_STR(s_usageOversub, STORAGE_USAGE_OVERSUB);
+PARSE_TUNING_BOOL(s_usageExact, STORAGE_USAGE_EXACT);
```
Eandalf-Bigstack: These four tunings are not used.

Eandalf-Bigstack (commented):
LGTM, we would be merging this PR after v3.1.20, into v3.2.0 instead.

Eandalf-Bigstack (approved):

PR #1457, opened 2026-09-09
feat(monasca): move the last consumers onto prometheus and remove monasca

Eandalf-Bigstack:
#### What type of PR is this?

- feature
- cleanup

#### What this PR does / why we need it

Monasca's last three consumers move onto Prometheus, and then Monasca is removed. This is phase 3 and phase 4 of issue #672.

- **The six API health checks read blackbox probes.** `health_{nova,glance,cinder,heat,octavia,designate}_check` read Monasca's `http_status` out of InfluxDB. `blackbox_exporter` has been probing those same six ports through the VIP since the exporter fleet landed; each target now carries a `service` label so the lookup stays a name rather than a port number
- **The two Monasca dashboards are rebuilt on Prometheus.** *Instance* and *Top Instances*, 23 panels, plus the three template variables. The two `influxdb-monasca` datasources are listed for deletion, not just dropped, so an already-provisioned cluster loses them too
- **The per-instance collector grows from 2 series to 12.** cpu, cores, memory, per-device disk bytes and requests, per-device network bytes and packets — everything the dashboards read. `sdk_watcher.sh` becomes `sdk_instance.sh` accordingly: Watcher is no longer the only consumer
- **The two per-project Kapacitor VM alert templates are re-sourced off `monasca.def`.** Kapacitor cannot read Prometheus, so the same libvirt read also posts influx line protocol to the Kapacitor write proxy, under Monasca's exact measurement, tag and field names — the templates change by one line
- **Monasca is removed.** `core/monasca`, `config_monasca.cpp`, the `HEAVY_COMPONENTS` entry, the haproxy listener, the InfluxDB database, the Kapacitor relay tasks, the filebeat input, the two Kafka topics, the `cli_cluster` service entry, and the whole `hex_sdk` surface
- **Senlin's retirement moves onto the same mechanism**, which is where the review discussion took this: `management cleanup cleanup_senlin` becomes `migrate_senlin_retire`, and picks up the database and grants the CLI command never removed
- **The per-VM `hex_sdk stats_*` consumers are repointed too** — `stats_topten_vm`, `stats_vm_chart` and `stats_inactive_vm_drop`, which the UI's Top Instances and per-instance charts are built on. They were missed by the first pass; see the review note below

#### Which issue(s) this PR fixes

Fixes #672 — phase 3 complete, and phase 4 for the four services

- Builds on #673, which moved Watcher itself

#### Special notes for your reviewer

[...]

PR #1249, opened 2026-08-11
feat(openstack): move the openstack cli and its plugins into the antelope venv

Eandalf-Bigstack:
> **Stacked on top of #1229.** This branch (`jim.lin/feat/openstack-a-wrapup-1`) is cut from
> `jim.lin/feat/openstack-a-wrapup`, so until #1229 merges the diff against `openstack-develop`
> will also show that PR's commits. Please review only the 26 commits on top, i.e. the diff
> from `jim.lin/feat/openstack-a-wrapup` to `jim.lin/feat/openstack-a-wrapup-1`.

#### What type of PR is this?

- feature
- cleanup
- bug fix

#### What this PR does / why we need it

- Move the **openstack CLI** and every one of its **osc plugins** into the python 3.10
  `/opt/openstack-antelope` venv. `/usr/bin/openstack` was the yoga `python3-openstackclient`
  rpm, so the CLI ran on `#!/usr/bin/python3` — and a stevedore entry point is only visible to
  the interpreter it was installed under. That single shebang is what forced designate, heat,
  manila and octavia to hold their clients as yoga rpms and cyborg, watcher, monasca and
  masakari to install a second copy into 3.9, each behind a comment saying it could go once the
  CLI moved. This is that move, and it is the last of #601.
- Empty the system python of openstack entirely — `ospurge` follows the CLI into the venv, and
  `python-scciclient` goes (ironic runs in the venv and deliberately does not install it there).
- Point the **dnf openstack repo** at antelope, and **promote the `NEXT_*` build variables** to
  the current ones, leaving the `NEXT_*` names blank for the next hop.
- Fix the two pre-existing octavia defects #1206 recorded and deliberately left, plus the
  neutron `--dns-domain` failure that has been carried unticketed since 2026-07-25.

#### Which issue(s) this PR fixes

- Close #601
- Refs #616, #593

#### Special notes for your reviewer

**Two behaviour changes worth a second look.**

- **`config_octavia.cpp` now derives `[health_manager] bind_ip` from `cubesys.mgmt.cidr`**
  instead of the hardcoded `172.16.0.<octet4>`. On this cluster that is `10.254.0.11` against a
  configured `172.16.0.11` — an address no interface carries. It never failed loudly because
  `net.ipv4.ip_nonlocal_bind` is 1, so the bind succeeded and heartbeats routed out the
  *provider* gateway; load balancers sat at `operating_status OFFLINE` while forwarding fine.
[...]

Issue #637, opened 2026-01-27
[Story] [Parallel Group] Upgrade Ironic from Antelope to Caracal


jdjgya:
## Description

As a CubeCOS operator, I want Ironic be in Caracal so that we could upgrade to Caracal.

Issues labeled with "Parallel Group" could be tasked in parallel.

## Acceptance Criteria

> A detailed list of conditions that must be met for the story to be accepted. These statements must focus on functional behavior and constraints rather than specific code-level implementation.

- [x] Migrate configs
- [x] Pass health check
- [x] Ignore the new features introduced in the new version

## Story Points: 3

>| Points | Effort / Resources | Description |
> | -- | -- | -- |
>| 1 | Low | Low complexity; can be resolved easily.
>| 3 | Medium | Moderate workload; requires some focus, but risks are well-managed.
>| 5 | High | High complexity; difficult to implement, requiring significant time and effort.
>| 8 | Very High | Extreme complexity or high uncertainty. If possible, this should be divided into multiple user stories.

## QA Verification

> How QA verifies this story once it ships — write **checkable claims**: the exact
> command or click path plus the expected outcome. "Verify it works" is not a checkable
> claim. These claims are what the Qase test suite is generated from, so write them for a
> reader who has not seen the code.

> [!IMPORTANT]
> **Ironic ships disabled (`ironic.enabled` defaults to `false`), and every check below needs it enabled.** `health_ironic_check()` short-circuits on `x$T_ironic_enabled != xtrue`, so a green `cluster check` on a default install proves nothing about this story. Enable ironic and let the commit bring up `openstack-ironic-{api,conductor,inspector}`, `ironic-neutron-agent` and `openstack-nova-ironic-compute` before starting. Steps 21–22 additionally need **#1360 / PR #1362** landed (the `role:service` grants); without it step 22 returns an empty list.

| # | Command / steps | Expected outcome |
|---|---|---|
| 1 | `/usr/bin/ironic-api --version; /usr/bin/ironic-conductor --version` | `24.1.5` twice |
| 2 | `/usr/bin/ironic-inspector --version` | `12.1.1` |
| 3 | `ps -eo args \| grep -E "bin/(ironic-api\|ironic-conductor\|ironic-inspector\|ironic-neutron-agent)" \| grep -v grep` | every process runs as `/opt/openstack-caracal/bin/python3.11 /usr/bin/ironic-…` |
[...]

PR #1467, opened 2026-09-09
feat(openstack): finish the caracal hop and promote it to the openstack release

### LINKED ISSUE #625: [Task] Language - Python upgrade from 3.10 to 3.11

Eandalf-Bigstack:
#### What type of PR is this?

- cleanup
- feature

Wraps up the Antelope → Caracal migration, the same shape #601 / [#1249](https://github.com/bigstack-oss/cubecos/pull/1249) took for Yoga → Antelope. This is #625.

- **Move `ospurge` into the caracal venv.** It is the last occupant of `/opt/openstack-antelope` — not a service, so no component hop carried it. #601 moved it out of the system python 3.9; nothing has moved it since
- **Retire the antelope venv, and the python 3.10 built from source to run it.** That build is a `./configure --enable-optimizations` plus `make altinstall` inside the chroot, so it is also the slowest single step in `core/heavyfs`
- **Promote the `CARACAL_*` variables** into the block under `# openstack version`, and rename all 350 references across 21 component makefiles. `OPENSTACK_HOME_DIR` is `/opt/openstack-caracal` and `PYTHON_VER` is `3.11`, which is what makes the python bump this issue asks for a consequence of the hop rather than a change of its own
- **Follow the dnf openstack repo across**, which changes the ceph release rpm that comes with it

Fixes #625 — python 3.10 → 3.11, by finishing the venv it rides on

- Refs #594 — the component upgrades this wraps up

**Stacked on [#1457](https://github.com/bigstack-oss/cubecos/pull/1457) / `jim.lin/feat/replace-monasca-6`.** That PR is not merged, so the diff against `develop` shows its 13 commits too. Only the three from `afda6b16` onward belong here.

**The stacking is load-bearing, not just a rebase convenience.** #1457 is what removes Monasca, and Monasca was an antelope-venv occupant. If this merges first, the build drops a venv Monasca still needs. **Merge #1457 first.**

**ospurge ports without code changes, but the patch it carries does not go away.** Upstream ospurge assigns `obj['container_name']`, and openstacksdk 1.0 made `Resource.__setitem__` reject undeclared keys. Caracal pins 3.0.0, and it was worth checking whether the hop had made the patch unnecessary — it has not. Unpatched under 3.0.0 the failure reproduces unchanged, so the patch directory follows the venv rather than being dropped.

**The `--no-deps` install has two undeclared dependencies that happen to be satisfied.** ospurge imports `pbr` and `six` and declares neither; both are already in the caracal venv (`six===1.16.0` is in caracal's own upper-constraints, every service pulls `pbr`). A venv holding ospurge alone fails at `import ospurge.main` — which is how my scratch venv first behaved. Recorded in the makefile rather than left to be rediscovered.

**`PROJ_PIP_CONSTRAINT` deliberately does not follow the promotion.** #601 decoupled it from the release name because the local file carries CVE pins the openstack constraint files do not (`pillow 11.3.0`, `waitress 3.0.2`); letting a rename move it would have put pillow back to 9.2.0 and waitress to 2.1.2, i.e. CVE-2023-50447 and CVE-2024-49768. Verified it still resolves to `rootfs-pip-constraints.txt` after the promotion, and the comment now says the rule holds for every hop rather than reading as a one-off.

**The ceph repo change is a consequence, and it would have failed the build rather than no-op.** The openstack release rpm decides which ceph release rpm comes with it: antelope required `centos-release-ceph-quincy`, caracal requires `centos-release-ceph-reef`. The quincy repo was disabled after the fact because it arrived unwanted and could not be erased without leaving a dependency unsatisfied. Caracal never installs it — and `dnf config-manager --set-disabled` on a repo that does not exist **exits 1** (`No matching repo to modify`), so leaving the line in a recipe would break the build. `centos-release-ceph-reef` is still named explicitly, so a future release pairing with squid shows up as a conflict rather than a silent ceph bump.

**Three stale comments corrected while in the neighbourhood**, each wrong before this PR: `project.mk` said ceph's rados/rbd bindings were still an antelope-venv occupant (`ceph.mk` installs those wheels only into caracal), `core/heavyfs` said ironic runs in the antelope venv (it moved in #637), and the long setuptools NOTE sat above the antelope venv creation and is re-anchored to caracal's — its reasoning about pinning at creation, and everything about pip's PEP-517 overlay and the `--no-build-isolation` horizon's nine sdist-only dependencies need, is still exactly true.

**What was not verified: none of this has been through a rootfs build.** The labs run a pre-#625 image, so the antelope venv and python 3.10 still exist there — this removes them from the build, not from a running node. The riskiest step is the one I could least exercise: whether dropping the 3.10 build leaves anything unbuilt. That was checked by grep and by a real `make` parse of every component makefile, not by building an image.

#### Additional documentation

Verified on `jim-1cc` (single control) and `accept-3cc` (3-node HA), plus the `jim` build jail for anything that is a build-time question.

For the requirement "nothing is left running on python 3.10", which is the acceptance item:

```
                          jim-1cc   accept-3cc
/opt/openstack-caracal/bin/python3.11    55           56     service processes
/opt/openstack-antelope/...               0            0
```

For the requirement "ospurge runs from the caracal venv", installed into the real venv exactly as `appfw.mk` does it:

```
[cc1 ~]# head -1 /opt/openstack-caracal/bin/ospurge
#!/opt/openstack-caracal/bin/python3.11
[cc1 ~]# ospurge --dry-run --verbose --purge-project=ospurge-victim
  ...Going to delete Object 'sdkprobe.txt' from Container 'log'
  rc=0
```

16 objects enumerated on jim-1cc, 2264 resources on accept-3cc, both rc=0.

For the requirement "the swift patch is still needed, and still works", measured both ways in a scratch 3.11 venv before touching the real one:

```
unpatched, openstacksdk 3.0.0:
  KeyError: 'container_name is not found. openstack.object_store.v1.obj.Object
             objects do not support setting arbitrary keys through the dict interface.'

patched, openstacksdk 3.0.0:
  sample: Object 'apiprobe.txt' from Container 'log'
  TOTAL: 16
```

For the requirement "every promoted variable resolves", from a real `make` run in the build jail:

```
OPENSTACK_RELEASE            = caracal
OPENSTACK_HOME_DIR           = /opt/openstack-caracal
PYTHON_VER / PATCH           = 3.11 / 3.11.15
OPS_GITHUB_BRANCH_01/02      = stable/2024.1 / unmaintained/2024.1
OPENSTACK_PIP_CONSTRAINT     = .../core/heavyfs/os-caracal-pip-upper-constraints.txt
OPENSTACK_INSTALLED_PIP_CONS = /opt/openstack-caracal/os-caracal-pip-upper-constraints.txt
PROJ_PIP_CONSTRAINT          = .../core/heavyfs/rootfs-pip-constraints.txt
CARACAL_OPENSTACK_HOME_DIR   = []
```

The last line is the point: the old name expands to nothing, which is the proof that no reference to it survives. `PROJ_PIP_CONSTRAINT` is the second point — unchanged, so the CVE pins hold.

For the requirement "every component makefile still parses together":

```
parsed 49 component makefiles
OSPURGE_SRCDIR             = /opt/openstack-caracal/lib/python3.11/site-packages
HORIZON_VENV_SITE_PACKAGES = /opt/openstack-caracal/lib/python3.11/site-packages
```

For the requirement "the dnf repo swap is clean", done by hand on jim-1cc rather than assumed:

```
before: centos-release-openstack-antelope, centos-release-ceph-quincy, centos-release-ceph-reef
        repos: centos-openstack-antelope, centos-ceph-reef

install centos-release-openstack-caracal ; remove centos-release-openstack-antelope
        -> also removed centos-release-ceph-quincy as an unused dependency

after:  centos-release-openstack-caracal, centos-release-ceph-reef
        repos: centos-openstack-caracal, centos-ceph-reef
        dnf check rc=0
```

and the reason the disable line had to go rather than being left harmless:

```
[cc1 ~]# dnf config-manager --set-disabled centos-ceph-quincy
Error: No matching repo to modify: centos-ceph-quincy.
rc=1
```

The node was restored to its build state afterwards, `dnf check` rc=0.

For the requirement "the antelope venv is genuinely empty before it is deleted": `project.mk` claimed two remaining occupants, ospurge and ceph's rados/rbd bindings. `ceph.mk` installs those wheels only into the caracal venv, and the antelope venv on jim-1cc has neither module — so the comment was stale and ospurge was the only one.


SekiXu (changes requested):
Reviewed the three commits from `afda6b16` that belong to this PR. **I found nothing wrong with them** — the block is entirely that this branch contains #1457, which I have asked for changes on. Repeating the short version so this page stands on its own: #1457 leaves `check_service monasca` at `core/main/proj_functions:969` calling a `health_monasca_check` it deletes, and `hex_sdk`'s usage banner goes to stdout, which corrupts the `system.services` JSON for every service on every control node; and `cube-cos-api` still reads `bucket: "monasca"` in 12 places. Details on #1457.

What I checked here, since the promotion is the kind of change that is either completely done or quietly half done:

- **The rename is complete.** `CARACAL_` appears exactly once in the tree after this PR, and it is prose in `project.mk:94`. Your `CARACAL_OPENSTACK_HOME_DIR = []` check was the right one to run.
- **No live antelope paths survive.** All four `openstack-antelope` hits are comments (`config_cyborg.cpp:258`, `config_glance.cpp:50`, `neutron.mk:52`, `nova.mk:54`). `cyborg_sudoers` authorises `/opt/openstack-caracal/bin/privsep-helper`, so the sudoers/`helper_command` pairing that has to stay byte-identical still does.
- **The deleted constraints file has no readers.** Zero references to `os-antelope-pip-upper-constraints.txt` remain.
- **`PROJ_PIP_CONSTRAINT` did not follow the promotion**, which is the pillow/waitress CVE point, and the comment now states the rule for every hop rather than for antelope specifically. Good change.
- **`-b "$(OPS_GITHUB_BRANCH_01)"` is genuinely inert**, as `project.mk:97` claims: `core/heavyfs/Makefile:224` is the only caller, and the only git-form `ROOTFS_PIP` entry is rancher's `client-python.git@master`, which carries its own ref.

One thing worth recording rather than fixing here: `OPS_GITHUB_BRANCH_01` is now `stable/2024.1`, and the GitHub mirror has deleted its `stable/*` branches — only `unmaintained/*` remains. It is unused today for the reason above, so this is not a defect in this PR, but the next hop that actually clones from that variable will hit it. Might be worth a note beside the pair.

Nothing else here needs anything from me. Happy to approve once #1457 is sorted and this rebases onto it.

SekiXu (approved):
Approving. #1457 is sorted and this contains its fixed head (`0f831306` is an ancestor), so the block I had is gone.

Re-ran my own checks against this head rather than carrying the earlier verdict, since the branch rebased:

- **The rename is complete.** `CARACAL_` has exactly one hit left and it is prose in `project.mk:94`.
- **No live antelope paths.** All four `openstack-antelope` hits are comments (`config_cyborg.cpp:258`, `config_glance.cpp:50`, `neutron.mk:52`, `nova.mk:54`), and `ANTELOPE` as a variable is gone entirely.
- **The deleted constraints file has no readers**, and `PROJ_PIP_CONSTRAINT` still points at `rootfs-pip-constraints.txt` — the pillow/waitress point holds.
- **The twenty service `.mk` files are a pure rename.** Normalised the five `CARACAL_*` names and the three patch-dir spellings, then paired every added and removed non-comment line: zero unpaired. Nothing semantic is riding along with the promotion, which is the failure mode this kind of change invites.
- **`PYTHON_VER` / `PYTHON_PATCH_VER` are consistent across all thirteen consumers**, and `core/ceph/ceph.mk:77` keeps its own `CEPH_PYTHON_VER` rather than silently coupling to the openstack value.

The quincy handling is right for the reason given: nothing to disable once caracal pulls reef, and leaving the line in would fail rather than no-op. Keeping `centos-release-ceph-reef` named explicitly so a future squid pairing surfaces as a conflict is the right instinct.

Still worth recording, and still not a defect here: `OPS_GITHUB_BRANCH_01` is `stable/2024.1` and the GitHub mirror has deleted its `stable/*` branches — only `unmaintained/*` remains. Inert today because the only git-form `ROOTFS_PIP` entry carries its own ref, but the next hop that actually clones from that variable will hit it.

Issue #767, opened 2026-04-14
[User Story] Move Kapacitor Capabilities to Prometheus Alertmanager


Eandalf-Bigstack:
## Description

As a CubeCOS operator, I would like to keep having the event alert mechanism in Antelope after we change the metric collecting mechanism so that we could still have event alerts.

Any alternatives to Prometheus Alertmanager is welcome, as long as it could provide event alert mechanism with our current monitoring pipeline.

## Acceptance Criteria

> A detailed list of conditions that must be met for the story to be accepted. These statements must focus on functional behavior and constraints rather than specific code-level implementation.

- [ ] Set up Prometheus Alertmanager
- [ ] Migrate existing data collecting and event alerting mechanisms from Kapacitor to Prometheus Alertmanager
- [ ] Remove Kapacitor

## Story Points: 5

>| Points | Effort / Resources | Description |
> | -- | -- | -- |
>| 1 | Low | Low complexity; can be resolved easily.
>| 3 | Medium | Moderate workload; requires some focus, but risks are well-managed.
>| 5 | High | High complexity; difficult to implement, requiring significant time and effort.
>| 8 | Very High | Extreme complexity or high uncertainty. If possible, this should be divided into multiple user stories.

steven-chiu-bigstack:
According to @traviswu-bigstack, we should move it to the next release due to huge development effort.

Eandalf-Bigstack:
## Two findings from spike #646 that change the shape of this story

Surveyed the live pipeline on jim-1cc and accept-3cc (`CUBE_3.1.20`). Both bear directly on AC #3, *"Remove Kapacitor"*.

### 1. Kapacitor is not only an alerting component — it is the write proxy and the HA replication layer

This is the one that resizes the story. Grep the pipeline for InfluxDB's port and you find almost nothing: **nothing writes InfluxDB `:8086` except `ceph-mgr`'s influx module.** Every other writer — `monasca-persister`, all three telegraf logstash persisters, all four logstash event mappers — POSTs InfluxDB line protocol to **Kapacitor on `:9092`**:

```
http://<shared_id>:9092/write?db=…&rp=…
```

Kapacitor then relays into InfluxDB through eight generated `relay_<db>_<rp>` stream tasks (`monasca`, `telegraf`, `ceph`, `events` × `def`, `hc`), each ending in `influxDBOut()`.

[...]

PR #1468, opened 2026-09-09
feat(kapacitor): upgrade to 1.8.6 from the signed repo, and fix the key that verifies it

### LINKED ISSUE #648: [Story] [Optional] Upgrade Kapacitor from 1.5.7 to 1.8.6

Eandalf-Bigstack:
#### What type of PR is this?

- feature
- bug fix

Issue #648, after its rescope: the InfluxDB major upgrade it originally asked for is [not being done](https://github.com/bigstack-oss/cubecos/issues/648#issuecomment-5546093221), so this is the Kapacitor version bump on its own.

- **Kapacitor 1.5.7 → 1.8.6.** We shipped a build released 2020-10-27; the channel `core/influxdb` already configures serves 1.8.6
- **Taken from the repo instead of a direct download URL**, because the rpms under `dl.influxdata.com/kapacitor/releases/` are **unsigned** and the same build in the repo is not
- **Pinned in both directions** — the exact NVR in `ROOTFS_DNF`, and `LOCKED_DNF` so `installdnf`'s duplicate pass cannot resolve it elsewhere
- **Fixes the repo's `gpgkey`**, which names a retired key that verifies nothing the repo currently serves — including the InfluxDB we already ship

Fixes #648 — Kapacitor upgraded and the alert mechanism verified

- Refs #767 — which will retire Kapacitor; this keeps the alerting path current until then

**Stacked on [#1467](https://github.com/bigstack-oss/cubecos/pull/1467) / `jim.lin/feat/wrapup-openstack-upgrade`.** That PR is not merged, so the diff against `develop` shows its commits too. Only the two from `fa84eb51` onward belong here.

**The gpgkey fix is not a kapacitor problem, which is why it is its own commit and goes first.** `influxdb.repo` has `gpgcheck=1` and names `https://repos.influxdata.com/influxdb.key` — the retired "InfluxDB Packaging Service" key (`2582E0C5`). InfluxData rotated to `influxdata-archive.key` (`9FC2F927`, signing subkey `0585BD3B`), and **both** packages we take from this repo are signed by that subkey — the `influxdb 1.12.4` already installed on jim-1cc reads `Key ID da61c26a0585bd3b`, and so does kapacitor 1.8.6. So `gpgcheck=1` has been reading as protection it was not providing, for InfluxDB as much as for the package I came here to bump. Same shape as the OpenSearch key rotation.

**Why the sourcing moved.** The 1.5.7 line was an explicit `dl.influxdata.com` URL fed to `ROOTFS_DNF_DL_FROM`, and `rpm -qpi` on that file reports `Signature : (none)` — an unsigned 87 MB binary installed as a local file, which bypasses repo `gpgcheck` entirely. The repo copy of the identical version is signed. Moving to the repo is what makes the version *verified* rather than merely *pinned*.

**Why both pins.** They guard different things and neither is sufficient. Without the NVR in `ROOTFS_DNF` the build silently follows the channel wherever it goes. Without `LOCKED_DNF`, a channel that moves past our NVR hands `installdnf` a duplicate pair and it keeps the newer — the drift #1150 pinned qemu against. No epoch in the lock, per the qemu note in `core/heavyfs/Makefile`.

**What the jump contains.** 1.6 through 1.8 are almost entirely Go toolchain and dependency bumps — 1.5.7 was built against Go 1.13, 1.8.6 against Go 1.25.10 — which is the security value and usually what a CVE scan is actually reporting. The single breaking change is 1.8.0 dropping Python 2 UDF support, and we ship no UDFs: nothing in `core/kapacitor` references one and the node's `kapacitor.conf` has no `[udf]` section. 1.7.0 rewrites the topic store, so an upgrade migrates it in place — the one thing worth verifying rather than reasoning about, and it is transparent.

**One thing that looked like a regression and was not.** The first VM alert probe on jim-1cc did not fire. That node runs the pre-#672 build, whose VM templates are still bound to `monasca.def`, and the probe was written to `telegraf.def`. `alert_host_cpu`, which needs no template redefinition, processed points normally on 1.8.6. The full alert path is verified on accept-3cc, where the templates have been re-sourced.

**Not verified: a rolling upgrade from 1.5.7 (QA step 10).** Both labs were upgraded node-by-node in place, which exercises the store migration and the restart but not the roll's mixed-version window. Agreed as out of scope for this issue.

**Also left alone deliberately: InfluxDB is unpinned** (`ROOTFS_DNF += influxdb`, no `LOCKED_DNF`, currently landing 1.12.4). The rescope comment calls pinning it worth doing regardless; it is out of scope here.

#### Additional documentation

Verified on `jim-1cc` (single control) and `accept-3cc` (3-node HA), upgrading in place from 1.5.7 over the existing task store.

For the requirement "the repo key verifies what the repo serves":

```
influxdb.key                  pub 2582E0C5  sub 87F70D56   InfluxDB Packaging Service   (retired)
influxdata-archive.key        pub 9FC2F927  sub 7DF8B07E
                                            sub 0585BD3B   InfluxData Package Signing Key

installed influxdb 1.12.4     Signature : ... Key ID da61c26a0585bd3b
repo kapacitor 1.8.6          Signature : ... Key ID da61c26a0585bd3b
```

Before the fix, on jim-1cc:

```
Public key for kapacitor-1.8.6-1.x86_64.rpm is not installed.
GPG Keys are configured as: https://repos.influxdata.com/influxdb.key
Import of key(s) didn't help, wrong key(s)?
Error: GPG check FAILED
```

After importing the archive key, `rpm -K` on both rpms: `digests signatures OK`. And with the corrected file in place on **accept-3cc**, which had no InfluxData key in its rpmdb at all, dnf imported `0x9FC2F927` unaided and completed the install — which is the end-to-end proof of the build change.

For QA 1–4, per node:

```
10.1.0.1   kapacitor-1.8.6-1.x86_64  active  tasks=22 disabled=0 relays=6
10.1.0.2   kapacitor-1.8.6-1.x86_64  active  tasks=22 disabled=0 relays=6
10.1.0.3   kapacitor-1.8.6-1.x86_64  active  tasks=22 disabled=0 relays=6

hex_sdk health_kapacitor_check   rc=0   (both clusters)
```

The task set is byte-identical before and after on every node, 0 disabled and 0 carrying an `Error:` — so the 1.7.0 topic-store migration is transparent. `/etc/kapacitor/kapacitor.conf` is preserved unchanged, with the rpm's version left as `.rpmnew`. The unit file *does* change, so a `systemctl daemon-reload` is needed before the restart.

For QA 5 and 6 — the write proxy, and the property that matters most, since InfluxDB 1.x OSS has no clustering and these relay tasks are the only reason three nodes agree:

```
POST http://10.1.0.100:9092/write?db=telegraf&rp=def   -> 204

10.1.0.1   qa_probe_1788978428,1788978428016649777,qa,1
10.1.0.2   qa_probe_1788978428,1788978428016649777,qa,1
10.1.0.3   qa_probe_1788978428,1788978428016649777,qa,1
```

`def` and `hc` are separate relay tasks, so `hc` was checked separately rather than assumed from `def`:

```
POST http://10.1.0.100:9092/write?db=telegraf&rp=hc    -> 204

10.1.0.1   qa_hc_1788978428,1788978436252301920,qa,1
10.1.0.2   qa_hc_1788978428,1788978436252301920,qa,1
10.1.0.3   qa_hc_1788978428,1788978436252301920,qa,1
```

For QA 7 and 8 — the alert mechanism, driven through the VIP on accept-3cc:

```
alert6 [alerts_triggered="1" crits_triggered="1" errors="0" ...]
alert_events  stream0 -> from1 [processed="1"]

events.instance:
  CPU00006C  project "admin" instance "Cirros_CubeStorage-e0e1"
             cpu usage 97% exceeds threshold 95%   severity C
```

The re-entry loop is the part that could have broken quietly: the alert's `.message()` is line protocol POSTed back to Kapacitor's own write endpoint, so `alert_events` processing it is what proves both halves still work on 1.8.6.


SekiXu (changes requested):
Reviewed the two commits from `fa84eb51` that belong to this PR. **I found nothing wrong with them.** The block is that this branch contains #1457 and #1467, and I have asked for changes on #1457 — it leaves `check_service monasca` at `core/main/proj_functions:969` calling a `health_monasca_check` it deletes, which corrupts the `system.services` JSON for every service on every control node, and `cube-cos-api` still reads `bucket: "monasca"` in 12 places. Details there.

On the two commits themselves:

- **The gpgkey fix being its own commit, and first, is right.** `gpgcheck=1` against a key that signs nothing the repo serves is worse than no check, because it reads as protection in a review. The InfluxDB we already ship being signed by the same subkey is the part that makes this not a kapacitor-only bug.
- **The double pin is correct and correctly spelled.** `ROOTFS_DNF += kapacitor-1.8.6-1` plus `LOCKED_DNF += kapacitor-1.8.6-1` is the qemu/openssl/ovn pairing, and the no-epoch form matches the note in `core/heavyfs/Makefile` — `LOCKED_DNF` is matched as a string against rpm filenames, which never carry one.
- **Moving off `ROOTFS_DNF_DL_FROM` is what actually buys the verification**, not the version bump. An unsigned local rpm bypasses repo `gpgcheck` entirely, so the old line was pinned but unverified.

One dependency worth stating out loud since it is new for this component: kapacitor now needs `core/influxdb/influxdb.repo` to be in the rootfs at dnf time. That happens through the `$(shell cp ...)` in `INFLUXDB_REPO`, which is expanded from `core/heavyfs/Makefile:72`. It already works for `ROOTFS_DNF += influxdb`, so this is not a new risk — but `core/kapacitor` now silently depends on a side effect that lives in `core/influxdb`, and if that `echo` line ever goes, two components break rather than one.

Happy to approve once the stack below it is sorted.

SekiXu (approved):
Approving. The stack below this is sorted — #1457's fixed head and #1467 are both ancestors — and the two commits that belong to this PR are unchanged from what I reviewed on `fa84eb51`, verified by diffing this head against `jim.lin/feat/wrapup-openstack-upgrade`.

Restating what held, so this page stands on its own:

- The gpgkey fix being its own commit, and first, is right. `gpgcheck=1` against a key that signs nothing the repo serves is worse than no check, because it reads as protection in review.
- `ROOTFS_DNF += kapacitor-1.8.6-1` plus `LOCKED_DNF += kapacitor-1.8.6-1` is the qemu/openssl/ovn pairing, and the no-epoch form is correct — `LOCKED_DNF` is matched as a string against rpm filenames, which never carry one.
- Moving off `ROOTFS_DNF_DL_FROM` is what actually buys the verification, not the version bump. An unsigned local rpm bypasses repo `gpgcheck` entirely, so the old line was pinned but unverified.

The dependency I noted stands as a note, not a change: `core/kapacitor` now needs `core/influxdb/influxdb.repo` in the rootfs at dnf time, which arrives through the `$(shell cp ...)` in `INFLUXDB_REPO`. It already works for `ROOTFS_DNF += influxdb`, so no new risk — but two components break rather than one if that line ever goes.

Issue #1490, opened 2026-09-14
[Bug] Cross-cluster volume migrate silently moves no data: every install shares one Ceph fsid


traviswu-bigstack:
## Description

> A clear and concise description of what the bug is.

Every CubeCOS cluster ships with the **same Ceph fsid**. `core/modules/config_ceph.cpp:77`
hardcodes it as a compile-time constant:

```cpp
static const char FSID[] = "c6e64c49-09cf-463b-9d1c-b6645b4b3b85";
```

and that constant is the default for the unpublished tunable `ceph.fsid`
(`config_ceph.cpp:128`), so unless an operator overrides a hidden tunable, two unrelated
installs are indistinguishable by fsid. Confirmed on two independent lab clusters — the sky
3-node cluster and the r630 single-node cluster both report
`c6e64c49-09cf-463b-9d1c-b6645b4b3b85`.

On its own that is untidy. It becomes a **data-integrity bug** when a second Ceph cluster is
registered as a Cinder RBD backend. Cinder's RBD driver uses the fsid to decide whether the
source and destination are the same cluster:

- `cinder/volume/drivers/rbd.py:2372` compares the destination's fsid (from `location_info`)
  against the source's. Identical fsids ⇒ it concludes "same cluster" and does **not** refuse.
- `rbd.py:2378` then compares pool names. CubeCOS always names the pool `cinder-volumes`, so
  they match too, and the driver takes the shortcut *"Migration in the same pool, just need to
  update volume's host value to destination host"*, returning `(True, None)`.

The manager accepts that as a completed migration and simply rewrites the volume's `host`.
**No data is copied.** The volume is left pointing at a cluster that has never held its image.

Observed: `cinder migrate` returned success in under 10 seconds; Cinder then reported the
volume on `r630ceph@r630ceph#r630ceph`; the RBD image was absent from the r630 cluster and
still present on sky; reading from where Cinder pointed failed with
`error reading header from volume-<id>: No such file or directory`. An attach after this
point cannot succeed, and a delete would be issued against the wrong cluster.

**Reachability.** Narrow, but real. Every CubeCOS volume type pins `volume_backend_name`, and
the scheduler refuses to place a typed volume on a foreign backend
(`NoValidBackend: Cannot place volume … on r630ceph@r630ceph`). The shortcut therefore fires
only for a volume on the unconstrained `__DEFAULT__` type. It is **not** reachable through
[...]

Issue #1484, opened 2026-09-13
[Task] CI runs none of the 19 core/sdk_sh/tests files


traviswu-bigstack:
## Description

`core/modules/Makefile` carries `SUBDIRS := tests`, which is how the `core/modules/tests/` suites reach CI. `core/sdk_sh` has no Makefile at all, so nothing reaches `core/sdk_sh/tests/` — 19 shell test files covering ceph, cinder, ganesha, gpu, neutron, ovn, power, health and advisor.

They pass when run by hand and are invisible to CI, so a regression in any of them lands unnoticed.

## Expected behavior

`core/sdk_sh/tests/` runs in CI alongside `core/modules/tests/`.

## Additional Context

Wiring this on needs a pass over the 19 files' current status first — switching them all on at once would block whichever PR does it on unrelated test debt across many feature areas. That is why this is filed separately from the change that noticed it.

Noticed while reviewing the Advisor upgrade-survival work, where the fix for the most hazardous line on the branch is pinned by exactly one of these files.

## Output artifacts (Definition of Done)

- [ ] **Handbook knowledge update** — land the durable, team-readable knowledge from this
      work into the bigstack-handbook cubecos kb via `/bigstack-core:save-to-handbook`
      (Topic / Runbook / Known-issue / ADR as fits).

PR #1492, opened 2026-09-14
Move a volume between storage backends while the VM keeps running

traviswu-bigstack:
#### What type of PR is this?

/kind feature

Gives CubeCOS a scriptable, honest way to move a volume between storage backends while the VM keeps running.

A lab spike proved the move itself already works — an attached volume moves live between two physically separate Ceph clusters, guest never stops, data byte-exact — but the product surface around it was thin: `iaas volume move` refused silently, and nothing but the CLI could start a move.

This adds three shell entry points in `sdk_cinder.sh`:

- `cinder_move_preflight <vol> <type>` — one decision table for "can this move, and why not", reporting **every** blocking reason rather than the first, so a tenant who fixes one does not discover the next on retry. It has two layers: Cinder/Nova API state, and a probe of the running libvirt domain on the compute node (the analogue of the live-resize feature's `check_live_resize`). The domain probe fails **open** — a probe that cannot run must never invent a refusal.
- `cinder_move_volume <vol> <type>` — preflight, then `cinder retype --migration-policy on-demand`. Never `cinder migrate`, which can repoint a volume at another Ceph cluster without copying a byte (#1490).
- `cinder_volume_image_name <vol>` — resolves an RBD image name honouring `name_id`. After a cross-backend move the image is `volume-<name_id>`, not `volume-<vol_id>`; anything mapping images back to volumes by name mis-attributes every migrated volume without this.

`iaas volume move` now consults the preflight and prints every refusal reason instead of exiting silently.

Also included: an operator-run acceptance script that exercises the whole path on real hardware.

Fixes #

**Not yet proven on hardware.** The acceptance script (`core/sdk_sh/tests/test_tier_change_e2e.sh`) has never been executed — it creates and destroys real resources and needs two registered Cinder backends. Running it is what proves this feature, and it is also the instrument that settles two open questions: whether `os-vol-mig-status-attr:migration_status` is visible to a given deployment's credentials, and whether `volume_type` flips at retype *acceptance* or at *completion*. Review concluded completion (~85% confidence) from the Cinder source; the run would make it certain. CubeCMP's worker keys its completion detection on that answer.

**The shell tests could not fail until late in this branch.** The `chk` helper printed actual vs expected without comparing, so every "N/N pass" was eyeball verification. It now compares, tracks failures and exits non-zero, following the `ck` convention already used by `test_cinder_storage_tier_registry.sh`. Verified by mutating the module (not the expectations) and confirming a non-zero exit. Note #1484: CI runs none of `core/sdk_sh/tests/`, so these still are not executed automatically.

**Two known limits, deliberate:**
- The fsid/pool guard resolves a volume type to its backend section via the `volume_backend_name` extra spec. A type created by hand without that spec now *refuses* moves rather than silently passing — fail-closed, with the reason stated.
- `hex_cli -c` always exits 0 regardless of `CLI_FAILURE` (pre-existing; `main()` discards `RunCommand`'s result), so automation must parse output, not exit codes.

Paired with cube-cos-api and cubecmp branches of the same name.

#### Additional documentation

```docs
kb/cubecos/architecture/live-volume-move-between-ceph-backends.md
kb/cubecos/known-issues/ceph-fsid-identical-across-installs.md
```


core/cinder/builtin_models/rbd-RBDDriver.yaml
```diff
@@ -0,0 +1,41 @@
+driver: "cinder.volume.drivers.rbd.RBDDriver"
+vendor: "Generic"
```
Eandalf-Bigstack, line 2: We would need to rename this file from `ceph-RBDDriver.yaml` to `rbd-RBDDriver.yaml` for other sdk_cinder.sh function to resolve it correctly.

| Driver Name | Model File Name |
| - | - |
| dell_emc.sc.storagecenter_fc.SCFCDriver | dell_emc-sc-storagecenter_fc-SCFCDriver.yaml |
| fujitsu.eternus_dx.eternus_dx_iscsi.FJDXISCSIDriver | fujitsu-eternus_dx-eternus_dx_iscsi-FJDXISCSIDriver.yaml |

traviswu-bigstack: You're right, and it's a real break rather than a naming preference — fixed in af7a398b.

`cinder_get_model_file_name` derives the name from the driver class, so there is exactly one name a lookup can find:

```sh
output="${output#"cinder.volume.drivers."}"
output="${output//./-}"
```

`cinder.volume.drivers.rbd.RBDDriver` → `rbd-RBDDriver`. With the file named `ceph-RBDDriver.yaml` every caller that resolves a model by driver — `cinder_get_storage`, `cinder_put_storage`, `cinder_delete_model`, the volume-type lookup — would have missed it, so registering an RBD backend through the feature would have failed at the model read. I checked the derivation against the models already on disk, and yours are the two data points that make the rule unambiguous:

| driver | derived | on disk |
| - | - | - |
| `cinder.volume.drivers.rbd.RBDDriver` | `rbd-RBDDriver.yaml` | ✅ now |
| `cinder.volume.drivers.dell_emc.sc.storagecenter_fc.SCFCDriver` | `dell_emc-sc-storagecenter_fc-SCFCDriver.yaml` | ✅ |
| `cinder.volume.drivers.nfs.NfsDriver` | `nfs-NfsDriver.yaml` | ✅ |

Renamed, `cinder.mk` updated, and the commit message now says where the name comes from so the next model does not repeat this. `type: "Ceph"` stays as the display type — that field is not what the lookup keys on.

The PR is in draft: the sky lab was repurposed, so the acceptance script has not been re-run since. It needs a lab pass before it is ready.

Eandalf-Bigstack (approved):

Issue #1322, opened 2026-08-22
[Bug] Load balancers fail intermittently: three per-node octavia config defects


traviswu-bigstack:
## Description

Three separate defects, filed together because they are indistinguishable to an operator: an
`openstack loadbalancer create` or `failover` lands at `provisioning_status ERROR`, and retrying
sometimes works. They look like one flaky subsystem. They are three different bugs, each affecting a
*subset of nodes*, which is why the outcome depends on which node's octavia worker picks up the
taskflow job.

The discriminator is the error text in `/var/log/octavia/worker.log`:

**1. `Multiple possible networks found, use a Network ID to be more specific (HTTP 409)`** — the
commit path authored `amp_boot_network_list` / `amp_secgroup_list` gated on `s_bInit`, which
`InitCheck()` clears whenever `/etc/appliance/state/octavia_init_done` is absent. That marker is only
ever set on the master, so on every non-master **control** node each octavia commit blanked both ids,
and `ComputeCreate` then booted the amphora with **no nics**. A node whose last octavia touch was a
commit keeps them empty indefinitely.

**2. `CertificateGenerationException: Failed to load CA Cert`** — the peer CA copy was
`scp -r root@<master>:/etc/octavia/certs/ /etc/octavia/certs/`. Because the destination already
exists, that lands the tree at `/etc/octavia/certs/certs/`, where `ca_01.pem` never appears at the
path `octavia.conf` points to. Amphora creates on those peers fail.

**3. LB `operating_status OFFLINE` while forwarding traffic fine** — `controller_ip_port_list` was
built from `s_lbHa` (the *amphora topology* tuning) rather than cluster HA, so with the default
`octavia.ha=false` each node wrote only its own health-manager endpoint. Rebooting that node
black-holes the amphora's heartbeats and octavia marks the LB ERROR.

## How to Reproduce

1. On a 3-node HA cluster, compare across nodes:
   `grep -E '^amp_boot_network_list|^amp_secgroup_list|^controller_ip_port_list' /etc/octavia/octavia.conf`
   — the ids are blank on non-master control nodes, and the endpoint list holds one entry.
2. `ls /etc/octavia/certs/certs/` on a peer → the CA tree is one directory too deep.
3. Create and fail over load balancers repeatedly. A share of attempts fail, with the error text
   depending on which node's worker ran the job.

## Expected behavior

Every node's octavia config carries the same cluster-wide ids, the CA lands where `octavia.conf`
[...]

PR #1498, opened 2026-09-15
fix(octavia): carry lb-mgmt state across a roll and target the real health-manager nodes

### LINKED ISSUE #1496: [Bug] Rolling update blanks the octavia lb-mgmt ids on every rolled node
## Description

A firmware rolling update blanks `amp_boot_network_list` and `amp_secgroup_list` in
`/etc/octavia/octavia.conf` on **every node it rolls**, regardless of role. This is distinct from
#1322 defect 1, whose fix (PR #1302) does hold for a plain reboot.

`config_octavia.cpp` registers only three paths for migration across the slot swap:

```
CONFIG_MIGRATE(octavia, "/etc/octavia/certs");
CONFIG_MIGRATE(octavia, KEYFILE);        // octavia_ssh_key
CONFIG_MIGRATE(octavia, KEYFILE_PUB);
```

`octavia.conf` is not among them. A roll installs a fresh rootfs into the inactive slot, and the
shipped `octavia.conf` does not contain the two keys at all. `Init()` loads that file into `oldCfg`,
so the carry-forward added by #1302 —

```cpp
cfg["controller_worker"]["amp_boot_network_list"] = oldCfg["controller_worker"]["amp_boot_network_list"];
```

— copies an empty string. The fix behaves exactly as written; on a rolled slot there is nothing to
carry forward. Its own comment names the case ("on a genuine first init oldCfg has nothing") — a
rolled slot is indistinguishable from a first init.

**This predates #1302 — it is not a regression from that fix.** The pre-#1302 commit path read:

```c
if (s_bInit) { cfg[...] = oldCfg[...]; }   // carry forward
else         { cfg[...] = ""; }            // blank
```

`Configs` is `std::map<std::string, std::map<std::string, std::string>>`, so on a rolled slot
`oldCfg[...]` default-constructs an empty string and **both branches produce empty**. `git log -S`
puts `amp_boot_network_list` and the `CONFIG_MIGRATE(octavia, ...)` set at the initial cube commit:
`octavia.conf` has never been migrated.

What #1302 changed is observability. Before it, non-master control nodes were blanked by *every*
commit and sat empty permanently, so there was no known-good baseline and a roll could not be shown
to have broken anything. Once the everyday blanking stopped, a pre-roll `reinit_octavia` could
establish one — which is what made this attributable on the fourth QA round. The same masking
applies to computes: pre-fix they took the `s_bInit == true` branch, which is where "computes are
never wiped" comes from. That holds for the commit/reboot path and not for the roll path.

The blanking is deterministic. What varies is whether a node is re-stamped afterwards, and that
depends on **roll order**, which is why it presents as intermittent:

- the master re-stamps itself via the master-only `cluster_start` → `reinit_octavia`;
- `os_octavia_init_peers` sweeps only first-three-compute nodes, and only at the moment the master
  runs it — a compute that rolls later is missed;
- control-only nodes are never re-stamped and never checked (ERR_CODE 10 sits in the
  `is_first_three_compute_node` loop). Harmless: they do not boot amphorae;
- a first-three-compute node is repaired by the end-of-roll `cluster_check_repair_async` — **unless
  `cube_repair_optout` is set**, in which case it stays blank indefinitely.

**Impact:** a compute left with empty ids fails every amphora build handed to its worker with nova's
misleading `Multiple possible networks found, use a Network ID to be more specific (HTTP 409)`.

## How to Reproduce

1. On a healthy cluster, record the ids on every node:
   `grep -E '^(amp_boot_network_list|amp_secgroup_list)' /etc/octavia/octavia.conf`
2. `hex_cli -c cluster -c rolling_update` and let it finish.
3. Repeat #1. Every rolled node except the master is blank; a node the on-target check skipped
   (never rebooted, never swapped slot) keeps its values.
4. The pre-roll file is still on disk — mount the inactive root slot read-only and grep it; the
   UUIDs are there.

Observed on `CUBE_3.1.10_20260903-1323_66cd48a`, 3 control + 2 compute, roll 2026-09-14 23:10 →
2026-09-15 01:29:

| Node | boot (CST) | boot commit blanks | re-stamped after | result |
| :--- | :--- | :--- | :--- | :--- |
| c1 master | 23:24 | yes, 23:30 | yes — `reinit_octavia` 23:39 | set |
| c2 control | 23:47 | yes, 23:55 | no | blank |
| c3 control | 00:23 | yes, 00:32 | no | blank |
| p5 compute | 01:14 | yes | no — peer sweep ran 23:39, before p5 rolled | blank |
| p4 compute | not rolled | — | — | set |

## Expected behavior

A rolling update leaves the two cluster-wide ids intact on every node it rolls, or re-derives them
when the node rejoins. Recovery must not depend on roll order, on which node is master, or on
health repair being enabled.

## Additional Context

Evidence that the mechanism is the missing migrate, not an empty `ReconfigMain` lookup:

- On c2 and c3, across the whole log spanning the roll,
  `grep -cE "reconfig_octavia|reinit_octavia|os_octavia_init" /var/log/hex_config.log` → **0**, and
  zero `lb-mgmt-net lookup empty` warnings. `ReconfigMain` never ran.
- c2 and c3 inactive slots hold `octavia.conf` dated 2026-09-14 23:03 with both UUIDs; the live
  slots are empty.
- p4 never rolled, so its unused slot holds the shipped file, dated 2026-09-03 19:48 (image build
  time) — the two keys are **absent**, not blank.
- #1322 QA-4 is the control: the certs survived the same roll, and they are the one octavia path in
  the migrate list.

Note when reading `hex_config.log`: it jumps 8h mid-boot (hex_config starts in UTC, then
`Set timezone to Asia/Taipei`), so time-filtered greps hide the master's octavia lines.

**Suggested fix — migrate the ids; health repair stays the net, it cannot be the fix.**

Repair is structurally the wrong layer here, for reasons the codebase already states:

- `sdk_health.sh:136` gates auto_repair on `! is_cluster_rolling` — repair is deliberately off for
  the entire duration of a roll (*"nodes are deliberately down and the roll owns service state;
  repairing fights it"*), which is exactly the window in which this defect is created;
- both the periodic auto_repair and the end-of-roll `cluster check_repair`
  (`core/main/proj_functions:1296`) are gated on `cube_repair_optout`, whose stated purpose is
  *"it suppresses health repair so faults stay visible"* — so an exemption for code 10 would defeat
  the marker;
- even fully enabled it is budgeted (`_health_repair_due "$count" "$maxerr" "$backoff"`), so the
  node's `octavia-worker` serves taskflow jobs with empty ids until it fires.

PR #1302 already fixed this exact class the other way, in `config_nova.cpp`:

```cpp
// It lives in the root slot, so without this an upgrade -- which boots the other
// slot -- leaves every upgraded node without nova-compute, and the roll stalls.
CONFIG_MIGRATE(nova, "/var/lib/nova/compute_id");
```

Same mechanism, same trigger. Mirroring it:

- persist the two ids in a small state file and `CONFIG_MIGRATE` that **one file**;
- `UpdateCfg` falls back to that file when `oldCfg` has nothing;
- **not** `CONFIG_MIGRATE(octavia, CONF)` — migrating a generated config would freeze stale keys
  across releases, so `octavia.conf` keeps being regenerated from the `.def`;
- no neutron lookup on the boot path, so it avoids the empty-lookup case #1302 already had to guard
  against.

Health repair keeps its present role: catching drift on first-three-compute nodes once the cluster
is settled.

Also needs: correcting `kb/cubecos/known-issues/octavia-lb-mgmt-ids-wiped-on-non-master.md`, whose
role table is about the pre-fix `InitCheck()` gate and does not describe this path; and correcting
#1322's QA-1/2/3 wording from "every control holds UUIDs" to "every node that runs `octavia-worker`".

Found by QA on #1322: https://github.com/bigstack-oss/cubecos/issues/1322#issuecomment-5674889982

## QA Verification

|---|---|---|
| 1 | Before the roll, on every node: `grep -E '^(amp_boot_network_list\|amp_secgroup_list)' /etc/octavia/octavia.conf` | both hold a 36-char UUID |
| 2 | `hex_cli -c cluster -c rolling_update`, wait for every node done, then repeat #1 | the same UUIDs on every node — master and non-master, control and compute |
| 3 | Repeat #2 with `cube_repair_optout` set for the whole roll | same result; recovery must not depend on health repair running |
| 4 | Repeat #2 with the roll ordered so a compute rolls **after** the master | same result; no dependence on roll order |
| 5 | On every node running `octavia-worker`: `hex_sdk os_octavia_cfg_ids_ok; echo $?` | `0` |
| 6 | After the roll, create 5 load balancers; then on every node `grep -c 'Multiple possible networks found' /var/log/octavia/worker.log` | all 5 reach `provisioning_status ACTIVE`; count is `0` on every node |

- [x] **Handbook knowledge update** — `kb/cubecos/known-issues/octavia-lb-mgmt-ids-lost-on-firmware-roll.md`
      + `.scenario.yaml`, landed by bigstack-oss/bigstack-handbook#665. Filed as its own note rather
      than an extension of `octavia-lb-mgmt-ids-wiped-on-non-master.md`, which describes the
      pre-fix `InitCheck()` gate and cannot explain the compute case; that note gains a pointer
      saying so. The same PR **reopens** `upgrade-drops-root-slot-state.md` (was
      `fixed-in:3.1.10`) — this is a third instance of the class it documents, so #1302 fixed the
      two instances it found, not the class.

### LINKED ISSUE #1497: [Bug] octavia controller_ip_port_list names control nodes, but health managers run on computes
## Description

`controller_ip_port_list` is built from the **control** node addresses, while
`octavia-health-manager` only ever runs on **compute** nodes. On any cluster where the two roles are
not the same machines, every amphora is told to send its heartbeats to addresses that exist on no
node, and automatic failover can never trigger.

In `config_octavia.cpp`:

- `Commit` calls `ControllerIpPortList(s_ha, s_mgmtCidr, cidrIp, s_ctrlAddrs)` — the list is mapped
  over `cubesys.control.addrs`;
- `CommitService` starts the health manager under `if (IsCompute(s_eCubeRole))`, alongside
  `octavia-worker`. Control gets `octavia-api` and `octavia-housekeeping` only.

#1322 defect 3 fixed the *cardinality* of this list (build it from cluster HA rather than the
amphora topology `s_lbHa`, so it stops naming a single endpoint). It did not change the *source
set*, so the list is still the wrong nodes. This is invisible on converged 1cc/3cc, where control
and compute are the same machines and the addresses coincide.

**Impact:** `amphora_health` never receives a heartbeat, so octavia can never declare an amphora
timed out and never fails one over. The data path is unaffected — load balancers keep forwarding —
so this presents only as `operating_status OFFLINE` plus failovers that silently never happen.

## How to Reproduce

On a cluster with separate control and compute nodes (e.g. 3 control + 2 compute):

1. On every node: `grep ^controller_ip_port_list /etc/octavia/octavia.conf`
2. Check which of those addresses exist: `ip -4 -o addr show octavia-hm0` on every node.
3. Check who is listening: `ss -lunp | grep 5555` on every node.
4. `mysql -u root -D octavia -e "select count(*) from amphora_health"` with load balancers up.
5. Power off the host of an active amphora and watch `/var/log/octavia/worker.log`.

Observed on `CUBE_3.1.10_20260903-1323_66cd48a`, 3 control (c1/c2/c3) + 2 compute (p4/p5):

```
controller_ip_port_list = 10.254.0.11:5555,10.254.0.12:5555,10.254.0.13:5555   (all 5 nodes)
octavia-hm0             exists only on p4 (10.254.0.14) and p5 (10.254.0.15)
UDP :5555 listeners     c1/c2/c3 = 0,  p4/p5 = 1
octavia-health-manager  c1/c2/c3 inactive,  p4/p5 active
amphora_health          0 rows
failover on host loss   0 attempts in worker.log
```

## Expected behavior

`controller_ip_port_list` names exactly the lb-mgmt addresses of the nodes actually running
`octavia-health-manager`, so every listed address has a listener and `amphora_health` is populated.

## Additional Context

- Topology: 3 control + 2 compute HA, `cubesys.ha = true`, `octavia.ha = false`.
- `health_manager.bind_ip` is correct on the compute nodes (p4 `10.254.0.14`, p5 `10.254.0.15`) —
  only the list is wrong. `bind_ip` on control nodes is unused and harmless.
- Nothing fails loudly because `net.ipv4.ip_nonlocal_bind` is 1, so binding a nonexistent address
  succeeds.

**Suggested fix.** Build the list from the nodes that run the health manager — the first-three-compute
set that `CommitService` and `os_octavia_init_peers` already use — instead of `s_ctrlAddrs`.

**Upgrade note required:** `controller_ip_port_list` is delivered to an amphora at build time, so
existing load balancers need `openstack loadbalancer failover` to pick up the corrected list.

Supersedes the remainder of #1322 defect 3. QA's retraction of the #1322 QA-5 PASS:
https://github.com/bigstack-oss/cubecos/issues/1322#issuecomment-5674889982

## QA Verification

|---|---|---|
| 1 | On a split control/compute cluster, on every node: `grep ^controller_ip_port_list /etc/octavia/octavia.conf` | the listed addresses are exactly the `octavia-hm0` addresses of the nodes running `octavia-health-manager`, identical on all nodes |
| 2 | For each address in that list, on its owning node: `ss -lunp \| grep 5555` | one listener per listed address; no listed address belongs to a node where `octavia-health-manager` is inactive |
| 3 | Create a load balancer, wait 2 min, then `mysql -u root -D octavia -e "select count(*) from amphora_health"` | non-zero, and `last_update` advances between two reads 60 s apart |
| 4 | Same load balancer reaches steady state: `openstack loadbalancer show <id>` | `provisioning_status ACTIVE` and `operating_status ONLINE` |
| 5 | Hard power off the compute hosting an amphora; watch `/var/log/octavia/worker.log` | a failover is triggered within the heartbeat timeout and the LB returns to `ACTIVE`/`ONLINE` |
| 6 | For a load balancer that existed before the upgrade: `openstack loadbalancer failover <id>`, then repeat #3 | heartbeats arrive for the rebuilt amphora |
| 7 | Repeat #1 on a converged 1cc | unchanged from today's behaviour (no regression where control == compute) |

- [x] **Handbook knowledge update** — `kb/cubecos/known-issues/octavia-health-manager-endpoint-list-wrong-nodes.md`
      + `.scenario.yaml`, landed by bigstack-oss/bigstack-handbook#665, which also corrects
      `octavia-health-manager-bind-ip-hardcoded.md` — its "list half fixed in #1249" claim is
      disproved for split control/compute topologies.

traviswu-bigstack:
#### What type of PR is this?

/kind bug

Two octavia per-node config defects found by QA on the 3c2p testbed while verifying #1322
against `CUBE_3.1.10_20260903-1323_66cd48a`. They are folded together because both are the same
shape — a cluster-discovered value that the commit path must never author — and they touch the
same code.

**1. A firmware roll blanks both lb-mgmt ids (#1496).** `octavia.conf` is not in the module's
migrate set, and the shipped file does not contain the two keys at all. A roll boots a freshly
installed slot, so `Init()` loads that file into `oldCfg` and the carry-forward added by #1302
copies an empty string. It behaves exactly as written; on a rolled slot there is nothing to carry
forward. This is role-independent, which is why a compute node was hit — something the original
`InitCheck()` defect could not explain.

**Not a regression from #1302.** The pre-#1302 gate was `if (s_bInit) carry-forward; else blank`,
and since `Configs` is a nested `std::map`, `oldCfg[...]` default-constructs `""` on a rolled slot —
so both branches produced empty there too. `git log -S` puts `amp_boot_network_list` and the
`CONFIG_MIGRATE(octavia, ...)` set at the initial cube commit. What #1302 changed is observability:
before it, non-master control nodes were blanked by every commit and had no known-good baseline to
compare a roll against.

Fixed the way #1302 already fixed the same class for nova (`CONFIG_MIGRATE(nova,
"/var/lib/nova/compute_id")`): keep the discovered values in `/etc/cube/cos/octavia/lb_mgmt`,
migrate that one file, and fall back to it when `oldCfg` has nothing. Not
`CONFIG_MIGRATE(octavia, CONF)` — migrating a generated config would freeze stale keys across
releases, so `octavia.conf` keeps being regenerated from the `.def`. A migrate *function* also
lifts the values off the previous slot's `octavia.conf`, so the upgrade that delivers this fix is
covered rather than being the one roll it cannot help.

**2. `controller_ip_port_list` named the wrong nodes (#1497).** The list was built from
`cubesys.control.addrs`, but `CommitService` starts `octavia-health-manager` under
`if (IsCompute(...))`. On the testbed all five nodes carried
`10.254.0.11/.12/.13:5555` while the only health managers were on p4 (`.14`) and p5 (`.15`) —
`amphora_health` had 0 rows and no failover could ever trigger. #1302's defect-3 fix corrected the
cardinality of this list but not its source set. Invisible on converged 1cc/3cc, where control and
compute are the same machines.

Derived from the nodes that actually run a health manager instead, via `cubectl`'s node list
(backed by `/etc/settings.cluster.json`, which `config_cluster` already migrates, so it resolves on
a rolled slot), keeping the current value when it cannot be read.

Fixes #1496
Fixes #1497

> Health repair is deliberately *not* the fix for #1496. `sdk_health.sh:136` gates auto_repair on
> `! is_cluster_rolling` — repair is off for the whole window in which the defect is created — and
> both it and the end-of-roll `cluster check_repair` (`proj_functions:1296`) are gated on
> `cube_repair_optout`, whose purpose is to keep faults visible. The existing `ERR_CODE=10` check
> stays exactly as it is, as a net.

> `UpdateCfg`'s first parameter is `s_lbHa` (the amphora topology), not cluster HA. The endpoint
> list must not be keyed off it — that was #1322 defect 3 — so `ControllerIpPortList` no longer
> takes an HA flag at all.

> **Upgrade note:** `controller_ip_port_list` is delivered to an amphora at build time, so existing
> load balancers need `openstack loadbalancer failover` to pick up the corrected list.

> Verified: compiles clean under the project's `-Wall -Werror`; `os_octavia_hm_nodes` checked
> against the live 3c2p cluster, where it returns `10.32.36.4,10.32.36.5` → `10.254.0.14:5555,
> 10.254.0.15:5555`, matching the actual `octavia-hm0` addresses. Not yet exercised through a real
> roll — that is the QA verification in both issues.

#### Additional documentation

```docs
kb/cubecos/known-issues/octavia-lb-mgmt-ids-wiped-on-non-master.md
kb/cubecos/known-issues/octavia-health-manager-endpoint-list-wrong-nodes.md
```


Eandalf-Bigstack:
@traviswu-bigstack Is this also going into branch release_3.1.10?

Eandalf-Bigstack (approved):

Issue #1167, opened 2026-07-24
[Spike] AMQP TLS enablement — root fix for RabbitMQ cleartext auth (follow-up to #884)


SekiXu:
## Problem Description

Follow-up to spike #884 / cubecos-private#73 (AMQP Cleartext Authentication, Nessus 87733).

PR #1051 shipped the **compensating control** — SERVICE-INT firewall rules restrict RabbitMQ 5672/15672/25672 to cluster nodes so external scanners can't reach the cleartext port. It is explicitly **not** a compliance fix: 5672 traffic is still cleartext, so anything already inside the cluster (or a compromised node) can still sniff credentials, and it does not satisfy encryption-in-transit (PCI 4.2.1 / NIST SC-8 / ISO A.10.1).

Per Jim's note on cubecos-private#73 (*"TLS consideration should be spiked by dev to see the impact afterward"*), this spike scopes the **root fix**: enable RabbitMQ TLS (5671 / amqps), move all AMQP clients onto it, then close 5672.

> ⚠️ The vulnerability is only truly fixed once **5672 is closed**. A dual-listen transition still leaves 5672 open, so Nessus will keep reporting until cutover completes.

## Scope — three change surfaces

Cluster runs RabbitMQ **3.11.28** (corrected 2026-09-04 — #1171 and #1337 have both merged since this was written). The change spans three parts.

### 1. RabbitMQ server (control nodes)
Generated by `core/modules/config_rabbitmq.cpp` `WriteConfig()`. Today it only emits a 5672 cleartext listener (`NODE_PORT=5672`). Add a 5671 ssl listener + ssl_options to the generated `/etc/rabbitmq/rabbitmq.conf`:

```ini
listeners.ssl.default            = 5671
ssl_options.cacertfile           = /var/www/certs/ca.pem
ssl_options.certfile             = /var/www/certs/server.cert
ssl_options.keyfile              = /var/www/certs/server.key
ssl_options.verify               = verify_none          # encrypt + client-verifies-server; no mutual TLS yet
ssl_options.fail_if_no_peer_cert = false
ssl_options.versions.1           = tlsv1.3
ssl_options.versions.2           = tlsv1.2
```

Closing cleartext = also emit `listeners.tcp = none`. Gate the whole thing behind a tuning (e.g. `rabbitmq.ssl.enabled`), default conservative (dual-listen first). Don't hand-edit rabbitmq.conf — `commit` regenerates it from the C++.

### 2. Certificates
[...]

PR #1499, opened 2026-09-16
refactor: collapse the fourteen AMQP client config writers into one shared helper

### LINKED ISSUE #1426: [Task] Collapse the twelve copy-pasted AMQP client config writers into one shared helper
## Description

> Overview of what should be done.

Prerequisite refactor for the AMQP TLS work scoped in #1167. Twelve `config_*.cpp` modules each carry
their own copy of the code that writes a service's `transport_url` and its `[oslo_messaging_rabbit]`
keys. Adding `ssl = true` / `ssl_ca_file` for TLS therefore means twelve edits, and **missing one is
silent until 5672 closes** — that service simply stops talking to the broker. This issue collapses the
twelve copies into one shared helper first, changing no generated output, so that the TLS change later
lands in one place with the HA-gate decision made once.

### What the twelve copies actually look like

Two function names, not one:

- **`UpdateMqConn()` — 9 modules:** `config_nova.cpp`, `config_neutron.cpp`, `config_heat.cpp`,
  `config_ironic.cpp`, `config_masakari.cpp`, `config_designate.cpp`, `config_watcher.cpp`,
  `config_octavia.cpp`, `config_cyborg.cpp`
- **`SetWorkerQueue()` — 3 modules:** `config_glance.cpp`, `config_cinder.cpp`, `config_manila.cpp`

`config_barbican.cpp` also has an `UpdateMqConn()` but is **not** an AMQP client — it writes
`kafka://…:9095`, and the `RabbitMqServers` occurrence at `config_barbican.cpp:228` is comment text,
not a call. keystone and placement don't use rabbit either. So the client set is **12**.

The bodies are genuinely identical — normalising the config-object and HA variable names, all twelve
write the same six keys: `DEFAULT.transport_url`, `DEFAULT.rpc_response_timeout = 1200`, and four
`oslo_messaging_rabbit` HA keys. The `SetWorkerQueue()` trio already takes the config by reference
(`SetWorkerQueue(Configs& config, const bool isHa, …)`), which is exactly the shared-helper signature.

### The HA gate is the design point, not the de-duplication

All twelve write their `[oslo_messaging_rabbit]` keys **inside `if (ha)` / `if (isHa)`** — verified on
`develop` @ `03054e90`, e.g. `config_nova.cpp:395`, `config_neutron.cpp:396`, `config_glance.cpp`'s
`SetWorkerQueue`. Confirmed on a 1cc dev node (`cubesys.ha = false`): `/etc/{nova,neutron,cinder}/*.conf`
contain the `[oslo_messaging_rabbit]` section header but **no keys at all** — the header comes from each
module's managed-section list (e.g. `config_glance.cpp:324`), unrelated to the HA branch.

That matters for the TLS follow-up: `RabbitMqServers()`
(`core/cube_sdk_library/src/cluster.cpp:180-197`) is the single source of truth for scheme and port and
**its non-HA branch also has to flip** to `rabbits://…:5671`. Putting `ssl = true` / `ssl_ca_file` inside
the existing `if (ha)` block would leave a non-HA node with an amqps URL and no ssl parameters — and
py-amqp with no `ca_certs` still defaults to `PROTOCOL_TLS_CLIENT` (`verify_mode=CERT_REQUIRED`,
`check_hostname=True`) against a self-signed cert, so that fails too. **The helper must therefore keep
the HA-conditional keys and the (future) unconditional ssl keys as separate concerns** — that shape is
what this refactor is buying.

## Requirements

> A list of technical requirements must be met for completion.

- [ ] One shared helper writes `DEFAULT.transport_url`, `DEFAULT.rpc_response_timeout` and the
      HA-gated `[oslo_messaging_rabbit]` keys, taking the config object by reference. All twelve
      modules call it; the twelve local copies are deleted.
- [ ] The helper's structure separates *unconditional* keys from *HA-gated* keys, so the TLS
      follow-up can add ssl parameters outside the HA condition without restructuring again.
- [ ] `config_barbican.cpp` is left alone, and a comment records why (kafka, not rabbit).
- [ ] **Acceptance is byte-identical output.** For each of the twelve services, the generated config
      file before and after the refactor is identical (`diff` clean) — captured on a node for both
      `cubesys.ha = false` and `cubesys.ha = true`, since the HA branch is where all the
      `[oslo_messaging_rabbit]` keys live and a non-HA-only check would prove nothing about them.
- [ ] No behaviour change ships with this issue: no port change, no scheme change, no ssl parameters.
      Those are #1167's implementation issue.

## Notes (Optional)

Split out of #1167 per its own next-step note. Evidence above was read on `develop` @ `03054e90` and on
a 1cc dev node running a 2026-09-02 build.

Sequencing: this issue is behaviour-preserving and independent — it does **not** need #1421 (the cert
SAN widening) and can land first.

- [ ] **Handbook knowledge update** — land the durable, team-readable knowledge from this
      The 12-module inventory and the HA-gate finding belong with the AMQP TLS material.

---

## QA Verification

> Rows below are checkable claims: the exact command plus the expected outcome. Run them
> **from a control node as root**, on a cluster built from a build that contains this change.
> Delivered by **PR #1499** on `refactor/1426-shared-amqp-config-helper` — five commits,
> `0c34ea7e` · `06eef0d0` · `e64d0acc` · `104facd9` · `762ed5c4`, base `origin/develop` @
> `25bbcebd`. Files touched: `core/cube_sdk_library/src/cluster.{hpp,cpp}` (the new writer),
> twelve `core/modules/config_*.cpp` (the call sites), and
> `core/cube_sdk_library/src/tests/` (new offline tests).
>
> **Scope of this section.** This is a `Task`, pointed, and it is a **behaviour-preserving
> refactor**: the acceptance criterion is that the generated config files are *identical*
> before and after. There is no new feature to exercise. Everything below tests
> *sameness*, not function.
>
> **This must be run on a cluster with `cubesys.ha = true`.** All four
> `[oslo_messaging_rabbit]` keys are written inside the HA branch, so on a single-node
> (`ha = false`) node §1, §2 and §5 pass trivially and prove nothing about them — and
> non-HA is precisely the half the dev already verified. **§1 and §4 decide the verdict.**
>
>
> **Every command below was executed read-only on a live `cubesys.ha = true` cluster before
> this section was written**, so the syntax and the shape of the output are known-good rather
> than plausible. Three drafts were corrected by that run; see the note under §1.3.
>
> Tick a row only if you actually ran it. The two easiest things to confuse here are
> `/etc/neutron/neutron.conf` with `/etc/neutron/vpn_agent.ini`, and `/etc/ironic/ironic.conf`
> with `/etc/ironic-inspector/inspector.conf` — in both pairs the two files are *supposed*
> to differ, and in different ways. See §2.

Twelve `config_*.cpp` modules each carried their own copy of the code that writes a service's
AMQP client settings — fourteen call sites in total, because neutron and ironic each write
two files. That code is now one function, `SetMqClientConfig()` in
`core/cube_sdk_library/src/cluster.cpp:199`, and the fourteen call sites call it.

The six keys it writes, and where:

| Section | Key | Written when |
|---|---|---|
| `[DEFAULT]` | `transport_url` | always |
| `[DEFAULT]` | `rpc_response_timeout = 1200` | always **except** `/etc/neutron/vpn_agent.ini` |
| `[oslo_messaging_rabbit]` | `rabbit_retry_interval = 1` | only when `cubesys.ha = true` |
| `[oslo_messaging_rabbit]` | `rabbit_retry_backoff = 2` | only when `cubesys.ha = true` |
| `[oslo_messaging_rabbit]` | `amqp_durable_queues = true` | only when `cubesys.ha = true` |
| `[oslo_messaging_rabbit]` | `rabbit_ha_queues = true` | only when `cubesys.ha = true` |

Three things that make results readable:

1. **An `[oslo_messaging_rabbit]` section header with no keys under it is normal on a non-HA
   node.** The header comes from each module's managed-section list (e.g.
   `config_glance.cpp:324`), which has nothing to do with the HA branch. On `ha = true` the
   header must have four keys under it; on `ha = false` the header alone is correct.
2. **`transport_url` names every control node on HA, not the VIP.** On `ha = true` it is
   `rabbit://openstack:<pass>@<ctrl1>:5672,openstack:<pass>@<ctrl2>:5672,…`. A single host in
   it on an HA cluster is a real failure.
3. **Role gating was deliberately left alone.** Not every service writes its config on every
   node: nova, octavia and cyborg admit compute nodes, the other nine are control-only, and
   neutron writes `neutron.conf` on control and `vpn_agent.ini` on compute. A file missing on
   a node is not necessarily a defect — check the node's role first. See §3.

### Preconditions

- **A cluster with `cubesys.ha = true`.** This is the whole point. On a single-node
  `ha = false` install every `[oslo_messaging_rabbit]` row below passes trivially and proves
  nothing about the half that is still unverified.
- At least one **compute** node reachable, for §3.
- For §4 only: the ability to capture a baseline **before** the new `hex_config` is in place.
  If the cluster is already running the new build, §4 is not available retrospectively — say
  so rather than skipping it silently, and §1/§2/§5 carry the verdict.

### 1 · On an HA cluster every AMQP client carries all six keys

This is the regression that would otherwise go unnoticed: a service that silently lost a key
keeps working until the broker stops accepting the shape it still asks for.

|---|---|---|
| 1.1 | `for f in /etc/cinder/cinder.conf /etc/cyborg/cyborg.conf /etc/designate/designate.conf /etc/glance/glance-api.conf /etc/heat/heat.conf /etc/ironic/ironic.conf /etc/ironic-inspector/inspector.conf /etc/manila/manila.conf /etc/masakari/masakari.conf /etc/neutron/neutron.conf /etc/nova/nova.conf /etc/octavia/octavia.conf /etc/watcher/watcher.conf; do [ -f "$f" ] && printf '%-42s %s\n' "$f" "$(grep -c '^rpc_response_timeout = 1200' $f)"; done` | Every file present on this node prints `1`. A `0` on a file that exists is a failure. (Files absent because of this node's role are expected — see §3.) |
| 1.2 | `for f in /etc/cinder/cinder.conf /etc/cyborg/cyborg.conf /etc/designate/designate.conf /etc/glance/glance-api.conf /etc/heat/heat.conf /etc/ironic/ironic.conf /etc/ironic-inspector/inspector.conf /etc/manila/manila.conf /etc/masakari/masakari.conf /etc/neutron/neutron.conf /etc/neutron/vpn_agent.ini /etc/nova/nova.conf /etc/octavia/octavia.conf /etc/watcher/watcher.conf; do [ -f "$f" ] && printf '%-42s %s\n' "$f" "$(sed -n '/^\[oslo_messaging_rabbit\]/,/^\[/p' $f | grep -cE '^(rabbit_retry_interval = 1\|rabbit_retry_backoff = 2\|amqp_durable_queues = true\|rabbit_ha_queues = true)$')"; done` | Every file present on this node prints `4` — all four HA keys, with these exact values. Any other number is a failure. |
| 1.3 | `sed -n '/^\[DEFAULT\]/,/^\[/p' /etc/nova/nova.conf \| grep '^transport_url'` | One line: `transport_url = rabbit://openstack:<pass>@<ctrl1>:5672,openstack:<pass>@<ctrl2>:5672,openstack:<pass>@<ctrl3>:5672` — **one `openstack:<pass>@host:5672` triple per control node**, comma-separated. A single host on a 3-node cluster is a failure. |
| 1.4 | `for f in /etc/cinder/cinder.conf /etc/glance/glance-api.conf /etc/neutron/neutron.conf /etc/octavia/octavia.conf /etc/nova/nova.conf; do [ -f "$f" ] && echo "$f: $(sed -n '/^\[DEFAULT\]/,/^\[/p' $f \| grep '^transport_url' \| md5sum \| cut -c1-8)"; done` | All five print the **same** 8-char digest — every client is pointed at the same broker list. A differing digest means one module diverged. |

> **Why 1.3 and 1.4 scope to `[DEFAULT]`, and why you must too.** Every one of these files
> carries **two** `transport_url` lines: the AMQP one in `[DEFAULT]`, and a second
> `transport_url = kafka://<vip>:9095` in `[oslo_messaging_notifications]` — the notification
> bus, which this ticket does not touch. A bare `grep '^transport_url'` returns both, and the
> digest comparison in 1.4 then compares the wrong thing. Verified on a live HA cluster: all
> fourteen files show `all=2 default=1`, except `/etc/neutron/vpn_agent.ini` which has only
> the one.

### 2 · The two deliberate asymmetries survived the refactor

Both predate this change and are preserved on purpose. If the refactor had "tidied" either
one, the output would no longer be byte-identical — which is the acceptance criterion.

|---|---|---|
| 2.1 | `grep -c '^rpc_response_timeout' /etc/neutron/vpn_agent.ini` (run on a **compute** node) | `0`. The VPN agent config has never carried `rpc_response_timeout` — verified with `git log -S` back to the initial commit. **This is correct, not a defect**; see *Known behaviour*. |
| 2.2 | `sed -n '/^\[oslo_messaging_rabbit\]/,/^\[/p' /etc/neutron/vpn_agent.ini \| grep -c '^rabbit_ha_queues = true'` (compute node) | `1`. Only `rpc_response_timeout` is suppressed for the VPN agent; it still gets all four HA keys. |
| 2.3 | `diff <(sed -n '/^\[oslo_messaging_rabbit\]/,/^\[/p' /etc/ironic/ironic.conf \| grep -E '^[a-z_]+ = ') <(sed -n '/^\[oslo_messaging_rabbit\]/,/^\[/p' /etc/ironic-inspector/inspector.conf \| grep -E '^[a-z_]+ = '); echo "rc=$?"` | `rc=0`, no output. ironic and its inspector are written by two calls to the same writer and must carry identical keys here. **The trailing `grep -E '^[a-z_]+ = '` is required** — the `sed` range includes the *next* section header, and those legitimately differ (`[oslo_middleware]` vs `[oslo_policy]`), so without it the row fails on a healthy cluster. |
| 2.4 | `diff <(sed -n '/^\[DEFAULT\]/,/^\[/p' /etc/ironic/ironic.conf \| grep -E '^(transport_url\|rpc_response_timeout) = ') <(sed -n '/^\[DEFAULT\]/,/^\[/p' /etc/ironic-inspector/inspector.conf \| grep -E '^(transport_url\|rpc_response_timeout) = '); echo "rc=$?"` | `rc=0`. Both files get the same `transport_url` and the same `rpc_response_timeout = 1200`. |

### 3 · Role gating is unchanged — the right services write on the right nodes

The refactor deliberately left every `IsControl` / `IsCompute` guard where it was. This is
what would catch a guard accidentally widened or dropped.

> **This section needs a node whose `cubesys.role` is `compute`, and only that.** On a
> `control-converged` node both guards are true, so all fourteen files are written and every
> row below passes without testing anything — confirmed on a live HA cluster where all three
> nodes are `control-converged` and all fourteen files carry the full key set. Check first
> with `grep '^cubesys.role' /etc/settings.txt` on each node. **If the cluster has no pure
> `compute` node, mark §3 not-exercisable and say so on the ticket** rather than ticking it;
> §1, §2 and §4 still carry the verdict.

|---|---|---|
| 3.1 | On each node: `grep '^cubesys.role' /etc/settings.txt` | Tells you which rows apply. `control-converged` → skip 3.2 and 3.3, they are vacuous there. |
| 3.2 | On a **pure compute** node: `for f in /etc/nova/nova.conf /etc/octavia/octavia.conf /etc/cyborg/cyborg.conf /etc/neutron/vpn_agent.ini; do printf '%-36s %s\n' "$f" "$(sed -n '/^\[DEFAULT\]/,/^\[/p' $f 2>/dev/null \| grep -c '^transport_url')"; done` | Each prints `1`. These four are the AMQP clients that legitimately write on compute nodes (`IsControl \|\| IsCompute`, plus neutron's VPN agent). |
| 3.3 | On a **pure compute** node: `for f in /etc/heat/heat.conf /etc/masakari/masakari.conf /etc/designate/designate.conf /etc/watcher/watcher.conf /etc/manila/manila.conf /etc/cinder/cinder.conf /etc/glance/glance-api.conf /etc/ironic/ironic.conf; do printf '%-36s %s\n' "$f" "$(grep -c '^rpc_response_timeout' $f 2>/dev/null \|\| echo absent)"; done` | These eight are **control-only**. Either the file is absent, or it carries no AMQP keys (`0`). A pure compute node that has acquired `rpc_response_timeout = 1200` for any of them means a guard was widened — a real failure. |
| 3.4 | `hex_config status 2>&1 \| tail -20` on each node touched | No module reports an error. |

### 4 · Byte-identical output — the acceptance criterion itself

This is the AC verbatim: *the generated config file before and after the refactor is
identical.* It needs a baseline captured before the new `hex_config` is in place.

|---|---|---|
| 4.1 | With the **old** binary still in place: `mkdir -p /tmp/qa1426-base && for f in /etc/cinder/cinder.conf /etc/cyborg/cyborg.conf /etc/designate/designate.conf /etc/glance/glance-api.conf /etc/heat/heat.conf /etc/ironic/ironic.conf /etc/ironic-inspector/inspector.conf /etc/manila/manila.conf /etc/masakari/masakari.conf /etc/neutron/neutron.conf /etc/neutron/vpn_agent.ini /etc/nova/nova.conf /etc/octavia/octavia.conf /etc/watcher/watcher.conf; do [ -f "$f" ] && cp -a "$f" /tmp/qa1426-base/$(echo $f \| tr / _); done; ls /tmp/qa1426-base \| wc -l` | Prints the number of AMQP client configs this node has. Keep this directory until §Cleanup. |
| 4.2 | Install the build containing PR #1499, then: `hex_config commit` | All modules report `commit(o)`. A module reporting `commit(x)` was **skipped**, not passed — its config was not regenerated and rows below would compare a file nothing rewrote. |
| 4.3 | `for b in /tmp/qa1426-base/*; do f=$(basename $b \| tr _ /); diff -u "$b" "$f" > /dev/null \|\| echo "DIFF: $f"; done; echo done` | Prints only `done`. **Any `DIFF:` line is a failure of the acceptance criterion** — the refactor was supposed to change no generated output. Attach the `diff -u` for that file to the defect. |
| 4.4 | `for b in /tmp/qa1426-base/*; do f=$(basename $b \| tr _ /); stat -c '%U:%G %a' "$b" "$f" \| uniq -c \| grep -q '^ *2 ' \|\| echo "PERM: $f"; done; echo done` | Prints only `done`. Ownership and mode must also be unchanged. |

### 5 · The failure mode this refactor introduced — test it deliberately

Before this change, losing a key meant editing one of fourteen places. Now all fourteen share
one writer, so a single mistake in it takes out **every** service at once — a much larger
blast radius, and the reason §1 is written as an exhaustive sweep rather than a spot check.
The offline unit tests exist to hold this line, so confirm they actually run and actually fail
when they should.

|---|---|---|
| 5.1 | In the x86 build container, from the source tree: `make -C core/cube_sdk_library/src/tests test` | `TEST test_mq_config_01` passes, exit 0. Note that a plain `make all` does **not** build this directory — it runs only under `make test`. |
| 5.2 | Edit `core/cube_sdk_library/src/tests/test_mq_config_01.cpp`, change the CASE1 expectation `config["DEFAULT"]["rpc_response_timeout"] == "1200"` to `"9999"`, re-run 5.1 | The target **fails**: `test_mq_config_01.cpp:53: Assertion failed`. If it still passes, the tests are not being executed and 5.1 proved nothing. Revert the edit afterwards. |

- **`/etc/neutron/vpn_agent.ini` has no `rpc_response_timeout`.** It never has, since the
  initial commit. This refactor preserves it deliberately (via the writer's
  `withRpcTimeout = false` argument) because the acceptance criterion is byte-identical
  output. A decision was taken on 2026-09-15 **not** to open a tracking issue for it. If you
  think it should have the key, that is new scope — a new pointed ticket, not a defect
  against this one.
- **`[oslo_messaging_rabbit]` appears with no keys under it on a non-HA node.** The header
  comes from the module's managed-section list, not from the HA branch. Correct, and not
  something this ticket changed.
- **`config_barbican.cpp` has an `UpdateMqConn()` that was not converted.** It writes
  `kafka://…:9095`, not rabbit. Intentionally out of scope; the comment at
  `config_barbican.cpp:228` already records why.
- **No port, scheme or TLS change ships here.** `5672` and `rabbit://` are unchanged on
  purpose; those belong to **#1167**.

### Cleanup

- `rm -rf /tmp/qa1426-base` once §4 is signed off.
- Revert the §5.2 test edit if you made it (`git checkout -- core/cube_sdk_library/src/tests/test_mq_config_01.cpp`).
- §4 leaves no other residue — it changes no configuration, it only regenerates it.

### Already verified by dev — and what that did not cover

**On a 1cc dev node with `cubesys.ha = false`, against this branch's final commit `762ed5c4`:**
the §4 comparison in full — both sides built in the x86 jail container from the same `25bbcebd`
source (stock, and with the five patches), 12/12 modules `commit(o)` on each, **145 generated
files, `diff -r` rc=0, md5 table identical file by file**, ownership and mode unchanged. On the
fourteen AMQP configs the patched binary produced, `vpn_agent.ini` carried `transport_url` but
no `rpc_response_timeout` while the other thirteen carried both — §2's asymmetry, reproduced on
a real node.

**Offline:** `test_mq_config_01` green in the x86 jail container, six cases, with the §5.2
negative control actually performed — the expectations are executed, not decorative.

**What that did not cover, and what makes §1, §2 and §4 above load-bearing:**

- **The refactored binary running on an HA node.** All four `[oslo_messaging_rabbit]` keys
  live inside the HA branch, so the 145-file zero-diff result says nothing about them. Two
  things narrow that gap: offline cases 2 and 4 pin the HA key set, and those expectations were
  checked read-only against a live `cubesys.ha = true` cluster — across all fourteen files the
  six keys and their exact values match what the pre-refactor code produces today, including
  `vpn_agent.ini` carrying the four HA keys but no `rpc_response_timeout`. What is left is that
  the new binary has not itself run there; whether each of the fourteen call sites passes the
  right arguments is read from the diff, not executed. **§1, §2 and §4 are what close that**,
  and they need exactly what QA has and dev did not: a cluster installed from a build
  containing this change.
- **Compute-node behaviour.** The 1cc run was single-node, and the only HA cluster available
  (sky 3cc) has all three nodes as `control-converged` — so the `IsControl || IsCompute`
  services (nova, octavia, cyborg) and `vpn_agent.ini` have **never** been observed on a node
  where the two guards give different answers. That is §3, and it needs a cluster with a pure
  `compute` node.

- [x] **QA Status declared at Done** — this is a `Task`, which carries a `QA Status`, and the
  one — **do not reopen this ticket**. §1 and §4 are where that is most likely: they are the only
  rows covering the `ha = true` half that has no real-node evidence behind it.

SekiXu:
Closes #1426.

Prerequisite refactor for the AMQP TLS work in #1167. Twelve `config_*.cpp` modules each carried their
own copy of the code that writes a service's `transport_url` and its `[oslo_messaging_rabbit]` keys —
fourteen call sites once neutron's two and ironic's two are counted. Adding `ssl = true` /
`ssl_ca_file` for TLS would mean fourteen edits, and missing one is silent: that service keeps working
until the broker stops accepting the connection shape it still asks for.

**This changes no generated config.** It is a behaviour-preserving refactor so that the TLS change
later lands in one place, with the HA-gate decision made once.

## What changed

`SetMqClientConfig()` in `core/cube_sdk_library/src/cluster.{hpp,cpp}` — one writer, taking the config
by reference the way the three `SetWorkerQueue()` copies already did. Its body keeps the unconditional
keys and the HA-gated keys as **separate blocks**, which is where #1167's ssl parameters go without
restructuring again.

- **glance / cinder / manila** carried byte-identical `SetWorkerQueue()` copies, comment included, and
  their signature is the one the shared writer adopted — so those three functions are deleted and the
  call sites changed in name only.
- **nova / octavia / cyborg / heat / masakari / designate / watcher / ironic** keep their
  `UpdateMqConn()` wrapper; only the body becomes a call. ironic writes twice, to its own config and to
  the inspector's, so it makes two calls.
- **neutron** has two asymmetric call sites. The control node's config gets the full key set; the
  compute node's VPN agent config has never carried `rpc_response_timeout` — confirmed with
  `git log -S` back to the initial commit, never present. That asymmetry is preserved explicitly via
  `withRpcTimeout = false`, because the acceptance criterion is byte-identical output. The
  unconditional `RabbitMqServers()` call at the top of the function is gone with it: it built a string
  that a node which is neither control nor compute then discarded.

**Role gating deliberately stays with the caller.** Which roles run a given service is a property of
that service's deployment, not of the message queue, and the nine `UpdateMqConn()` copies disagree
about it three different ways (`IsControl` only ×5, `IsControl || IsCompute` ×3, neutron's two-stage).
Folding that into the helper would have produced one function with a role flag, which is harder to
change, not easier.

`config_barbican.cpp` is untouched — it writes `kafka://…:9095`, and the `RabbitMqServers` occurrence
at `config_barbican.cpp:228` is comment text, not a call. The comment already there is more complete
than anything this PR would add.

## Why `cluster.cpp` rather than a new `core/rabbitmq/rabbitmq_util.{h,cpp}`

The `mysql_util` precedent would need a new set of `.mk` link settings, and it would split the two
things #1167 has to touch — scheme and port live in `RabbitMqServers()` in `cluster.cpp`, the ssl
parameters would live in the new file. That is the opposite of what this issue is buying.

## Evidence

Taken against this branch's final commit `762ed5c4`, on a 1cc dev node with `cubesys.ha = false`.

Both sides were built in the x86 jail container from the same `25bbcebd` base — stock, and with the
five patches — so the only variable is this change:

| | ref | sha256 | `strings \| grep -c SetMqClientConfig` |
|---|---|---|---|
| stock | `25bbcebd` | `6d158ff0…` | 0 |
| patched | `762ed5c4` | `80bfdf9d…` | 4 |

Each binary then committed the **same** settings file, and the generated tree was snapshotted:

```
stock   (PID 2352543)   12/12 modules commit(o)   145 files
patched (PID 2443459)   12/12 modules commit(o)   145 files

diff -r cfg-stock cfg-patched   →  rc=0, empty output
md5sum table, file by file      →  identical, 145 vs 145
```

On the fourteen AMQP client configs the patched binary produced, `/etc/neutron/vpn_agent.ini` carries
`transport_url` but no `rpc_response_timeout`, while the other thirteen carry both — the asymmetry
described above, reproduced on a real node rather than argued from the diff.

**Offline unit tests (new):** `core/cube_sdk_library/src/tests/test_mq_config_01.cpp`, six cases
pinning the exact key set and values for each shape the fourteen call sites produce — including the
absence of the HA section on a single-node cluster, and neutron's missing `rpc_response_timeout`.
Verified non-vacuous by a negative control: changing the expected `rpc_response_timeout` to `9999`
turns two assertions red and fails the target; reverting restores green. Note that `make all` does not
build that directory — the existing `utils/tests` is not built either. They run under `make test`.

These tests exist because the real-node evidence is expensive and taken rarely: `hex_config`'s dry run
returns at the top of every module's `Commit()` and produces no config at all, and a module-scoped
commit rewrites cluster-wide settings. Both cheap routes are closed.

Three traps are worth recording for anyone reproducing the comparison, since each one produces a
confident-looking wrong answer:

- `hex_config commit /etc/settings.txt` does nothing. Current and new are then the same file, the
  digest matches, and every module reports `commit(x)` — a skip, not a pass. The settings file passed
  in has to differ from `/etc/settings.txt`.
- The "before" side cannot be the binary already on the node. That one carries unrelated `develop`
  drift, which shows up as a diff that has nothing to do with this change.
- An empty `/var/log/hex_config.log` does not mean the commit did not run. `rsyslog`'s `imjournal`
  rate limit (20000 messages / 600s) was tripped by another service mid-run and dropped 12098
  messages, silently blanking the whole window. `journalctl -t hex_config` had all 1594 lines,
  including every module's `commit(o)`.

## Not verified here: the refactored writer on an HA node

The HA branch is where all four `[oslo_messaging_rabbit]` keys live, so the non-HA run above says
nothing about them. Two things narrow that gap; one remains open.

Offline cases 2 and 4 pin the HA key set for the new writer. Those expectations were then checked
against a live `cubesys.ha = true` cluster: across all fourteen config files the six keys and their
exact values match what the pre-refactor code produces today — including `vpn_agent.ini` carrying the
four HA keys but no `rpc_response_timeout`. The test expectations are therefore grounded in production
output, not derived from the diff they are meant to check.

What remains open is that the refactored binary has not itself run on an HA node. Whether each of the
fourteen call sites passes the right arguments is read from the diff, not executed. That needs a
cluster installed from a build containing this change — QA's environment, not a hot-patch. The ticket
carries the rows for it.

## Scope

No port change, no scheme change, no ssl parameters — those are #1167's implementation issue. Only
`*.cpp` / `*.hpp` and one tests `Makefile` change, so nothing enters the rootfs and no ISO rebuild is
required.


Eandalf-Bigstack (commented):
LGTM, we would be merging this PR after v3.1.20, into v3.2.0 instead.

Eandalf-Bigstack (approved):

